PR #71, opened 2013-12-01
Inject a default section for section-less INI files

unho:
This is a workaround for iniparse inability to parse INI files without any section.

Fixes bug #342
Fixes bug #1428


unho:
Closing this since it probably will be never merged.

PR #90, opened 2014-02-17
Convert optrecurse to use argparse

dwaynebailey:
This is a first pass of removing ``optparse`` from ``misc/optrecurse``.  It is probably ready for testing more then it is ready for review.

The strategy has been to remove optparse then pick through breakage to remove optparse methods and classes and replace them with argparse.

At the current state we can:

1. Pass all tests
2. We don't actually work for ``--manpage`` - lots of work needed there
3. There might be some optrecurse positional logic that is broken

At the moment I am not looking are prettiness of code and styleguides.  In fact changes have tried to minimise disruption of layout so that you can see what changed.

When the optparse ripout has finished I'd like a separate PR to deal with:

1. Moving ``argparse._Action()`` classes to use ``add_argument`` i.e. use public methods
2. Create a styleguide for layout of add_argument,  at the moment they are inconsistent and hard to read
3. Remove any setting of ``add_argument`` values that are just repeating the default.

Once we are at this state of cleanup we can then look at moving options around.  I think the aim here is to reduce to size of optrecurse and look to moving options up to storage classes.


unho:
@dwaynebailey I'd say this is worth fixing now. Just checked and there is only 7 files with references to `optparse` on latest master.

claudep:
Isn't argparse Python 2.7 only?

unho:
> Isn't argparse Python 2.7 only?
@claudep I would say it isn't https://docs.python.org/3/library/argparse.html

unho:
On Python 2.6 we will need to install it as a package anyway. But that is a minor issue.

claudep:
Yes, I meant it isn't in the Python 2.6 standard lib.

unho:
Sure, but as I said installing the `argparse` solves this. Actually this is what we are already doing on master https://github.com/translate/translate/blob/master/requirements/required.txt#L3

unho:
@claudep Are you up to finish replacing `optparse` with `argparse`?

claudep:
I have done such work in Django, so I might try that. Will see if I find time in the next days.

unho:
@claudep Thanks. We are looking for a release soon, in a few weeks, before next Pootle release tentatively scheduled for 10 March.

dwaynebailey:
PR replaces by #3379

PR #133, opened 2014-07-16
a check for Python brace format strings like 'hello {name}'

sietse:
pythonbraceformat
-----------------

Checks whether Python brace format strings match.

Python supports both a variant of the `printf` formatting system,
and its own formatting language which uses placeholders enclosed in
braces. The placeholders can be named, numbered, or anonymous; the
former two are filled in from positional args, the latter from keyword
arguments. Example:

    'the {} {0} hungry {insect}'.format('very', insect='caterpiller')
    # --> 'the very very hungry caterpiller'

The `pythonbraceformat` filter checks for the following problems:

* named placeholders that are present in the original, but missing in
 the translation, and vice versa.
* originals and translations that require different numbers of
 positional args.

When the translation has variables not in the original, this can lead to
program crashes. The translation not using all variables the original
uses is safe. Nonetheless, this filter triggers in both cases.

See also: [PEP 3101 -- Advanced String Formatting](http://legacy.python.org/dev/peps/pep-3101/)


unho:
Merged after some squashing and cleanup in 9c42f72d...98c0f74b.

PR #3208, opened 2014-08-29
Do not append dirs without files in it

williambr:
Otherwise python distutils --record would list a directory when it
shouldn't.


unho:
Merged in edb8a4c4. Thanks!

PR #3218, opened 2014-10-13
po2moz: Extensions are either "EXT" or "EXT.po".

okulev:
Firefox OS l10n has a file named "settings.device.properties" that has many dots and po2moz thinks that extension is ".device.properties" instead of just ".properties". The pull request fixes this by forcing extensions to be either "EXT" or "EXT.po".


unho:
@dwaynebailey Any input on the status of this PR?

unho:
Just for reference, the change proposed by @dwaynebailey is merged as e146f421.

PR #3219, opened 2014-10-13
Fix printf filter to allow reordering boost::format positional directive...

mlandres:
That's my approach to handle [boost::format](http://www.boost.org/doc/libs/1_55_0/libs/format/doc/format.html) positional  directives.

The patch basically treats the boost  directives `%1%` like it's printf `%1$` pendant. 
At the same time it prevents treating `%%` (a literal %) as a directive.


unho:
@mlandres This contribution seems to come from no existing repository and we can't merge it. Please fork our repo, recreate these changes and push again to this PR if possible, and if not create another PR and close this one.

mlandres:
I recreated the patch in #3267

PR #3224, opened 2014-10-27
Dump current units content instead of file

SirAnthony:
`_file` is not updating anywhere and hold content of loaded file, and when new translated units manually added `__str__` still returns file content instead of current units content.


nijel:
See https://github.com/translate/translate/issues/3249 for explanation what this pull request solves.

unho:
Merged as 40ab1e74. @SirAnthony Thanks for your contribution.

PR #3247, opened 2015-01-23
punctuation copied from ja

babycaseny:
these punctuation also applicable to Chinese.


unho:
Merged in 9c92ad3c. Thanks for your contribution.

PR #3250, opened 2015-02-19
Styleguide: Reorganize string-related guidelines

jleclanche:
- Remove the note on arbitrary single vs double quoting, always use double.
- Rename "String quoting" section into "Strings" section
- Move "String formatting" to a subsection of "Strings"

### DIFF
--- a/docs/developers/styleguide.rst
+++ b/docs/developers/styleguide.rst
@@ -119,6 +119,48 @@ Example:
             return self.name.upper() + '!!!!111'
 
 
+Strings
+~~~~~~~
+
+- Double quotes are suggested over single quotes, but always try to respect the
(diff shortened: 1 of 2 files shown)


julen:
As discussed on IRC, in terms of quoting let's try to word something along the lines of "respect the surrounding quoting style, use your preference consistently in new files"

jleclanche:
@julen can we close the matter once and for all? I still havent heard any opposition to double quoting. From anyone. The closest thing I've heard was from Leandro and it was "I have an argument and I won't tell you what it is".

julen:
Sure, but I see no need to enforce any quoting style. Suggesting one or the other doesn't change the situation much, so up to you, I'm fine with that.

julen:
`s/preferred/suggested/` and good to merge.

PR #3251, opened 2015-02-20
resx addnote() bug fix - avoid endless comment duplication if there is leading/trailing whitespace

sjhale:
Discovered an issue whereby existing comments were being continually duplicated if the original comment contained leading or trailing whitespace. I've added a unit test to cover this scenario and fixed it by adding .strip() to the existing comment check in resx.py


jleclanche:
gtm

unho:
Merged in b880dcee. Thanks!

PR #3252, opened 2015-02-26
Include both types of punctuations in translate/lang/zh.py

babycaseny:


unho:
@babycaseny Can you provide an explanation of which problem solves adding those characters?

unho:
Closing this since it seems to be a duplicate of already merged #3247.

PR #3253, opened 2015-02-27
Mozilla lang files: import only real comments, not meta tags

flodolo:
These days reference .lang files carry a lot of meta tags, no point in associating them to strings.

https://github.com/mozilla-l10n/langchecker/wiki/.lang-files-format

### DIFF
--- a/translate/storage/mozilla_lang.py
+++ b/translate/storage/mozilla_lang.py
@@ -67,7 +67,7 @@ class LangStore(txt.TxtFile):
         super(LangStore, self).__init__(inputfile, flavour, encoding)
 
     def parse(self, lines):
-        #Have we just seen a ';' line, and so are ready for a translation
+        # Have we just seen a ';' line, and so are ready for a translation
         readyTrans = False
         comment = ""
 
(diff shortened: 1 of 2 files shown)


translate/storage/mozilla_lang.py
```diff
@@ -91,7 +91,8 @@ def parse(self, lines):
                 readyTrans = False  # We already have our translation
                 continue
 
-            if line.startswith('#'):  # A comment
+            if line.startswith('#') and not line.startswith('##'):
```
unho: If comments always start with `# ` (note the space) then it might be easier to alter the condition to:

```python
            if line.startswith('# '):
                # Read comments, but not meta tags (e.g. '## TAG')
```

flodolo: Normally comments should start with `# `, but it's not impossible for a comment to miss the space. I think it's safer to exclude `##`.
This is also the same check we do while parsing .lang files in our tools.

unho: If there is not specification to follow, then I agree.

PR #3261, opened 2015-04-23
data.py: Fix pt_BR Plural-Forms expression

josephgbr:
The correct expression can be confirmed in http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html


unho:
@josephgbr Next time please create a separate branch to keep your proposed changes. Merged. Thanks.

Issue #3263, opened 2015-04-25
Incorrect "Plural-Form"s for some languages


fbobraga:
I'v compared te content in the file https://github.com/translate/translate/blob/master/translate/lang/data.py with the page in http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html, and found it:
- "Plural-Form" does not match:

cy  "Welsh"
hy  "Armenian"
is  "Icelandic"
jv  "Javanese"
mk  "Macedonian"
sl  "Slovenian"
tg  "Tajik"
tr  "Turkish"
wo  "Wolof"
- in the file, but not in the page:

bn_IN   "Bengali (India)"
ca@valencia "Catalan; Valencian (Valencia)"
en_GB   "English (United Kingdom)"
en_ZA   "English (South Africa)"
ht  "Haitian; Haitian Creole"
nqo "Ko""
st  "Sotho Southern"
ve  "Venda"
zh_CN   "Chinese (China)"
zh_HK   "Chinese (Hong Kong)"
zh_TW   "Chinese (Taiwan)"
zu  "Zulu"
- in the page, but not in the file:

ach "Acholi"
anp "Angika"
as  "Assamese"
ay  "Aymará"
brx "Bodo"
cgg "Chiga"
doi "Dogri"
es_AR   "Argentinean Spanish"
hne "Chhattisgarhi"
jbo "Lojban"
kl  "Greenlandic"
mni "Manipuri"
mnk "Mandinka"
my  "Burmese"
no  "Norwegian (old code)"
rw  "Kinyarwanda"
sat "Santali"
sd  "Sindhi"
se  "Northern Sami"
ur  "Urdu"
uz  "Uzbek"
zh  "Chinese [2]"
zh  "Chinese [3]"

unho:
Created two PRs to add the missing plural rules:

* https://github.com/translate/l10n-guide/issues/32
* #3264

These still don't address the inconsistent plural rules problem.

nijel:
translate-toolkit side will be handled in https://github.com/translate/translate/issues/3578, the l10n-guide issues should be addressed in that repository (if somebody is interested in maintaining it).

PR #3264, opened 2015-04-27
Add new language plurals

unho:
From http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html

@dwaynebailey Since this impacts Pootle it might not be a good idea to add these languages.

References https://github.com/translate/translate/issues/3263

Related PR: https://github.com/translate/l10n-guide/pull/32


dwaynebailey:
@unho I think you're missing the point here.  The errors exist for the same reason that I raised the issue about HTML quality check pages where you wanted to dump in all checks without generating them from the check comments.  That is that if they are not autogenerated then you will always end up with a mess.  If you want to fix this then please do it properly, with one place to make all errors. I'm a bit sad actually that you didn't pick this up since we talked about the same type of issue only last week.

I'm not sure why we wouldn't want to add these languages i.t.o the comment re Pootle.

unho:
@dwaynebailey TTK is the place where plurals are kept. But there is also a table on the localization guide. I don't see how we can create the plural rules from the other project without coupling the former with the other project.

I was trying to say that if those languages are not considered to be core, we might want not to add them so we don't overcrowd Pootle UI. But given the live filtering we now have in Pootle UI that shouldn't be a big problem.

dwaynebailey:
@unho coupling is the least of evils, it is then autogenerated from the toolkit which would be the right thing to do.  Otherwise you are defending the pain of coupling by the definate pain of these things always being out of sync and needing multiple place to configure.

I'm not concerned about overcrowding Pootle UI.  My personal view is that we should actually have all ISO languages in Pootle in the long term.

unho:
@dwaynebailey Since you say it should be autogenerated from the toolkit, and this is a PR for the toolkit, I am going to take that as a gtm.

dwaynebailey:
@unho yip gtm

PR #3266, opened 2015-05-10
Adding removeuntranslated options to po2json and po2resx

sgarrad:
Refactored po2prop option removeuntranslated into convert.py and added
the ability to remove the values for po2resx and po2json.

### DIFF
--- a/docs/commands/json2po.rst
+++ b/docs/commands/json2po.rst
@@ -48,10 +48,10 @@ Options (json2po):
 
 Options (po2json):
 
---version            show program's version number and exit
--h, --help           show this help message and exit
---manpage            output a manpage based on the help
---progress=PROGRESS    show progress as: :doc:`dots, none, bar, names, verbose <option_progress>`
+--version             show program's version number and exit
+-h, --help            show this help message and exit
+--manpage             output a manpage based on the help
+--progress=PROGRESS   show progress as: :doc:`dots, none, bar, names, verbose <option_progress>`
 --errorlevel=ERRORLEVEL
                       show errorlevel as: :doc:`none, message, exception,
                       traceback <option_errorlevel>`
@@ -59,10 +59,11 @@ Options (po2json):
 -x EXCLUDE, --exclude=EXCLUDE   exclude names matching EXCLUDE from input paths
 -o OUTPUT, --output=OUTPUT      write to OUTPUT in JSON format
 -t TEMPLATE, --template=TEMPLATE  read from TEMPLATE in JSON format
--S, --timestamp      skip conversion if the output file has newer timestamp
---threshold=PERCENT  only convert files where the translation completion is above PERCENT
---fuzzy              use translations marked fuzzy
---nofuzzy            don't use translations marked fuzzy (default)
+-S, --timestamp       skip conversion if the output file has newer timestamp
+--threshold=PERCENT   only convert files where the translation completion is above PERCENT
+--fuzzy               use translations marked fuzzy
+--nofuzzy             don't use translations marked fuzzy (default)
+--removeuntranslated  remove key value from output if it is untranslated
 
 .. _json2po#examples:
 
--- a/docs/commands/resx2po.rst
+++ b/docs/commands/resx2po.rst
@@ -44,10 +44,10 @@ Options (resx2po):
 
 Options (po2resx):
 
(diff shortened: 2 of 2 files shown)


unho:
@sgarrad Please rebase this on top of latest master. Also fix the broken build and split this into separate commits:

* missing fuzzy option in po2resx
* removeuntranslated option

unho:
@sgarrad Don't mind about the missing fuzzy option in po2resx since we already merged that minor change.

unho:
@sgarrad What is the purpose of adding such option to only those commands? And can you please add some tests?

sgarrad:
@unho Only adding them to the po2json and po2resx is based off personal requirement to needing them for only those convertors. However, this can likely be used with any convertor that supports fuzzy matching.

Logic can be different through, for example in the po2prop, the command was already present, however the results are no key or value in the output property file, however, with json and resx for instanace, removing the key and value would likely be bad, so removing the translation i.e. replacing with u"" is a better approach. This also fits in with the existing logic for when content is present/ not present.

I'll add the tests and rebase the commit.

sgarrad:
Not sure, what classes as coverage fail, but I've added tests for both po2resx and po2json for the `--removeuntranslated` option, including rebasing to the latest.

DmitryGonchar:
This option would be useful

mkjaxxo:
Is there a change to get the removeuntranslated options to po2json in the near future? This would be a great enhancement and as DmitryGonchar wrote: Currently untranslated strings will result in empty text in the software instead of the default fallback language.

unho:
> Is there a chance to get the removeuntranslated options to po2json in the near future? This would be a great enhancement and as DmitryGonchar wrote: Currently untranslated strings will result in empty text in the software instead of the default fallback language.

`--removeuntranslated` forces the output to have blank translations instead of falling back to the source text. For example this PO:

```po
#: .stringone
msgid "Foo Bar"
msgstr ""
```

Gets converted to this JSON (uses the original text):

```json
{
    "stringone": "Foo Bar"
} 
```

And with `--removeuntranslated` gets converted to this other JSON (uses blank text):

```json
{
    "stringone": ""
} 
```

@mkjaxxo @DmitryGonchar I certainly don't know your needs (perhaps your JSON translation backend replaces empty translation with the right text), but default current po2json behavior looks sane enough. Anyway I am looking at trying to get this landed.

mkjaxxo:
Thanks for your reply @unho.

Currently we have this source file:

`{
  "id.Name" : "Name",
  "id.Type" : "Type",
  "id.Code" : "Code"
}`

wich results in 

`{
  "id.Name" : "",
  "id.Type" : "translated Type",
  "id.Code" : ""
}`


In the Software, now "" is displayed as id.Name and id.Code and not the source text form the default language. I thought --removeuntranslated would end up in 

`{
  "id.Type" : "translated Type"
}`

unho:
@mkjaxxo Just checked the actual behavior and I thought that the actual output was:

```json
{
  "id.Name" : "Name",
  "id.Type" : "translated Type",
  "id.Code" : "Code"
}
```

but it turns out it is (which is also what `--removeuntranslated` gives us):

```json
{
  "id.Name" : "",
  "id.Type" : "translated Type",
  "id.Code" : ""
}
```

Would be the first result (the one I thought it was outputting) ok for your workflow?

unho:
Just compared this PR and the current behavior of `--removeuntranslated` in other commands, and in other commands it actually removes the fuzzy and untranslated units (outputting only translated units).

mkjaxxo:
Thanks @unho. "remove untranslated" would be perfect. But if po2json would add the source string for untranslated strings to the target file (result 1 from https://github.com/translate/translate/pull/3266#issuecomment-273804377), would be fine too. Way more useful than "blank untranslated".

unho:
@mkjaxxo I have created https://github.com/translate/translate/pull/3549 which add `--removeuntranslated` support for `po2json`. I considered that this close to final 2.0.0 release it is not a good idea to alter the default behavior for `po2json` so it outputs the original source text for untranslated units.

unho:
I am closing this PR since it provides a broken fix. https://github.com/translate/translate/pull/3549 replaces most of it (except the resx part).

mkjaxxo:
Thanks a lot @unho for your work and your support!

PR #3267, opened 2015-05-12
Fix printf filter to allow reordering boost::format positional direct…

mlandres:
That's my approach to handle [boost::format](http://www.boost.org/doc/libs/1_55_0/libs/format/doc/format.html) positional  directives.

The patch basically treats the boost  directives `%1%` like it's printf `%1$` pendant. 
At the same time it prevents treating `%%` (a literal %) as a directive.

(replaces #3219 as the repo was lost)


unho:
@mlandres This looks good despite the checker fails when parsing the string "**_%1$+5d_ %1$d**". I am merging this along with some tests I have added. I would appreciate if you can look into fixing the problem with that string.

unho:
Merged in 7dc18a5b.

PR #3268, opened 2015-05-12
printf: fix handling multiple occurrences of positional directives

mlandres:
In reply to #3267:  'the checker fails when parsing the string "%1$+5d %1$d"'

The current code IMO enforces that the same positional parameter always has the same format, even if it occurs multiple times. The above string causes a 'Different printf variable' exception when comparing str2(%1$+5d) against str1(%1$d) after it successfully compared str2(%1$+5d) against str1(%1$+5d).

If multiple occurrences of the same positional parameter (%1$) with different formats are allowed, the inner loop comparing the parameters in str1 against a specific one in str2, should end as soon as a match was found. That's what the above patch tries.

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -914,6 +914,7 @@ class StandardChecker(TranslationChecker):
 
             if str2ord:
                 str1ord = None
+                gotmatch = False
 
                 for var_num1, match1 in enumerate(printf_pat.finditer(str1)):
                     count1 = var_num1 + 1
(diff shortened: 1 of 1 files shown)


unho:
@mlandres Travis build is broken. Please fix it.

mlandres:
Fixed. If we can't leave the loop immediately after a match was found without braking subsequent checks, simply remember if there was a match.

unho:
Merged in e8ebfe76. Thanks!

PR #3277, opened 2015-07-30
Add test for posegment

miurahr:
- add a missing test for posegment

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

### DIFF
--- a/translate/tools/test_posegment.py
+++ b/translate/tools/test_posegment.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from translate.lang import factory as lang_factory
+from translate.misc import wStringIO
+from translate.storage import po
+from translate.tools import posegment
+
+
+class TestPOSegment:
+
+    def posegment(self, posource, sourcelanguage, targetlanguage, stripspaces=True, onlyaligned=True):
+        """helper that convert po source without requiring files"""
+        inputfile = wStringIO.StringIO(posource)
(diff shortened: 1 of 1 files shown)


unho:
Several lines are obviously longer than 80 characters. We enforce a 79 characters line length with a soft limit of 84 characters for some special cases. See our styleguide http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html

translate/tools/test_posegment.py
```diff
@@ -0,0 +1,41 @@
...
+#: test/test.py:112
+msgid ""
+"Please let us know if you have any specific needs (A/V requirements, "
+"multiple microphones, a table, etc).  Note for example that 'audio out' is "
+"not provided for your computer unless you tell us in advance."
+msgstr ""
+"特に必要な物(A/V機器、複数のマイク、テーブルetc)があれば教えて下さい。例とし"
+"て、コンピュータからの「音声出力」は事前にお知らせ頂いていない場合は提供でき"
+"ないことに注意して下さい。"
+
```
unho: Is this blank line actually necessary?

dwaynebailey:
@unho actually in tests we're pretty soft and have PEP8 turned off IIRC

translate/tools/test_posegment.py
miurahr: This test case is come from live application.
A long line is copied from test_pogrep.
I'd like to update to drop blank line.

PR #3279, opened 2015-07-30
posegment: Japanese test and fix

miurahr:
posegment has unknown bug that fails test that has some complexity.

This test should pass.

It is depend on #3277

### DIFF
--- a/translate/lang/ja.py
+++ b/translate/lang/ja.py
@@ -31,9 +31,9 @@ from translate.lang import common
 class ja(common.Common):
     """This class represents Japanese."""
 
-    listseperator = u"、"
+    listseperator = u"、､，,"
 
-    sentenceend = u"。！？!?…"
+    sentenceend = u"。｡！？!?…"
 
     # Compared to common.py, we make the space after the sentence ending
     # optional and don't demand an uppercase letter to follow.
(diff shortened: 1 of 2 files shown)


friedelwolff:
As far as I know, posegment only segments if it detects the same number of sentences in both languages. If I read the test correctly, the English has 3 sentences, and the Japanese has four sentences. Although some alignment might be possible, posegment does not implement something advanced enough, and simply refuses to guess. Maybe it can incorporate more advanced sentence alignment methods in future. (I'm writing based on memory, I didn't check the current code.)

miurahr:
Thnak you for your comment. 
I tried simplifying to minimum and found a root cause.

Test case I used is a file download from live example of django documentation.

It includes half width punctuation sign in japanese msgstr.
http://www.unicode.org/charts/PDF/UFF00.pdf

miurahr:
I've add a fix.

miurahr:
This fixes #3280

jleclanche:
lgtm but please squash your test commits

unho:
> This fixes #3280

@miurahr If any of the commits in this PR fixes some issue, can you please update that commit message to include the reference to the issue?

unho:
Merged in 8453724a. Thanks!

PR #3282, opened 2015-08-11
Remove import-time logging calls

jleclanche:


translate/storage/po.py
```diff
@@ -36,13 +36,10 @@
 
 if platform.python_implementation() == "CPython":
     if usecpo == "1":
-        logging.info("Using cPO")
         from translate.storage.cpo import *  # pylint: disable=W0401,W0614
     elif usecpo == "2":
-        logging.info("Using new fPO")
         from translate.storage.fpo import *  # pylint: disable=W0401,W0614
     else:
-        logging.info("Using Python PO")
```
unho: What is the reason to drop these logging calls?

dwaynebailey:
I don't think this is the correct appraoch.  Those should be maybe debug calls, ttk should use a named logger and Pootle should set the logger to the correct level by default.

translate/storage/po.py
jleclanche: They're useless and annoying.

jleclanche:
@dwaynebailey Ideologically maybe. Concretely, how are these useful? If we need to check the PO implementation for debug purposes (how often does that happen?), there's other ways.

translate/storage/po.py
unho: Testing the new Pootle command to list contributors I did finally saw why you want to drop this. Anyway I think it is better to use `logging.debug`.

dwaynebailey:
@jleclanche it is a bandaid.  We'll likely want logging elsewhere in Pootle and it will come back because the root isn't addressed.  So I think the correct solution is still to turn it to debug, since it is a debug issue and then Pootle will handle it as expected.

jleclanche:
We discussed this further on IRC. The logging calls are unnecessary, and having them at import-time is harmful and they'll show up far too often.

Landing on @dwaynebailey's gtm on IRC.

PR #3285, opened 2015-08-17
Removed usage of remaining print-as-a-statement

claudep:
print 'as a statement' does not exist any longer in Python 3.

### DIFF
--- a/translate/search/match.py
+++ b/translate/search/match.py
@@ -90,8 +90,6 @@ class matcher(object):
         for store in stores:
             self.extendtm(store.units, store=store, sort=False)
         self.candidates.units.sort(key=sourcelen, reverse=self.sort_reverse)
-        # print "TM initialised with %d candidates (%d to %d characters long)" % \
-        #        (len(self.candidates.units), len(self.candidates.units[0].source), len(self.candidates.units[-1].source))
 
     def extendtm(self, units, store=None, sort=True):
         """Extends the memory with extra unit(s).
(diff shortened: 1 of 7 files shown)


translate/search/match.py
```diff
@@ -90,8 +90,8 @@ def inittm(self, stores, reverse=False):
         for store in stores:
             self.extendtm(store.units, store=store, sort=False)
         self.candidates.units.sort(key=sourcelen, reverse=self.sort_reverse)
-        # print "TM initialised with %d candidates (%d to %d characters long)" % \
-        #        (len(self.candidates.units), len(self.candidates.units[0].source), len(self.candidates.units[-1].source))
+        # print("TM initialised with %d candidates (%d to %d characters long)" % \
+        #        (len(self.candidates.units), len(self.candidates.units[0].source), len(self.candidates.units[-1].source)))
```
unho: Just wondering if we should remove all this commented code.

unho:
lgtm

translate/search/match.py
claudep: Maybe replaced by a logging call? Anyway, that could be another patch.

unho: Sure.

jleclanche:
You can probably remove most of these comments...

claudep:
OK, patch rebased with commented prints removed.

unho:
Merged. Thanks!

PR #3286, opened 2015-08-17
Added relative imports where appropriate

claudep:
Python 3 does not accept implicit relative imports.
Making those explicit.

### DIFF
--- a/translate/storage/placeables/__init__.py
+++ b/translate/storage/placeables/__init__.py
@@ -45,14 +45,14 @@ The placeables model follows the XLIFF standard's list of placeables.
 Please refer to the XLIFF specification to get a better understanding.
 """
 
-import base
-import interfaces
-import general
-import xliff
-from base import *
-from base import __all__ as all_your_base
-from strelem import StringElem
-from parse import parse
+from . import base
+from . import interfaces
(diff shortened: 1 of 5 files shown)


translate/storage/placeables/__init__.py
```diff
@@ -45,14 +45,14 @@
...
-import general
-import xliff
-from base import *
-from base import __all__ as all_your_base
-from strelem import StringElem
-from parse import parse
+from . import base
+from . import interfaces
+from . import general
+from . import xliff
```
unho: Would be really necessary to import these modules to export them again in this module `__all__`? I ask because you will be accessing them using `translate.storage.placeables.base` anywhere.

claudep: Maybe not, but I didn't focus on the import roles, only on the importing syntax.

I tried that: https://gist.github.com/claudep/455819983e3eb63649ca and didn't get any failures, should this be part of this patch?

unho: @claudep Lets be cautious and leave that cleanup out. If you rebase this PR I will merge it.

unho:
Merged. Thanks!

PR #3287, opened 2015-08-17
Replaced cStringIO.StringIO by io.BytesIO

claudep:
The cStringIO module is no more in Python 3, and the io module
has been introduced in Python 2.6.

### DIFF
--- a/translate/convert/prop2mozfunny.py
+++ b/translate/convert/prop2mozfunny.py
@@ -37,6 +37,8 @@ def prop2inc(pf):
             else:
                 for blank in pendingblanks:
                     yield blank
+                if isinstance(comment, unicode):
+                    comment = comment.encode("UTF-8")
                 # TODO: could convert commented # x=y back to # #define x y
                 yield comment + "\n"
         if unit.isblank():
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -26,7 +26,7 @@ files (pofile).
 import copy
 import re
 import textwrap
-from cStringIO import StringIO
+from io import BytesIO
 
 from translate.lang import data
 from translate.misc import quote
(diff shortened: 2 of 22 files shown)


claudep:
The docs build error is:
```/home/travis/build/translate/translate/translate/misc/wStringIO.py:docstring of translate.misc.wStringIO.CatchStringOutput.seek:5: WARNING: Definition list ends without a blank line; unexpected unindent.```
I'm afraid the issue might be in a Python lib docstring (there are no `translate.misc.wStringIO.CatchStringOutput.seek` method). Any idea?

File: translate/storage/pypo.py [hunk 1]
```diff
@@ -26,7 +26,7 @@
 import copy
 import re
 import textwrap
-from cStringIO import StringIO
+from six import BytesIO
```
unho: Why not `from io import BytesIO`?

translate/convert/prop2mozfunny.py
```diff
@@ -37,6 +37,8 @@ def prop2inc(pf):
             else:
                 for blank in pendingblanks:
                     yield blank
+                if isinstance(comment, unicode):
+                    comment = comment.encode("UTF-8")
```
unho: This seems an unrelated change. What is its purpose?

translate/storage/pypo.py
claudep: Good catch, we don't need six at this stage.

translate/convert/prop2mozfunny.py
claudep: `prop2inc` produces mixed bytestring/unicode lines, which is an issue when using writelines to a bytes file object.

unho:
> I'm afraid the issue might be in a Python lib docstring (there are no translate.misc.wStringIO.CatchStringOutput.seek method). Any idea?

I guess that might be the cause. New commits have been merged to master, so please rebase and lets see if we are lucky and the error goes away.

julen:
> I'm afraid the issue might be in a Python lib docstring (there are no `translate.misc.wStringIO.CatchStringOutput.seek` method). Any idea?

In *docs/api/misc.rst* under the `wStringIO` section try to remove `:inherited-members:`. Such option can cause trouble when docstrings come from 3rd party sources.

claudep:
Yeah, that's what I did to stop Travis complaining!

unho:
@claudep Please rebase this PR so we can cleanly merge it.

claudep:
Done.

unho:
Merged. Thanks!

PR #3288, opened 2015-08-18
Removing duplicate messages in PO file parsers

phlax:
PO files can't cope with duplicate messages, so they are removed when parsing
and a warning is logged

Fixes translate/pootle#3564

### DIFF
--- a/translate/convert/test_po2ts.py
+++ b/translate/convert/test_po2ts.py
@@ -78,7 +78,7 @@ msgstr "b"
 '''
         tsfile = self.po2ts(posource)
         print(tsfile)
-        assert tsfile.find("English") != tsfile.rfind("English")
+        assert tsfile.find("English") == tsfile.rfind("English")
 
 
 class TestPO2TSCommand(test_convert.TestConvertCommand, TestPO2TS):
--- a/translate/storage/cpo.py
+++ b/translate/storage/cpo.py
@@ -667,11 +667,17 @@ class pofile(pocommon.pofile):
                         markedpos.append(thepo)
                     thepo.setcontext(" ".join(thepo.getlocations()))
                     uniqueunits.append(thepo)
+                else:
+                    if self.filename:
+                        logger.warning("Duplicate message ignored "
+                                       "in '%s': '%s'" % (self.filename, id))
+                    else:
+                        logger.warning("Duplicate message ignored: '%s'" % id)
             else:
                 if not id:
                     if duplicatestyle == "merge":
                         addcomment(thepo)
-                    else:
+                    elif duplicatestyle == "msgctxt":
                         thepo.setcontext(" ".join(thepo.getlocations()))
                 id_dict[id] = thepo
                 uniqueunits.append(thepo)
@@ -735,7 +741,7 @@ class pofile(pocommon.pofile):
                 return False
         return True
 
-    def parse(self, input):
+    def parse(self, input, duplicatestyle="merge"):
         if hasattr(input, 'name'):
             self.filename = input.name
         elif not getattr(self, 'filename', ''):
@@ -776,6 +782,10 @@ class pofile(pocommon.pofile):
             self.addunit(newunit, new=False)
             newmessage = gpo.po_next_message(self._gpo_message_iterator)
         self._free_iterator()
+
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
+            self.removeduplicates(duplicatestyle=duplicatestyle)
 
     def __del__(self):
         # We currently disable this while we still get segmentation faults.
--- a/translate/storage/fpo.py
+++ b/translate/storage/fpo.py
@@ -29,6 +29,7 @@ directly, but can be used once cpo has been established to work."""
 # - accept only unicodes everywhere
 
 import copy
+import logging
 import re
 import six
 from io import BytesIO
@@ -38,6 +39,8 @@ from translate.misc.multistring import multistring
 from translate.storage import base, cpo, pocommon, poparser
 from translate.storage.pocommon import encodingToUse
 
+
+logger = logging.getLogger(__name__)
 
 lsep = " "
 """Separator for #: entries"""
@@ -488,7 +491,7 @@ class pofile(pocommon.pofile):
             #only add a temporary header
             self._cpo_store.makeheader(charset=self._encoding, encoding="8bit")
 
-    def parse(self, input):
+    def parse(self, input, duplicatestyle="merge"):
         """Parses the given file or file source string."""
         try:
             if hasattr(input, 'name'):
@@ -507,6 +510,9 @@ class pofile(pocommon.pofile):
                 self.units = self.units[1:]
         except Exception as e:
             raise base.ParseError(e)
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
+            self.removeduplicates(duplicatestyle=duplicatestyle)
 
     def removeduplicates(self, duplicatestyle="merge"):
         """Make sure each msgid is unique ; merge comments etc from duplicates into original"""
@@ -541,11 +547,17 @@ class pofile(pocommon.pofile):
                         markedpos.append(thepo)
                     thepo._msgctxt += " ".join(thepo.getlocations())
                     uniqueunits.append(thepo)
+                else:
+                    if self.filename:
+                        logger.warning("Duplicate message ignored "
+                                       "in '%s': '%s'" % (self.filename, id))
+                    else:
+                        logger.warning("Duplicate message ignored: '%s'" % id)
             else:
                 if not id:
                     if duplicatestyle == "merge":
                         addcomment(thepo)
-                    else:
+                    elif duplicatestyle == "msgctxt":
                         thepo._msgctxt += u" ".join(thepo.getlocations())
                 id_dict[id] = thepo
                 uniqueunits.append(thepo)
--- a/translate/storage/pocommon.py
+++ b/translate/storage/pocommon.py
@@ -201,13 +201,13 @@ class pofile(poheader.poheader, base.TranslationStore):
     # We don't want windows line endings on Windows:
     _binary = True
 
-    def __init__(self, inputfile=None, encoding=None):
+    def __init__(self, inputfile=None, encoding=None, duplicatestyle=None):
         super(pofile, self).__init__(unitclass=self.UnitClass)
         self.units = []
         self.filename = ''
         self._encoding = encodingToUse(encoding)
         if inputfile is not None:
-            self.parse(inputfile)
+            self.parse(inputfile, duplicatestyle)
         else:
             self.init_headers()
 
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -24,6 +24,7 @@ files (pofile).
 """
 
 import copy
+import logging
 import re
 import six
 import textwrap
@@ -36,6 +37,8 @@ from translate.misc.multistring import multistring
 from translate.storage import base, pocommon, poparser
 from translate.storage.pocommon import encodingToUse
 
+
+logger = logging.getLogger(__name__)
 
 lsep = "\n#: "
 """Separator for #: entries"""
@@ -758,7 +761,7 @@ class pofile(pocommon.pofile):
     """A .po file containing various units"""
     UnitClass = pounit
 
-    def parse(self, input):
+    def parse(self, input, duplicatestyle="merge"):
         """Parses the given file or file source string."""
         if hasattr(input, 'name'):
             self.filename = input.name
@@ -769,6 +772,10 @@ class pofile(pocommon.pofile):
         # clear units to get rid of automatically generated headers before parsing
         self.units = []
         poparser.parse_units(poparser.ParseState(input, pounit), self)
+
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
+            self.removeduplicates(duplicatestyle=duplicatestyle)
 
     def removeduplicates(self, duplicatestyle="merge"):
         """Make sure each msgid is unique ; merge comments etc from
@@ -804,11 +811,17 @@ class pofile(pocommon.pofile):
                         markedpos.append(thepo)
                     thepo.msgctxt.append('"%s"' % escapeforpo(" ".join(thepo.getlocations())))
                     uniqueunits.append(thepo)
+                else:
+                    if self.filename:
+                        logger.warning("Duplicate message ignored "
+                                       "in '%s': '%s'" % (self.filename, id))
+                    else:
+                        logger.warning("Duplicate message ignored: '%s'" % id)
             else:
                 if not id:
                     if duplicatestyle == "merge":
                         addcomment(thepo)
-                    else:
+                    elif duplicatestyle == "msgctxt":
                         thepo.msgctxt.append('"%s"' % escapeforpo(" ".join(thepo.getlocations())))
                 id_dict[id] = thepo
                 uniqueunits.append(thepo)
--- a/translate/storage/test_cpo.py
+++ b/translate/storage/test_cpo.py
@@ -99,8 +99,7 @@ class TestCPOFile(test_po.TestPOFile):
     def test_merge_duplicates_msgctxt(self):
         """checks that merging duplicates works for msgctxt"""
         posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("msgctxt")
         print(pofile)
         assert len(pofile.units) == 2
@@ -111,8 +110,7 @@ class TestCPOFile(test_po.TestPOFile):
     def test_merge_blanks(self):
         """checks that merging adds msgid_comments to blanks"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
         assert len(pofile.units) == 2
         print(pofile.units[0].msgidcomments)
@@ -124,8 +122,7 @@ class TestCPOFile(test_po.TestPOFile):
     def test_msgid_comment(self):
         """checks that when adding msgid_comments we place them on a newline"""
         posource = '#: source0\nmsgid "Same"\nmsgstr ""\n\n#: source1\nmsgid "Same"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("msgid_comment")
         assert len(pofile.units) == 2
         assert po.unquotefrompo(pofile.units[0].msgidcomments) == "_: source0\n"
@@ -139,8 +136,7 @@ class TestCPOFile(test_po.TestPOFile):
     def test_keep_blanks(self):
         """checks that keeping keeps blanks and doesn't add msgid_comments"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("keep")
         assert len(pofile.units) == 2
         # check we don't add msgidcomments
--- a/translate/storage/test_po.py
+++ b/translate/storage/test_po.py
@@ -171,11 +171,10 @@ class TestPOUnit(test_base.TestTranslationUnit):
 class TestPOFile(test_base.TestTranslationStore):
     StoreClass = po.pofile
 
-    def poparse(self, posource):
+    def poparse(self, posource, duplicatestyle=None):
         """helper that parses po source without requiring files"""
-        dummyfile = wStringIO.StringIO(posource)
-        pofile = self.StoreClass(dummyfile)
-        return pofile
+        return self.StoreClass(wStringIO.StringIO(posource),
+                               duplicatestyle=duplicatestyle)
 
     def poregen(self, posource):
         """helper that converts po source to pofile object and back"""
@@ -589,8 +588,7 @@ msgstr[1] "Koeie"
     def test_merge_duplicates(self):
         """checks that merging duplicates works"""
         posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        #assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
         assert len(pofile.units) == 1
         assert pofile.units[0].getlocations() == ["source1", "source2"]
@@ -608,7 +606,7 @@ msgstr ""
 msgid "test"
 msgstr ""
 '''
-        pofile = self.poparse(posource)
+        pofile = self.poparse(posource, duplicatestyle="allow")
         print(str(pofile))
         pofile.removeduplicates("merge")
         print(str(pofile))
--- a/translate/storage/test_pypo.py
+++ b/translate/storage/test_pypo.py
@@ -209,11 +209,32 @@ class TestPYPOFile(test_po.TestPOFile):
         regenposource = str(pofile)
         assert regenposource.count("_:") == 1
 
+    def test_duplicates_default(self):
+        """checks that duplicates are ignored by default"""
+        posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
+        pofile = self.poparse(posource)
+        assert len(pofile.units) == 1
+        assert str(pofile.units[0]).count("source1") == 1
+        assert str(pofile.units[0]).count("source2") == 0
+
+    def test_duplicates_allow(self):
+        """
+        checks that you can create a pofile with duplicates - this is useful
+        or converting to other formats
+        """
+        posource = ('#: source1\nmsgid "test me"\nmsgstr ""\n\n'
+                    '#: source2\nmsgid "test me"\nmsgstr ""\n')
+        pofile = self.poparse(posource, duplicatestyle="allow")
+        assert len(pofile.units) == 2
+        assert str(pofile.units[0]).count("source1") == 1
+        assert str(pofile.units[0]).count("source2") == 0
+        assert str(pofile.units[1]).count("source1") == 0
+        assert str(pofile.units[1]).count("source2") == 1
+
     def test_merge_duplicates_msgctxt(self):
         """checks that merging duplicates works for msgctxt"""
         posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("msgctxt")
         print(pofile)
         assert len(pofile.units) == 2
@@ -223,8 +244,7 @@ class TestPYPOFile(test_po.TestPOFile):
     def test_merge_blanks(self):
         """checks that merging adds msgid_comments to blanks"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
         assert len(pofile.units) == 2
         print(pofile.units[0].msgidcomments)
@@ -307,7 +327,7 @@ msgstr[0] "toet"
 msgstr[1] "toetse"
 '''
 
-        pofile = self.poparse(posource)
+        pofile = self.poparse(posource, duplicatestyle="allow")
 
         assert pofile.units[1].prev_msgctxt == []
         assert pofile.units[1].prev_source == multistring([u"trea"])


translate/storage/pypo.py
```diff
@@ -759,16 +761,16 @@ class pofile(pocommon.pofile):
...
+        if hasattr(input, 'name'):
+            self.filename = input.name
+        elif not getattr(self, 'filename', ''):
+            self.filename = ''
+        if isinstance(input, str):
+            input = BytesIO(input)
+        # clear units to get rid of automatically generated headers before parsing
+        self.units = []
+        poparser.parse_units(poparser.ParseState(input, pounit), self)
+        self.removeduplicates(duplicatestyle="discard")
```
unho: Please add a commit that cleans that superfluous `if` so we can separate the cleanup for the actual changes.

translate/storage/pypo.py
```diff
@@ -804,6 +806,11 @@ def addcomment(thepo):
                         markedpos.append(thepo)
                     thepo.msgctxt.append('"%s"' % escapeforpo(" ".join(thepo.getlocations())))
                     uniqueunits.append(thepo)
+                else:
+                    filename = ''
+                    if self.filename:
+                        filename = " in '%s'" % self.filename
+                    logger.warn("Duplicate message ignored%s: '%s'" % (filename, id))
```
unho: The following looks less convoluted to me:

```python
                else:
                    if self.filename:
                        logger.warn("Duplicate message ignored in '%s': '%s'" %
                                    (self.filename, id))

                    logger.warn("Duplicate message ignored: '%s'" % id)
```

translate/storage/pypo.py
```diff
@@ -35,6 +36,7 @@
 from translate.storage import base, pocommon, poparser
 from translate.storage.pocommon import encodingToUse
 
+logger = logging.getLogger(__name__)
```
unho: Do add an extra blank line before this.

translate/storage/pypo.py
unho: Shouldn't be `logger.warning()` instead?

translate/storage/fpo.py
```diff
@@ -540,6 +543,11 @@ def addcomment(thepo):
                         markedpos.append(thepo)
                     thepo._msgctxt += " ".join(thepo.getlocations())
                     uniqueunits.append(thepo)
+                else:
+                    filename = ''
+                    if self.filename:
+                        filename = " in '%s'" % self.filename
+                    logger.warn("Duplicate message ignored%s: '%s'" % (filename, id))
```
unho: See comment below about less convoluted syntax and `logger.warning()`.

translate/storage/cpo.py
```diff
@@ -666,6 +666,11 @@ def addcomment(thepo):
                         markedpos.append(thepo)
                     thepo.setcontext(" ".join(thepo.getlocations()))
                     uniqueunits.append(thepo)
+                else:
+                    filename = ''
+                    if self.filename:
+                        filename = " in '%s'" % self.filename
+                    logger.warn("Duplicate message ignored%s: '%s'" % (filename, id))
```
unho: See comment below about using a simpler syntax and using `logger.warning()`.

unho:
I am concerned about that this breaks several tests. Have you considered altering `parse()` so you can control whether duplicates are removed or not?

phlax:
@unho yep i was working on that earlier

I should be ready to push shortly

phlax:
The thing with this though is that it changes the default behaviour.

I think this is correct because it should run removedupes on parse and return valid po by default

phlax:
@unho Ive tidied this up a bit, fixed the tests and added a couple of tests

I have made the default for removeduplicates to discard - we could set this back to merge and still have it run removeduplicates on parse.

translate/convert/test_po2ts.py
```diff
@@ -77,9 +77,7 @@ def test_duplicates(self):
 msgstr "b"
 '''
         tsfile = self.po2ts(posource)
-        print(tsfile)
-        assert tsfile.find("English") != tsfile.rfind("English")
-
```
unho: Please do not remove this blank line. Please do follow PEP8 here.

translate/storage/fpo.py
```diff
@@ -31,12 +31,14 @@
...
 import re
 from io import BytesIO
+import logging
 
 from translate.lang import data
 from translate.misc.multistring import multistring
 from translate.storage import base, cpo, pocommon, poparser
 from translate.storage.pocommon import encodingToUse
 
+logger = logging.getLogger(__name__)
```
unho: Please do add an extra blank line between this and the imports.

translate/storage/test_po.py
```diff
@@ -588,12 +587,11 @@ def test_multiline_obsolete(self):
...
         """checks that merging duplicates works"""
         posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        #assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
         assert len(pofile.units) == 1
         assert pofile.units[0].getlocations() == ["source1", "source2"]
-        print(pofile)
+
```
unho: Please remove this blank line. PEP8 recommends single blank line between methods.

translate/storage/pypo.py
```diff
@@ -804,11 +810,17 @@ def addcomment(thepo):
                         markedpos.append(thepo)
                     thepo.msgctxt.append('"%s"' % escapeforpo(" ".join(thepo.getlocations())))
                     uniqueunits.append(thepo)
+                else:
```
unho: IMHO we should check here for `discard`.

translate/storage/pypo.py
```diff
@@ -757,20 +760,23 @@ class pofile(pocommon.pofile):
...
+            input = BytesIO(input)
+        # clear units to get rid of automatically generated headers before parsing
+        self.units = []
+        poparser.parse_units(poparser.ParseState(input, pounit), self)
+
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
+            self.removeduplicates(duplicatestyle=duplicatestyle)
+
+    def removeduplicates(self, duplicatestyle="discard"):
```
unho: IMHO we should keep the previous default style since that probably will require less changes to TTK.

translate/storage/pypo.py
phlax: the reason i didnt was that anything you put here would discard it anyway just not trigger the warning

translate/storage/pypo.py
```diff
@@ -757,20 +760,23 @@ class pofile(pocommon.pofile):
...
+        elif not getattr(self, 'filename', ''):
+            self.filename = ''
+        if isinstance(input, str):
+            input = BytesIO(input)
+        # clear units to get rid of automatically generated headers before parsing
+        self.units = []
+        poparser.parse_units(poparser.ParseState(input, pounit), self)
+
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
```
unho: Either check against `None` here or use default `allow` in the method parameter above.

phlax: im not really following

we cant have duplicatestyle="allow" as the default in parse or it wouldnt fix the problem - it would revert to old behaviour

unho: Just noticed what a stupid thing is to have `def parse(self, input, duplicatestyle="allow"):` since having mutable objects as default values is highly discouraged. So it should be instead:

```python
        if duplicatestyle is not None and duplicatestyle != "allow":
```


@phlax What I was expecting of this PR is to keep previous behavior by default, but allowing to discard duplicates in case it was explicitly specified when calling `parse()` (necessary for Pootle). I was hoping that this would be a non-invasive change that didn't require to touch existing tests at all, despite it would be necessary to add new tests to test new behavior. Have you tried that approach?

unho: > Just noticed what a stupid thing is to have `def parse(self, input, duplicatestyle="allow"):` since having mutable objects as default values is highly discouraged. So it should be instead:

>```python
        if duplicatestyle is not None and duplicatestyle != "allow":
```

I actually think that this is probably better just at the beginning of `removeduplicates()`.

File: translate/storage/pypo.py (hunk 9 above)
Comment (unho) [inline translate/storage/pypo.py]: ok

**Comment by unho** [comment]
@phlax Can you please rebase and merge the "superfluous if True" commit directly to `master`? Thanks in advance.

File: translate/storage/pypo.py (hunk 11 above)
Comment (phlax) [inline translate/storage/pypo.py]:
```
 if duplicatestyle is not None and duplicatestyle != "allow":
```
looks good for removeduplicates

strings are *immutable*

Comment (phlax) [inline translate/storage/pypo.py]: @unho 

>@phlax What I was expecting of this PR is to keep previous behavior by default, but allowing > to discard duplicates in case it was explicitly specified when calling parse() (necessary for 
> Pootle)

we can do that but that would be more invasive for because when get_class(file)() gets called it would need to have the duplicatestyle param set - therefore all of the storagetypes would need to implement it or we would need to hardcode against po files

**Comment by phlax** [comment]
@unho i dont have push power on upstream master

do you want a pull request?

File: translate/convert/test_po2ts.py (hunk 6 above)
Comment (phlax) [inline translate/convert/test_po2ts.py]: Apologies - removing

File: translate/storage/pypo.py (hunk 10 above)
Comment (phlax) [inline translate/storage/pypo.py]: k - i reset teh default to merge for removeduplicates

this broke 0 tests - which kinda suggests removeduplicates' default needs to be tested...

File: translate/convert/test_po2ts.py [hunk 12]
```diff
@@ -77,8 +77,7 @@ def test_duplicates(self):
 msgstr "b"
 '''
         tsfile = self.po2ts(posource)
-        print(tsfile)
```
unho: What is the reason to remove this?

translate/storage/test_cpo.py
```diff
@@ -111,12 +109,9 @@ def test_merge_duplicates_msgctxt(self):
     def test_merge_blanks(self):
         """checks that merging adds msgid_comments to blanks"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
         assert len(pofile.units) == 2
-        print(pofile.units[0].msgidcomments)
-        print(pofile.units[1].msgidcomments)
```
unho: Please do not remove these print calls. Perhaps they are not necessary, but unless they are strictly necessary to be removed this is a cleanup that doesn't belong to this commit and makes more difficult to review the actual changes.

translate/storage/test_cpo.py
```diff
@@ -139,8 +133,7 @@ def test_msgid_comment(self):
     def test_keep_blanks(self):
         """checks that keeping keeps blanks and doesn't add msgid_comments"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
```
unho: Is there any reason to remove these assert calls?

translate/convert/test_po2ts.py
phlax: i assumed that these were debugging

but, yep, i should have put this cleanup in a separate commit

translate/storage/test_cpo.py
phlax: k, agreed - ill fix

translate/storage/test_cpo.py
phlax: yes - this assert is call is now made in line 138

phlax:
@unho ive removed the cleanup code and tidied up

translate/storage/test_cpo.py
phlax: now in separate commit

translate/convert/test_po2ts.py
phlax: now in separate commit

translate/storage/test_po.py
```diff
@@ -18,7 +18,8 @@ def test_roundtrip_quoting():
     for special in specials:
         quoted_special = pypo.quoteforpo(special)
         unquoted_special = pypo.unquotefrompo(quoted_special)
-        print("special: %r\nquoted: %r\nunquoted: %r\n" % (special, quoted_special, unquoted_special))
+        # print("special: %r\nquoted: %r\nunquoted: %r\n" % (
+        #     special, quoted_special, unquoted_special))
```
unho: This is commented rather than removed.

translate/convert/test_po2ts.py
unho: I think they are for debugging, for when some test fails and you can simply go up and see what the prints did output before the failure. Does this provide some noticeable performance improvement? If yes please move this to a separate PR.

unho:
@phlax In the future **please** do not bring more and more unrelated cleanup commits into a PR. Instead do create new PRs since those cleanups most likely are going to be quickly merged and they actually difficult reviewing the former PR.

translate/storage/cpo.py
```diff
@@ -465,7 +475,7 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
                 self.markfuzzy()
 
     def isheader(self):
-        #return self.source == u"" and self.target != u""
+        # return self.source == u"" and self.target != u""
```
unho: It is better to remove the commented code.

translate/storage/cpo.py
```diff
@@ -492,8 +502,9 @@ def _domarkfuzzy(self, present=True):
         gpo.po_message_set_fuzzy(self._gpo_message, present)
 
     def makeobsolete(self):
-        # FIXME: libgettexpo currently does not reset other data, we probably want to do that
-        # but a better solution would be for libgettextpo to output correct data on serialisation
+        # FIXME: libgettexpo currently does not reset other data, we probably
+        # want to do that but a better solution would be for libgettextpo to
+        # output correct data on serialisation
```
unho: As per our styleguide http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#comments Write sentences correctly: proper capitalisation and punctuation.

translate/storage/cpo.py
```diff
@@ -687,8 +709,8 @@ def addcomment(thepo):
     def __str__(self):
 
         def obsolete_workaround():
-            # Remove all items that are not output by msgmerge when a unit is obsolete.  This is a work
-            # around for bug in libgettextpo
+            # Remove all items that are not output by msgmerge when a unit is
+            # obsolete.  This is a work around for bug in libgettextpo
```
unho: Same comment about comment punctuation as above.

translate/storage/fpo.py
```diff
@@ -32,6 +32,7 @@
 import re
 import six
 from io import BytesIO
+import logging
```
unho: Please place this comment in the proper place. Check our styleguide http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#imports

translate/storage/fpo.py
```diff
@@ -107,13 +110,13 @@ def gettarget(self):
     def settarget(self, target):
         """Sets the msgstr to the given (unescaped) value"""
         self._rich_target = None
-#        assert isinstance(target, unicode)
-#        target = data.forceunicode(target)
+        #        assert isinstance(target, unicode)
+        #        target = data.forceunicode(target)
```
unho: Better drop commented code.

translate/storage/fpo.py
```diff
@@ -207,7 +210,7 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
         """
 
         def mergelists(list1, list2, split=False):
-            #decode where necessary
+            # decode where necessary
```
unho: Proper capitalization please.

translate/storage/cpo.py
```diff
@@ -575,7 +587,6 @@ def buildfromunit(cls, unit, encoding=None):
         elif isinstance(unit, pocommon.pounit):
             newunit = cls(unit.source, encoding)
             newunit.target = unit.target
-            #context
```
unho: Here you are removing the comment while below it is arranged properly. I prefer it to be arranged properly since seems to be communicating something useful.

translate/storage/fpo.py
```diff
@@ -485,10 +488,10 @@ def _build_cpo_from_self(self):
             if not unit.isblank():
                 self._cpo_store.addunit(cpo.pofile.UnitClass.buildfromunit(unit, self._encoding))
         if not self._cpo_store.header():
-            #only add a temporary header
+            # only add a temporary header
```
unho: Proper capitalization please.

translate/storage/fpo.py
```diff
@@ -496,9 +499,9 @@ def parse(self, input):
             elif not getattr(self, 'filename', ''):
                 self.filename = ''
             tmp_header_added = False
-#            if isinstance(input, str) and '"Content-Type: text/plain; charset=' not in input[:200]:
-#                input = basic_header + input
-#                tmp_header_added = True
+            # if isinstance(input, str) and '"Content-Type: text/plain; charset=' not in input[:200]:
+            #     input = basic_header + input
+            #     tmp_header_added = True
```
unho: Better remove commented code.

translate/storage/pocommon.py
```diff
@@ -90,7 +90,10 @@ def geterrors(self):
         return errordict
 
     def markreviewneeded(self, needsreview=True, explanation=None):
-        """Marks the unit to indicate whether it needs review. Adds an optional explanation as a note."""
+        """
+        Marks the unit to indicate whether it needs review. Adds an optional
+explanation as a note.
+        """
```
unho: I think it is preferable to use:

```python
        """Marks the unit to indicate whether it needs review.

        Adds an optional explanation as a note.
        """
```

translate/storage/pocommon.py
```diff
@@ -196,18 +207,19 @@ def encodingToUse(encoding):
 
 class pofile(poheader.poheader, base.TranslationStore):
     Name = "Gettext PO file"  # pylint: disable=E0602
-    Mimetypes = ["text/x-gettext-catalog", "text/x-gettext-translation", "text/x-po", "text/x-pot"]
+    Mimetypes = ["text/x-gettext-catalog", "text/x-gettext-translation",
+                 "text/x-po", "text/x-pot"]
```
unho: From our styleguide: For lists or tuples with many items, break immediately after the opening brace

So:

```python
    Mimetypes = [
        "text/x-gettext-catalog",
        "text/x-gettext-translation",
        "text/x-po",
        "text/x-pot",
    ]
```

translate/storage/pypo.py
```diff
@@ -28,15 +28,18 @@
 import six
 import textwrap
 from io import BytesIO
+import logging
```
unho: Please place import in the right place according to our styleguide.

translate/storage/pypo.py
```diff
@@ -411,7 +424,7 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
         """
 
         def mergelists(list1, list2, split=False):
-            #decode where necessary
+            # decode where necessary
```
unho: Please use proper capitalization.

translate/storage/pypo.py
```diff
@@ -482,8 +498,10 @@ def mergelists(list1, list2, split=False):
                 self.markfuzzy()
 
     def isheader(self):
-        #return (self._msgidlen() == 0) and (self._msgstrlen() > 0) and (len(self.msgidcomments) == 0)
-        #rewritten here for performance:
+        # return (self._msgidlen() == 0)
+        # and (self._msgstrlen() > 0)
+        # and (len(self.msgidcomments) == 0)
```
unho: Remove commented code.

translate/storage/pypo.py
```diff
@@ -482,8 +498,10 @@ def mergelists(list1, list2, split=False):
...
+        # rewritten here for performance:
```
unho: If the commented code is removed then this comment is useless.

translate/storage/pypo.py
```diff
@@ -804,24 +836,32 @@ def addcomment(thepo):
...
                 uniqueunits.append(thepo)
         self.units = uniqueunits
 
     def __str__(self):
-        """Convert to a string. Double check that unicode is handled somehow
-        here"""
+        """
+        Convert to a string.
+        Double check that unicode is handled somehow here
+        """
```
unho: Better:

```python
        """ Convert to a string.

        Double check that unicode is handled somehow here
        """
```

translate/storage/test_po.py
```diff
@@ -435,8 +417,8 @@ def test_uniforum_po(self):
         pofile = self.poparse(posource)
         assert len(pofile.units) == 2
         # FIXME we still need to handle this correctly for proper Uniforum support if required
-        #assert pofile.units[0].getlocations() == "File: somefile, line: 300"
-        #assert pofile.units[1].getlocations() == "File: anotherfile, line: 200"
+        # assert pofile.units[0].getlocations() == "File: somefile, line: 300"
+        # assert pofile.units[1].getlocations() == "File: anotherfile, line: 200"
```
unho: Note that this commented code must be kept since is placed there for future use as previous comment states.

translate/storage/test_po.py
```diff
@@ -537,64 +511,65 @@ def test_makeobsolete_plural(self):
...
 
     def test_makeobsolete_msgctxt(self):
         """Tests making a unit with msgctxt obsolete"""
-        posource = '#: test.c\nmsgctxt "Context"\nmsgid "test"\nmsgstr "rest"\n'
-        poexpected = '#~ msgctxt "Context"\n#~ msgid "test"\n#~ msgstr "rest"\n'
+        posource = ('#: test.c\n'
+                    'msgctxt "Context"\n'
+                    'msgid "test"\nmsgstr "rest"\n')
+        poexpected = ('#~ msgctxt "Context"\n'
+                      '#~ msgid "test"\n#~ msgstr "rest"\n')
```
unho: You seem to have split this into several lines for readability. I like that. But it seems the `msgstr` in `poexpected` is not on its own line.

translate/storage/test_po.py
```diff
@@ -537,64 +511,65 @@ def test_makeobsolete_plural(self):
...
 
     def test_multiline_obsolete(self):
         """Tests for correct output of mulitline obsolete messages"""
-        posource = '#~ msgid ""\n#~ "Old thing\\n"\n#~ "Second old thing"\n#~ msgstr ""\n#~ "Ou ding\\n"\n#~ "Tweede ou ding"\n'
+        posource = ('#~ msgid ""\n'
+                    '#~ "Old thing\\n"\n'
+                    '#~ "Second old thing"\n'
+                    '#~ msgstr ""\n'
+                    '#~ "Ou ding\\n"\n'
+                    '#~ "Tweede ou ding"\n')
```
unho: In other test files the source is put in a triple quoted string, which allows to just copy-paste the real sample because we don't need to escape. Also it is easier to read. I suggest you to rework all these `posource` to use triple quoted strings, like:

```python
        posource = """#~ msgid ""
#~ "Old thing\n"
#~ "Second old thing"
#~ msgstr ""
#~ "Ou ding\n"
#~ "Tweede ou ding"
"""
```

translate/storage/test_po.py
```diff
@@ -749,7 +722,8 @@ def test_id(self):
         # commented out for conformance to gettext.
 #        assert pofile.units[4].getid() == "tree\0trees"
 
-    def test_non_ascii_header_comments(self):
+    # method with same name below?
+    def __test_non_ascii_header_comments(self):
```
unho: Does this test work now that its name starts with `__test` instead of `test`? Anyway I suggest renaming the method by changing what is after `test`.

unho:
@phlax Will review the actual changes re duplicates when the unrelated cleanups are out of this PR.

translate/storage/test_po.py
phlax: yep - but i think ill just remove it

translate/storage/cpo.py
phlax: I didn't write this sentence!

translate/storage/cpo.py
phlax: Same comment as abvoe - re *pep8*

translate/storage/fpo.py
phlax: cool - ill remove

translate/storage/cpo.py
phlax: I dont think it was. I think it was a bit of left over debugging

translate/storage/fpo.py
phlax: k

translate/storage/pocommon.py
phlax: it really pains me to do that - it breaks pep8

claudep: Absolutely not (AFAIK).

phlax: My preferred style is:

```
 Mimetypes = [
       "text/x-gettext-catalog",
       "text/x-gettext-translation",
       "text/x-po",
       "text/x-pot"]
```

translate/storage/pypo.py
phlax: I dont really agree on this one

translate/storage/test_po.py
phlax: i avoid triple quoted strings except at module level

translate/storage/test_po.py
phlax: no it doesnt - but as it wasnt being called already i didnt want to activate it

translate/storage/pocommon.py
phlax: actually your right - its just emacs wont allow it - it forces:

```
Mimetypes = [
    "text/x-gettext-catalog",
    "text/x-gettext-translation",
    "text/x-po",
     "text/x-pot",
     ]

File: translate/storage/fpo.py (hunk 19 above)
Comment (phlax) [inline translate/storage/fpo.py]: logging is in python stdlib - ie in the right place

File: translate/storage/test_po.py (hunk 35 above)
Comment (phlax) [inline translate/storage/test_po.py]: i think with this function we should either:
- remove it altogether, or
- integrate what is there into the test below

**Comment by phlax** [comment]
@unho good to review

File: translate/storage/fpo.py [hunk 36]
```diff
@@ -31,6 +31,7 @@
 import copy
 import re
 import six
+import logging
```
unho: This import must be two lines up. Imports must be alphabetically arranged. See our styleguide on imports.

translate/storage/pypo.py
```diff
@@ -27,6 +27,7 @@
 import re
 import six
 import textwrap
+import logging
```
unho: This import must be put in the right place. See comment about imports being alphabetically ordered.

translate/storage/cpo.py
```diff
@@ -735,7 +741,7 @@ def isempty(self):
                 return False
         return True
 
-    def parse(self, input):
+    def parse(self, input, duplicatestyle=None):
```
unho: Please provide a default value. The same goes for the other two implementations of this method.

unho: Forgot about this.

translate/storage/test_cpo.py
```diff
@@ -99,8 +99,7 @@ def test_msgidcomments(self):
     def test_merge_duplicates_msgctxt(self):
         """checks that merging duplicates works for msgctxt"""
         posource = '#: source1\nmsgid "test me"\nmsgstr ""\n\n#: source2\nmsgid "test me"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("msgctxt")
```
unho: I think it is better to instead do the following:

```python
        pofile = self.poparse(posource, duplicatestyle="msgctxt")
```

translate/storage/test_cpo.py
```diff
@@ -111,8 +110,7 @@ def test_merge_duplicates_msgctxt(self):
     def test_merge_blanks(self):
         """checks that merging adds msgid_comments to blanks"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("merge")
```
unho: I think it is better to instead do the following:

```python
        pofile = self.poparse(posource, duplicatestyle="merge")
```

translate/storage/test_cpo.py
```diff
@@ -124,8 +122,7 @@ def test_merge_blanks(self):
     def test_msgid_comment(self):
         """checks that when adding msgid_comments we place them on a newline"""
         posource = '#: source0\nmsgid "Same"\nmsgstr ""\n\n#: source1\nmsgid "Same"\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("msgid_comment")
```
unho: I think it is better to instead do the following:

```python
        pofile = self.poparse(posource, duplicatestyle="msgid_comment")
```

translate/storage/test_cpo.py
```diff
@@ -139,8 +136,7 @@ def test_msgid_comment(self):
     def test_keep_blanks(self):
         """checks that keeping keeps blanks and doesn't add msgid_comments"""
         posource = '#: source1\nmsgid ""\nmsgstr ""\n\n#: source2\nmsgid ""\nmsgstr ""\n'
-        pofile = self.poparse(posource)
-        assert len(pofile.units) == 2
+        pofile = self.poparse(posource, duplicatestyle="allow")
         pofile.removeduplicates("keep")
```
unho: I think it is better to instead do the following:

```python
        pofile = self.poparse(posource, duplicatestyle="keep")
```

unho: What is the difference between `keep` and `allow`?

translate/storage/pypo.py
```diff
@@ -770,6 +773,10 @@ def parse(self, input):
         self.units = []
         poparser.parse_units(poparser.ParseState(input, pounit), self)
 
+        # duplicates are now removed by default unless duplicatestyle=allow
+        if duplicatestyle != "allow":
+            self.removeduplicates(duplicatestyle=duplicatestyle)
+
     def removeduplicates(self, duplicatestyle="merge"):
```
unho: Can you please expand the docstring to list all the acceptable `duplicatestyle` values?

translate/storage/test_pypo.py
```diff
@@ -209,11 +209,32 @@ def test_combine_msgidcomments(self):
...
+        assert str(pofile.units[0]).count("source1") == 1
+        assert str(pofile.units[0]).count("source2") == 0
+
+    def test_duplicates_allow(self):
+        """
+        checks that you can create a pofile with duplicates - this is useful
+        or converting to other formats
+        """
+        posource = ('#: source1\nmsgid "test me"\nmsgstr ""\n\n'
+                    '#: source2\nmsgid "test me"\nmsgstr ""\n')
```
unho: Please use the following instead which is much more easy to read:

```python
        posource = """#: source1
msgid "test me"
msgstr ""


#: source2
msgid "test me"
msgstr ""
"""
```

Please follow the same pattern in the newly added `posource`.

PR #3291, opened 2015-08-18
Replaced dict.iter* by six.iter*(dict)

claudep:
iteritems/iterkeys/itervalues do not exist in Python 3 as it's now
the default method used for iteration by items()/keys()/values().

### DIFF
--- a/tests/odf_xliff/test_odf_xliff.py
+++ b/tests/odf_xliff/test_odf_xliff.py
@@ -21,6 +21,7 @@
 import difflib
 import os
 import os.path as path
+import six
 import sys
 import zipfile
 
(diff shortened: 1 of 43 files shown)


jleclanche:
You can probably replace some of those `iteritems()` calls by simple `items()` calls.

claudep:
Sure, but it might be a bit tricky to find those where the iteritems call was purposefully done to prevent a memory issue. Worth the effort?

claudep:
I will still see what I can do and add a complementary commit.

claudep:
I added the commit 6f18aefc3 where I *supposed* it was harmless to do so, after a quick glance. If you find them acceptable, I'll merge the commits.

unho:
IMHO replacing `iteritems()` calls by simple `items()` calls is complicating the review and might have unnoticed performance impact. So please drop the second commit.

@claudep Will merge after dropping the second commit and rebase on top of latest master.

claudep:
Done.

unho:
Merged. Thanks!

PR #3292, opened 2015-08-19
Avoided raw unicode strings for Python 3 compatibility

claudep:
Raw unicode strings syntax (e.g. ur'mystring') is not valid syntax on
Python 3.
Read https://www.python.org/dev/peps/pep-0414/#exclusion-of-raw-unicode-literals
This patch uses various techniques to workaround this issue: removing either
the raw or unicode markers when they are not essential, or using unicode_literals
to mimic the Python 3 behavior.

### DIFF
--- a/translate/lang/el.py
+++ b/translate/lang/el.py
@@ -23,6 +23,8 @@
 .. seealso:: http://en.wikipedia.org/wiki/Greek_language
 """
 
+from __future__ import unicode_literals
+
 import re
 
 from translate.lang import common
@@ -32,9 +34,9 @@ class el(common.Common):
     """This class represents Greek."""
 
     # Greek uses ; as question mark and the middot instead
-    sentenceend = u".!;…"
+    sentenceend = ".!;…"
 
-    sentencere = re.compile(ur"""
+    sentencere = re.compile(r"""
         (?s)        # make . also match newlines
         .*?         # anything, but match non-greedy
         [%s]        # the puntuation for sentence ending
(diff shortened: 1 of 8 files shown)


translate/storage/placeables/general.py
```diff
@@ -71,7 +71,7 @@ class NumberPlaceable(Ph):
     """Placeable for numbers."""
 
     istranslatable = False
-    regex = re.compile(ur"[-+]?[0-9]+([\u00a0.,][0-9]+)*")
+    regex = re.compile(u"[-+]?[0-9]+([\u00a0.,][0-9]+)*")
```
unho: Is it not necessary to keep it raw?

translate/storage/placeables/general.py
```diff
@@ -211,7 +211,7 @@ class PunctuationPlaceable(Ph):
     # FIXME this should really be a list created as being the inverse of what
     # is available on the translators keyboard.  Or easily expanded by their
     # configuration.
-    regex = re.compile(ur'''([™©®]|          # Marks
+    regex = re.compile(u'''([™©®]|          # Marks
```
unho: Since a char is being removed here the remaining lines of the strings are no longer aligned. Please remove a whitespace on each of the following lines to keep this aligned.

translate/storage/placeables/general.py
claudep: It depends on the original intent. Apparently, Python resolves unicode escapes, even in raw strings. So if the 'r' prefix was intended to keep the \u00a0 sequence "as is", we should double the backslashes, regardless of the prefix. But if the intent was just to insert a non-breakable space in the string, I think it is fine to remove the prefix.

translate/lang/el.py
```diff
@@ -43,19 +45,19 @@ class el(common.Common):
...
+    valid_latin_accel = "abcdefghijklmnopqrstuvwxyz" + \
+                        "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + \
+                        "1234567890"
 
     # Valid greek characters for use as accelerators (accented characters
     # and "ς" omitted)
-    valid_greek_accel = u"αβγδεζηθικλμνξοπρστυφχψω" + \
-                        u"ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩ"
+    valid_greek_accel = "αβγδεζηθικλμνξοπρστυφχψω" + \
+                        "ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩ"
```
unho: Since these lines are being touched I suggest using the following in order to drop string concatenation and `\` usage:

```python
    valid_greek_accel = ("αβγδεζηθικλμνξοπρστυφχψω"
                         "ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩ")
```

The same above.

translate/storage/placeables/general.py
jleclanche: This is why you're not supposed to align on whitespace :p

translate/storage/placeables/general.py
unho: ok

unho:
Merged. Thanks!

PR #3295, opened 2015-08-20
Removed unreachable KeyboardInterrupt catching

claudep:
`except Exception` is not catching KeyboardInterrupt, so there is no use
examining the exception for it. Only bare excepts are catching it.


unho:
Rebase the PR and will be merged.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3296, opened 2015-08-21
Removed obsolete code in fpo.py

claudep:
This code was deprecated since 2009 (ba8e61b93ef0).

### DIFF
--- a/translate/storage/fpo.py
+++ b/translate/storage/fpo.py
@@ -32,11 +32,10 @@ import copy
 import logging
 import re
 import six
-from io import BytesIO
 
 from translate.lang import data
 from translate.misc.multistring import multistring
-from translate.storage import base, cpo, pocommon, poparser
+from translate.storage import base, cpo, pocommon
 from translate.storage.pocommon import encodingToUse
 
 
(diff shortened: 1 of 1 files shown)


translate/storage/fpo.py
```diff
@@ -329,10 +328,6 @@ def hasplural(self):
         source = self.source
         return isinstance(source, multistring) and len(source.strings) > 1
 
-    def parse(self, src):
-        raise DeprecationWarning("Should not be parsing with a unit")
```
unho: It is safe to remove `DeprecationWarning` as well since it is only used here and in some other commented line.

unho:
Will merge after addressing raised nit or replying with some reason not to do that.

translate/storage/fpo.py
claudep: Remove where?

unho: In https://github.com/claudep/translate/blob/remove_fpo_depr/translate/storage/fpo.py#L291

claudep: Oh, the commented line? Sure, I can.

unho:
If you rebase this PR will be merged.

claudep:
Rebased. I guess the test failures are unrelated to this patch...

unho:
Merged. Thanks a lot!

PR #3297, opened 2015-08-21
Fixed cpo error formatting

claudep:
When testing with USECPO=1, this is fixing a bunch of errors from
the error callback function:
TypeError: cannot concatenate 'str' and 'int' objects


unho:
Will merge after rebase.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3298, opened 2015-08-21
Replaced unicode by six.text_type

claudep:
"unicode" does not exist any longer in Python 3 (as literals are
unicode by default).

### DIFF
--- a/translate/convert/factory.py
+++ b/translate/convert/factory.py
@@ -21,6 +21,7 @@
 """Factory methods to convert supported input files to supported translatable files."""
 
 import os
+import six
 
 
 #from translate.convert import prop2po, po2prop, odf2xliff, xliff2odf
@@ -39,6 +40,7 @@ converters = {}
 #        converters[extension].append(module.formats[extension])
 
 
+@six.python_2_unicode_compatible
 class UnknownExtensionError(Exception):
 
     def __init__(self, afile):
@@ -47,10 +49,8 @@ class UnknownExtensionError(Exception):
     def __str__(self):
         return 'Unable to find extension for file: %s' % (self.file)
 
-    def __unicode__(self):
-        return unicode(str(self))
 
-
+@six.python_2_unicode_compatible
 class UnsupportedConversionError(Exception):
 
     def __init__(self, in_ext=None, out_ext=None, templ_ext=None):
@@ -63,9 +63,6 @@ class UnsupportedConversionError(Exception):
         if self.templ_ext:
             msg += ' with template %s' % (self.templ_ext)
         return msg
-
-    def __unicode__(self):
-        return unicode(str(self))
 
 
 def get_extension(filename):
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -135,6 +135,7 @@ def tagproperties(strings, ignore):
     return properties
 
 
+@six.python_2_unicode_compatible
 class FilterFailure(Exception):
     """This exception signals that a Filter didn't pass, and gives an
     explanation or a comment.
@@ -144,15 +145,12 @@ class FilterFailure(Exception):
         if not isinstance(messages, list):
             messages = [messages]
 
-        assert isinstance(messages[0], unicode)  # Assumption: all of same type
+        assert isinstance(messages[0], six.text_type)  # Assumption: all of same type
 
         self.messages = messages
 
-    def __unicode__(self):
-        return unicode(u", ".join(self.messages))
-
     def __str__(self):
-        return str(u", ".join(self.messages))
+        return u", ".join(self.messages)
 
 
 class SeriousFilterFailure(FilterFailure):
@@ -450,7 +448,7 @@ class UnitChecker(object):
                 filterresult = self.run_test(filterfunction, unit)
             except FilterFailure as e:
                 filterresult = False
-                filtermessage = unicode(e)
+                filtermessage = six.text_type(e)
             except Exception as e:
                 if self.errorhandler is None:
                     raise ValueError("error in filter %s: %r, %r, %s" %
@@ -508,7 +506,7 @@ class TranslationChecker(UnitChecker):
 
             for pluralform in unit.target.strings:
                 try:
-                    if not test(self.str1, unicode(pluralform)):
+                    if not test(self.str1, six.text_type(pluralform)):
                         filterresult = False
                 except FilterFailure as e:
                     filterresult = False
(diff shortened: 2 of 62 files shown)


translate/convert/factory.py
```diff
@@ -64,9 +61,6 @@ def __str__(self):
             msg += ' with template %s' % (self.templ_ext)
         return msg
 
-    def __unicode__(self):
-        return unicode(str(self))
```
unho: By removing these `__unicode__()` methods we are now relying on `__str__()` ones. And those don't seem to convert to Unicode strings on Python 2, so perhaps the output of `__str__()` must also be adjusted to be converted using `six.text_type`.

translate/services/tmserver.py
```diff
@@ -36,7 +36,7 @@ class TMServer(object):
 
     def __init__(self, tmdbfile, tmfiles, max_candidates=3, min_similarity=75,
             max_length=1000, prefix="", source_lang=None, target_lang=None):
-        if not isinstance(tmdbfile, unicode):
+        if not isinstance(tmdbfile, six.text_type):
```
unho: `six` is not imported in this module.

translate/storage/fpo.py
```diff
@@ -210,7 +210,7 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
 
         def mergelists(list1, list2, split=False):
             #decode where necessary
-            if unicode in [type(item) for item in list2] + [type(item) for item in list1]:
+            if str in [type(item) for item in list2] + [type(item) for item in list1]:
```
unho: Why is `str` used here instead of `six.text_type`?

translate/convert/factory.py
unho: Just wondering if these should be adjusted to use `@six.python_2_unicode_compatible` as done below.

translate/storage/pypo.py
```diff
@@ -415,7 +415,7 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
 
         def mergelists(list1, list2, split=False):
             #decode where necessary
-            if unicode in [type(item) for item in list2] + [type(item) for item in list1]:
+            if str in [type(item) for item in list2] + [type(item) for item in list1]:
```
unho: As above, why not use `six.text_type`?

translate/storage/test_tiki.py
```diff
@@ -4,6 +4,8 @@
 # tiki unit tests
 # Author: Wil Clouser <wclouser@mozilla.com>
 # Date: 2008-12-01
+import six
```
unho: Please add a blank line before the imports.

translate/storage/tiki.py
```diff
@@ -66,7 +68,7 @@ def __init__(self, source=None, encoding="UTF-8"):
         self.location = []
         super(TikiUnit, self).__init__(source)
 
-    def __unicode__(self):
+    def __str__(self):
```
unho: I am a bit worried that this effectively overrides the `__str__()` method of `base.TranslationUnit`. I would be ok with this if we can tell for sure we don't need the base implementation anywhere, but I am afraid that is a long shot.

translate/convert/factory.py
claudep: I don't think that having unicode output is so much important for exceptions, but yes, I can use `@six.python_2_unicode_compatible` to prevent any regression.

translate/storage/fpo.py
claudep: I thought that `unicode` was a mistake, because the following code is making all non-unicode list elements unicode.
I understood it as: if the lists contain any *non* unicode element, render them unicode. But now I see that it might also be: if at least one element is unicode, make all others unicode too, else let them bytestrings. Could you confirm the latter is the intented behavior?

translate/storage/tiki.py
claudep: If you are looking at the `TikiStore.__str__` implementation below, you'll see that it effectively calls `unicode()` (six.text_type()) on units to include them in the final file. With the `python_2_unicode_compatible` decorator, the correct method should be called in all cases.

translate/convert/factory.py
unho: Please do.

translate/storage/tiki.py
unho: ok

translate/storage/fpo.py
unho: First of all, I didn't write this code and my assumptions might be mistaken.

Reading the code within the `if` my understanding is that if any unicode string was present on any of the lists, then it converts all the other non-unicode strings to unicode. This makes sense to me since when you are merging together two files, one of which is unicode and the other not, you want them to be in the same encoding to merge them, and unicode feels like the right encoding.

@dwaynebailey Perhaps you can shed some light here.

claudep: I've just pushed a new fixed/rebased version and used the six.text_type here (keeping old behavior).

translate/services/tmserver.py
unho: Now it is imported.

translate/storage/fpo.py
```diff
@@ -209,8 +209,8 @@ def merge(self, otherpo, overwrite=False, comments=True, authoritative=False):
         """
 
         def mergelists(list1, list2, split=False):
-            #decode where necessary
-            if unicode in [type(item) for item in list2] + [type(item) for item in list1]:
+            # Decode where necessary (either all bytestrings or all unicode)
+            if six.text_type in [type(item) for item in list2] + [type(item) for item in list1]:
                 for position, item in enumerate(list1):
                     if isinstance(item, str):
```
unho: This `str` appearance here should be `bytes`, isn't it?

translate/convert/factory.py
unho: For the record this is now fixed.

translate/storage/fpo.py
claudep: Absolutely, but I have a separate patch locally which does the str/bytes replacement in all the code.

claudep: See 0690e5017993e6a9c96875d7fbf2f5d5c1cf248f

unho: It is fine then.

translate/filters/checks.py
```diff
@@ -144,13 +144,10 @@ def __init__(self, messages):
...
-        assert isinstance(messages[0], unicode)  # Assumption: all of same type
+        assert isinstance(messages[0], six.text_type)  # Assumption: all of same type
 
         self.messages = messages
 
-    def __unicode__(self):
-        return unicode(u", ".join(self.messages))
-
     def __str__(self):
         return str(u", ".join(self.messages))
```
unho: Shouldn't this use `six.text_type()` instead?

claudep: I think it would be better to use the same strategy as in `translate/convert/factory.py` and use the 
python_2_unicode_compatible decorator. I will update the patch.

unho:
Merged. Thanks!

PR #3301, opened 2015-08-22
Fixed wrong usage of importorskip

### LINKED ISSUE #3206: incorrect usage of importorskip in storage/test_cpo.py
My understanding of the pytest documentation is that `importorskip` accepts a module name and not any arbitrary expression to be evaluated. In `storage/test_cpo.py`, the intention of:

```python
cpo = importorskip("not sys.platform.startswith('linux')")
```
is to only run the tests on linux, but the tests do not run on linux either:

```
SKIP [1] /«BUILDDIR»/translate-toolkit-1.12.0+dfsg1/translate/storage/test_cpo.py:7: could not import "not sys.platform.startswith('linux')"
```

Looking more closely:

```
In [1]: import sys
In [2]: sys.platform
Out[2]: 'linux2'
In [3]: from pytest import importorskip, mark, raises
In [4]: cpo = importorskip("not sys.platform.startswith('linux')")
---------------------------------------------------------------------------
Skipped                                   Traceback (most recent call last)
Skipped: could not import "not sys.platform.startswith('linux')"
```

I think the following (untested!) code is more in line with what was actually intended:

```python
import pytest
if not sys.platform.startswith('linux'):
    pytest.skip("cpo can only be tested on linux")
```

claudep:
Fixes #3206. Thanks Stuart Prescott for the report.

### DIFF
--- a/translate/storage/cpo.py
+++ b/translate/storage/cpo.py
@@ -608,7 +608,7 @@ class pounit(pocommon.pounit):
 class pofile(pocommon.pofile):
     UnitClass = pounit
 
-    def __init__(self, inputfile=None, encoding=None, unitclass=pounit, noheader=False):
+    def __init__(self, inputfile=None, encoding=None, unitclass=pounit, noheader=False, **kwargs):
         self._gpo_memory_file = None
         self._gpo_message_iterator = None
         self.units = []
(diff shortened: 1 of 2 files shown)


unho:
Merged. Thanks!

PR #3302, opened 2015-08-23
Reviewed/modernized open() calls

claudep:
By generalizing the `with open...` syntax, we also ensure files are
properly closed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -358,17 +358,16 @@ def parse_requirements(file_name):
     Copied from cburgmer/pdfserver.
     """
     requirements = []
-    for line in open(file_name, 'r').read().split('\n'):
-        # Ignore comments, blank lines and included requirements files
-        if re.match(r'(\s*#)|(\s*$)|(-r .*$)', line):
-            continue
+    with open(file_name, 'r') as fh:
+        for line in fh:
+            # Ignore comments, blank lines and included requirements files
+            if re.match(r'(\s*#)|(\s*$)|(-r .*$)', line):
+                continue
 
-        if re.match(r'\s*-e\s+', line):
-            requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
-        elif re.match(r'\s*-f\s+', line):
-            pass
-        else:
-            requirements.append(line)
+            if re.match(r'\s*-e\s+', line):
+                requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
+            elif not re.match(r'\s*-f\s+', line):
+                requirements.append(line.rstrip('\n'))
 
     return requirements
 
@@ -419,9 +418,8 @@ def buildmanifest_in(f, scripts):
 def standardsetup(name, version, custompackages=[], customdatafiles=[]):
     # TODO: make these end with .py ending on Windows...
     try:
-        manifest_in = open("MANIFEST.in", "w")
-        buildmanifest_in(manifest_in, translatescripts + translatebashscripts)
-        manifest_in.close()
+        with open("MANIFEST.in", "w") as manifest_in:
+            buildmanifest_in(manifest_in, translatescripts + translatebashscripts)
     except IOError as e:
         sys.stderr.write("warning: could not recreate MANIFEST.in, continuing anyway. (%s)\n" % e)
 
--- a/translate/convert/odf2xliff.py
+++ b/translate/convert/odf2xliff.py
@@ -36,11 +36,6 @@ from translate.storage.xml_extract.extract import ParseState, build_store
 
 def convertodf(inputfile, outputfile, templates):
     """Convert an ODF package to XLIFF."""
-    # Since the convertoptionsparser will give us an open file, we risk that
-    # it could have been opened in non-binary mode on Windows, and then we'll
-    # have problems, so let's make sure we have what we want.
-    inputfile.close()
-    inputfile = file(inputfile.name, mode='rb')
 
     store = factory.getobject(outputfile)
 
--- a/translate/convert/po2wordfast.py
+++ b/translate/convert/po2wordfast.py
@@ -85,9 +85,9 @@ class WfOptionParser(convert.ArchiveConvertOptionParser):
         if not options.targetlanguage:
             raise ValueError("You must specify the target language")
         super(WfOptionParser, self).recursiveprocess(options)
-        self.output = open(options.output, 'w')
-        #options.outputarchive.wffile.setsourcelanguage(options.sourcelanguage)
-        self.output.write(str(options.outputarchive.wffile))
+        with open(options.output, 'wb') as self.output:
+            #options.outputarchive.wffile.setsourcelanguage(options.sourcelanguage)
+            self.output.write(str(options.outputarchive.wffile))
 
 
 def main(argv=None):
--- a/translate/misc/optrecurse.py
+++ b/translate/misc/optrecurse.py
@@ -523,13 +523,13 @@ class RecursiveOptionParser(optparse.OptionParser, object):
         """Opens the input file."""
         if fullinputpath is None:
             return sys.stdin
-        return open(fullinputpath, 'r')
+        return open(fullinputpath, 'rb')
 
     def openoutputfile(self, options, fulloutputpath):
         """Opens the output file."""
         if fulloutputpath is None:
             return sys.stdout
-        return open(fulloutputpath, 'w')
+        return open(fulloutputpath, 'wb')
 
     def opentempoutputfile(self, options, fulloutputpath):
         """Opens a temporary output file."""
--- a/translate/storage/cpo.py
+++ b/translate/storage/cpo.py
@@ -717,8 +717,9 @@ class pofile(pocommon.pofile):
 
         def writefile(filename):
             self._gpo_memory_file = gpo.po_file_write_v2(self._gpo_memory_file, filename, xerror_handler)
-            with open(filename) as tfile:
-                return tfile.read()
+            with open(filename, 'rb') as tfile:
+                content = tfile.read()
+            return content
 
         outputstring = ""
         if self._gpo_memory_file:
--- a/translate/storage/factory.py
+++ b/translate/storage/factory.py
@@ -63,14 +63,14 @@ _ext is a pseudo extension, that is their is no real extension by that name.
 def _examine_txt(storefile):
     """Determine the true filetype for a .txt file"""
     if isinstance(storefile, six.string_types) and os.path.exists(storefile):
-        storefile = open(storefile)
+        storefile = open(storefile, 'rb')
     try:
         start = storefile.read(600).strip()
     except AttributeError:
         raise ValueError("Need to read object to determine type")
     # Some encoding magic for Wordfast
     from translate.storage import wordfast
-    if wordfast.TAB_UTF16 in start.split("\n")[0]:
+    if wordfast.TAB_UTF16 in start.split(b"\n")[0]:
         encoding = 'utf-16'
     else:
         encoding = 'iso-8859-1'
--- a/translate/storage/mo.py
+++ b/translate/storage/mo.py
@@ -54,10 +54,9 @@ MO_MAGIC_NUMBER = 0x950412de
 
 def mounpack(filename='messages.mo'):
     """Helper to unpack Gettext MO files into a Python string"""
-    f = open(filename)
-    s = f.read()
-    print("\\x%02x" * len(s) % tuple(map(ord, s)))
-    f.close()
+    with open(filename, 'rb') as fh:
+        s = fh.read()
+        print("\\x%02x" * len(s) % tuple(map(ord, s)))
 
 
 def my_swap4(result):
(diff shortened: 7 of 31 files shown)


setup.py
```diff
@@ -358,17 +358,18 @@ def parse_requirements(file_name):
...
+            # Ignore comments, blank lines and included requirements files
+            if re.match(r'(\s*#)|(\s*$)|(-r .*$)', line):
+                continue
+
+            if re.match(r'\s*-e\s+', line):
+                requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
+            elif re.match(r'\s*-f\s+', line):
+                pass
+            else:
+                requirements.append(line.rstrip('\n'))
```
unho: Is not simpler to drop the `pass` branch?

```python
            if re.match(r'\s*-e\s+', line):
                requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
            elif not re.match(r'\s*-f\s+', line):
                requirements.append(line.rstrip('\n'))
```

translate/convert/odf2xliff.py
```diff
@@ -36,11 +36,6 @@
 
 def convertodf(inputfile, outputfile, templates):
     """Convert an ODF package to XLIFF."""
-    # Since the convertoptionsparser will give us an open file, we risk that
-    # it could have been opened in non-binary mode on Windows, and then we'll
-    # have problems, so let's make sure we have what we want.
-    inputfile.close()
-    inputfile = file(inputfile.name, mode='rb')
```
unho: What is the reason to remove this given what says the comment?

translate/convert/po2wordfast.py
```diff
@@ -85,9 +85,9 @@ def recursiveprocess(self, options):
         if not options.targetlanguage:
             raise ValueError("You must specify the target language")
         super(WfOptionParser, self).recursiveprocess(options)
-        self.output = open(options.output, 'w')
-        #options.outputarchive.wffile.setsourcelanguage(options.sourcelanguage)
-        self.output.write(str(options.outputarchive.wffile))
+        with open(options.output, 'wb') as self.output:
```
unho: Why use `wb` when the previous code was using `w`?

translate/misc/optrecurse.py
```diff
@@ -523,13 +523,13 @@ def openinputfile(self, options, fullinputpath):
...
             return sys.stdin
-        return open(fullinputpath, 'r')
+        return open(fullinputpath, 'rb')
 
     def openoutputfile(self, options, fulloutputpath):
         """Opens the output file."""
         if fulloutputpath is None:
             return sys.stdout
-        return open(fulloutputpath, 'w')
+        return open(fulloutputpath, 'wb')
```
unho: What is the reason to use `b` in this module opens?

translate/storage/cpo.py
```diff
@@ -717,8 +717,9 @@ def obsolete_workaround():
 
         def writefile(filename):
             self._gpo_memory_file = gpo.po_file_write_v2(self._gpo_memory_file, filename, xerror_handler)
-            with open(filename) as tfile:
-                return tfile.read()
+            with open(filename, 'rb') as tfile:
```
unho: Why `rb`?

translate/storage/factory.py
```diff
@@ -63,18 +63,18 @@
 def _examine_txt(storefile):
     """Determine the true filetype for a .txt file"""
     if isinstance(storefile, six.string_types) and os.path.exists(storefile):
-        storefile = open(storefile)
+        storefile = open(storefile, 'rb')
```
unho: Why `b`?

translate/storage/factory.py
```diff
@@ -63,18 +63,18 @@
...
         raise ValueError("Need to read object to determine type")
     # Some encoding magic for Wordfast
     from translate.storage import wordfast
-    if wordfast.TAB_UTF16 in start.split("\n")[0]:
+    if wordfast.TAB_UTF16 in start.split(b"\n")[0]:
         encoding = 'utf-16'
     else:
         encoding = 'iso-8859-1'
-    start = start.decode(encoding).encode('utf-8')
+    start = start.decode(encoding)
```
unho: These cleanups seem unrelated to `open`. If they are please put them in a separate commit.

translate/storage/mo.py
```diff
@@ -53,10 +53,9 @@
 
 def mounpack(filename='messages.mo'):
     """Helper to unpack Gettext MO files into a Python string"""
-    f = open(filename)
-    s = f.read()
-    print("\\x%02x" * len(s) % tuple(map(ord, s)))
-    f.close()
+    with open(filename, 'rb') as fh:
```
unho: Why `b`?

translate/storage/project.py
```diff
@@ -186,7 +186,8 @@ def export_file(self, fname, destfname):
             via the call to
             :meth:`~translate.storage.projstore.ProjectStore.get_file`
             if *fname* is not found in the project."""
-        open(destfname, 'w').write(self.store.get_file(fname).read())
+        with open(destfname, 'wb') as fp:
```
unho: Why `b`?

translate/storage/qm.py
```diff
@@ -73,10 +73,9 @@
 
 def qmunpack(file_='messages.qm'):
     """Helper to unpack Qt .qm files into a Python string"""
-    f = open(file_)
-    s = f.read()
-    print("\\x%02x" * len(s) % tuple(map(ord, s)))
-    f.close()
+    with open(file_, 'rb') as fh:
```
unho: Why `b`?

translate/storage/test_mo.py
```diff
@@ -156,17 +155,14 @@ def test_output(self):
                 # can skip the checks here.
                 continue
 
-            mo_msgfmt_f = open(MO_MSGFMT)
-            mo_pocompile_f = open(MO_POCOMPILE)
-
-            try:
+            with open(MO_MSGFMT, 'rb') as mo_msgfmt_f:
```
unho: Why `b`?

translate/storage/test_mo.py
```diff
@@ -156,17 +155,14 @@ def test_output(self):
...
                 mo_msgfmt = mo_msgfmt_f.read()
-                print("msgfmt output:")
-                print(repr(mo_msgfmt))
+            print("msgfmt output:")
+            print(repr(mo_msgfmt))
+
+            with open(MO_POCOMPILE, 'rb') as mo_pocompile_f:
```
unho: Why `b`?

translate/storage/test_mo.py
```diff
@@ -156,17 +155,14 @@ def test_output(self):
...
-                print("pocompile output:")
-                print(repr(mo_pocompile))
-                assert mo_msgfmt == mo_pocompile
-            finally:
-                mo_msgfmt_f.close()
-                mo_pocompile_f.close()
+            print("pocompile output:")
+            print(repr(mo_pocompile))
+
+            assert mo_msgfmt == mo_pocompile
```
unho: Please note that this must be only run if the two files have been successfully opened and read.

translate/storage/test_mo.py
```diff
@@ -156,17 +155,14 @@ def test_output(self):
...
+            with open(MO_POCOMPILE, 'rb') as mo_pocompile_f:
                 mo_pocompile = mo_pocompile_f.read()
-                print("pocompile output:")
-                print(repr(mo_pocompile))
-                assert mo_msgfmt == mo_pocompile
-            finally:
-                mo_msgfmt_f.close()
-                mo_pocompile_f.close()
+            print("pocompile output:")
+            print(repr(mo_pocompile))
```
unho: Please put within the `with` block.

translate/storage/test_mo.py
```diff
@@ -156,17 +155,14 @@ def test_output(self):
...
-            mo_msgfmt_f = open(MO_MSGFMT)
-            mo_pocompile_f = open(MO_POCOMPILE)
-
-            try:
+            with open(MO_MSGFMT, 'rb') as mo_msgfmt_f:
                 mo_msgfmt = mo_msgfmt_f.read()
-                print("msgfmt output:")
-                print(repr(mo_msgfmt))
+            print("msgfmt output:")
+            print(repr(mo_msgfmt))
```
unho: Please put within the `with` block.

translate/tools/poconflicts.py
```diff
@@ -180,7 +180,8 @@ def str_len(x):
             for target, unit, filename in translations:
                 unit.othercomments.append("# (poconflicts) %s\n" % filename)
                 conflictfile.units.append(unit)
-            open(fulloutputpath, "w").write(str(conflictfile))
+            with open(fulloutputpath, "wb") as fh:
```
unho: Why `b`?

translate/tools/porestructure.py
```diff
@@ -101,13 +101,13 @@ def processfile(self, options, fullinputpath):
                     self.checkoutputsubdir(options, os.path.dirname(outputpath))
                     fulloutputpath = os.path.join(options.output, outputpath)
                     if os.path.isfile(fulloutputpath):
-                        outputfile = open(fulloutputpath, 'r')
+                        outputfile = open(fulloutputpath, 'rb')
```
unho: Why `b`?

translate/tools/porestructure.py
```diff
@@ -101,13 +101,13 @@ def processfile(self, options, fullinputpath):
...
                         outputpofile = po.pofile(outputfile)
                     else:
                         outputpofile = po.pofile()
                     outputpofile.units.append(pounit)   # TODO:perhaps check to see if it's already there...
-                    outputfile = open(fulloutputpath, 'w')
-                    outputfile.write(str(outputpofile))
+                    with open(fulloutputpath, 'wb') as fh:
```
unho: Why `b`?

translate/tools/poterminology.py
```diff
@@ -450,7 +449,8 @@ def outputterminology(self, options):
         termitems = self.extractor.filter_terms(terms, nonstopmin=options.nonstopmin, sortorders=options.sortorders)
         for count, unit in termitems:
             termfile.units.append(unit)
-        open(options.output, "w").write(str(termfile))
+        with open(options.output, "wb") as fh:
```
unho: Why `b`?

claudep:
I owe you an explanation about widen usage of the binary mode. With Python 2, adding a 'b' mode to opening/writing files has little effect, except on Windows. With Python 3, the default modes ('r', 'w') are a text mode where Python tries do encode/decode using a default encoding (locale.getpreferredencoding()). Read https://docs.python.org/3.4/library/functions.html#open. I don't think that's what we want with ttk. At several places, we are sniffing the files to detect the encoding, so we can decode the strings using the proper encoding.

unho:
> I owe you an explanation about widen usage of the binary mode. With Python 2, adding a 'b' mode to opening/writing files has little effect, except on Windows. With Python 3, the default modes ('r', 'w') are a text mode where Python tries do encode/decode using a default encoding (locale.getpreferredencoding()). Read https://docs.python.org/3.4/library/functions.html#open. I don't think that's what we want with ttk. At several places, we are sniffing the files to detect the encoding, so we can decode the strings using the proper encoding.

Tests are passing which is a good sign. But obviously we are only testing on Linux with Python 2 and I am worried about if TTK would be work OK on Windows with these changes. It is not a big concern for me since I had trouble setting up TTK on Windows and I don't think this would impact a big number of users.

I am more worried about changing behavior for Python 3, but since that is our final goal I presume you are testing with Python 3 and it works. If it does I am OK with merging the PR provided that other issues are addressed.

There are other not related to `b` usage to be addressed in this PR, so please do.

setup.py
claudep: Technically yes, but I think the current code is a bit simpler to read. But if you prefer this syntax, just confirm you'd like this to be changed.

translate/convert/odf2xliff.py
claudep: As noted elsewhere in the patch, we are now systematically opening input files in 'rb' mode.

translate/storage/factory.py
claudep: ok, I will delay this change to a further commit.

translate/storage/test_mo.py
claudep: I don't understand why. IMHO, the file can be closed as soon as the content have been read.

translate/storage/test_mo.py
claudep: Sure, but I don't see what is different than before at this regard.

setup.py
unho: I prefer this to be changed. That `pass` branch does not help understanding the code in any way.

translate/storage/test_mo.py
unho: What if the file is not opened at all? We would be trying to output something that wasn't parsed at all.

translate/storage/test_mo.py
unho: You are right.

translate/storage/test_mo.py
unho: Sorry, you are right. Somehow I was mixing different concepts on my head. My apologies.

translate/convert/odf2xliff.py
unho: ok

unho:
Will merge after last concerns are addressed and branch is rebased.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3304, opened 2015-08-24
Checks: Python braces not needed for LibreOffice

dwaynebailey:
Exclusion of checks should be easier than this.  But for now do it this way.


unho:
gtm

PR #3305, opened 2015-08-24
Replaced str isinstance tests by bytes

claudep:


claudep:
Python 3 status before/after:
=========== 1356 failed, 438 passed, 16 skipped, 54 xfailed, 251 error in 123.60 seconds ===========
=========== 1299 failed, 564 passed, 16 skipped, 54 xfailed, 182 error in 165.01 seconds ===========

unho:
Please include an explanation in the commit message saying that Python 2 `str` is `bytes` in Python 3, and that a `bytes` alias was added in Python 2.6 and 2.7, so in short it is safe to just use `bytes`.

Will merge after commit message adjust and PR rebase.

unho:
> Python 3 status before/after:
=========== 1356 failed, 438 passed, 16 skipped, 54 xfailed, 251 error in 123.60 seconds ===========
=========== 1299 failed, 564 passed, 16 skipped, 54 xfailed, 182 error in 165.01 seconds ===========

That is still a loooooot of failing tests.

claudep:
Sure, but knowing that at the start of this work, tests weren't even collected, that's already a good step forward!
Rebased.

unho:
Merged. Thanks!

PR #3306, opened 2015-08-24
Moved imports for Python 3 (unichr, reduce, xrange)

claudep:


translate/lang/ngram.py
```diff
@@ -131,7 +132,7 @@ def classify(self, text):
         ngram = _NGram(text)
         r = 'guess'
 
-        min = sys.maxint
+        min = six.MAXSIZE
```
unho: Note that `six.MAXSIZE` is not a replacement for `sys.maxint` https://pythonhosted.org/six/#six.MAXSIZE After reading the code where `min` is used I suppose this works fine.

unho:
Will merge after rebase.

translate/lang/ngram.py
claudep: I suppose so, too.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3307, opened 2015-08-24
Transformed iterators to lists when needed

claudep:
Several constructs in Python 3 do not return lists any longer, but
iterators instead, like filter(), map(), .keys(), .values(), etc.
When iterating, this is not a problem, but where we need a list,
the iterator has to be consumed in a list.


unho:
Will merge after rebase.

claudep:
I think it is up-to-date. I had rebase already.

unho:
Merged. Thanks!

PR #3309, opened 2015-08-24
Made unit classes encoding agnostic

claudep:
The leading idea is that the file classes should know about
their encoding for serialization, but not the units.

### DIFF
--- a/translate/storage/csvl10n.py
+++ b/translate/storage/csvl10n.py
@@ -117,6 +117,7 @@ def to_unicode(text, encoding='utf-8'):
     return text.decode(encoding)
 
 
+@six.python_2_unicode_compatible
 class csvunit(base.TranslationUnit):
     spreadsheetescapes = [("+", "\\+"), ("-", "\\-"), ("=", "\\="), ("'", "\\'")]
 
@@ -257,20 +258,20 @@ class csvunit(base.TranslationUnit):
 
         #self.source, self.target = self.remove_spreadsheet_escapes(self.source, self.target)
 
-    def todict(self, encoding='utf-8'):
+    def todict(self, **kwargs):
         #FIXME: use apis?
         #source, target = self.add_spreadsheet_escapes(self.source, self.target)
         source = self.source
         target = self.target
         output = {
-            'location': from_unicode(self.location, encoding),
-            'source': from_unicode(source, encoding),
-            'target': from_unicode(target, encoding),
-            'id': from_unicode(self.id, encoding),
+            'location': self.location,
+            'source': source,
+            'target': target,
+            'id': self.id,
             'fuzzy': str(self.fuzzy),
-            'context': from_unicode(self.context, encoding),
-            'translator_comments': from_unicode(self.translator_comments, encoding),
-            'developer_comments': from_unicode(self.developer_comments, encoding),
+            'context': self.context,
+            'translator_comments': self.translator_comments,
+            'developer_comments': self.developer_comments,
         }
 
         return output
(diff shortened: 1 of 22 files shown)


claudep:
This is not really a mandatory step for the Python 3 port, but while working on it, I realized it will simplify things a bit, because many Python 3 test failures are related to encodings. It seems to me that it might be a good design to let the encoding stuff to file classes, not units.

dwaynebailey:
+1 from me.  I can't think of a case where a unit would have an encoding different from the file, well at least not a case that isn't broken.

translate/storage/properties.py
```diff
@@ -478,11 +479,8 @@ def encoding(self):
             return self.personality.default_encoding
 
     def __str__(self):
-        """Convert to a string. Double check that unicode is handled
-        somehow here."""
-        source = self.getoutput()
-        assert isinstance(source, six.text_type)
-        return source.encode(self.encoding)
```
unho: In this particular format `self.encoding` might be referring to the personality encoding. See `@encoding` a few lines below.

unho: Anyway the `assert` can be removed.

translate/storage/lisa.py
```diff
@@ -245,7 +248,7 @@ def getNodeText(self, languageNode, xml_space="preserve"):
 
     def __str__(self):
         return etree.tostring(self.xmlelement, pretty_print=True,
-                              encoding='utf-8')
+                              encoding='unicode')
```
unho: Why the switch from `utf-8` to `unicode`?

translate/storage/csvl10n.py
```diff
@@ -263,14 +264,14 @@ def todict(self, encoding='utf-8'):
         source = self.source
         target = self.target
         output = {
-            'location': from_unicode(self.location, encoding),
-            'source': from_unicode(source, encoding),
-            'target': from_unicode(target, encoding),
-            'id': from_unicode(self.id, encoding),
```
unho: This method still accepts an encoding argument that is no longer used.

translate/storage/properties.py
claudep: The point is that I'd like to not having to encode at all at unit level, if possible. But your comment made me see some other .encode() occurrences in this class, which I'd like to also remove (to be tested).

translate/storage/lisa.py
claudep: Because setting the encoding to 'unicode' keeps the output as an unicode string. On Python 3, `__str__` methods are required to return unicode content (not utf-8-encoded bytestrings).

translate/storage/csvl10n.py
claudep: I hesitated to remove it. Should we keep backwards compatibility to such APIs? An alternative would be to replace it with **kwargs.

translate/storage/properties.py
unho: A minor correction, `@encoding` is actually just above this `__str__()`.

Yes, there are more `encoding` and `encode` on the unit.

translate/storage/csvl10n.py
unho: Yes, sorry, I meant to replace it with `**kwargs`.

translate/storage/lisa.py
unho: Can you please provide some link that allows to get more information on this?

claudep: Sure: http://lxml.de/api/lxml.etree-module.html#tostring
And on Python 3 `__str__`: https://docs.python.org/3/reference/datamodel.html?highlight=__str__#object.__str__

unho: Thanks a lot!

translate/storage/properties.py
claudep: I've inspected other encode calls, but saw that personality.encode() is related to formatting, not the unicode to bytestring process that str.encode() does. Therefore, I'll stick with the current version of this patch in this file.

unho:
Merged. Thanks!

PR #3311, opened 2015-08-25
Fixed ordering issue in Greek punctuation conversion

claudep:
As recent Python version have hash randomization by default,
we had random failure with Greek test_punctranslate because depending
on the order of the dict, replacement could follow: "?" -> ";" -> "·".


unho:
Will merge after rebase.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3312, opened 2015-08-25
Replaced cmp() by new sorting constructs

claudep:
Also added basic tests for poterminology so some of the changed cmp functions are covered by tests.


unho:
Merged. Thanks!

PR #3314, opened 2015-08-26
Use binary mode to open files in base storage

claudep:
This is in line with commit 2f889c19e2.
Also marked with b/bytes some more strings.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -782,10 +782,7 @@ class TranslationStore(object):
         """Write the string representation to the given file (or filename)."""
         storestring = str(self)
         if isinstance(storefile, six.string_types):
-            mode = 'w'
-            if self._binary:
-                mode = 'wb'
-            storefile = open(storefile, mode)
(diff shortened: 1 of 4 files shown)


translate/storage/base.py
```diff
@@ -795,13 +792,10 @@ def save(self):
         """Save to the file that data was originally read from, if
         available."""
         fileobj = getattr(self, "fileobj", None)
-        mode = 'w'
-        if self._binary:
-            mode = 'wb'
+        mode = 'wb'
```
unho: It is preferable to avoid creating variables that are not strictly necessary.

unho: Also having the mode string in the `open` call (or wherever the mode is used) makes it easier to understand the code.

unho:
Next time please try to keep these changes in separate commits.

claudep:
Rebased and commit separated.

unho:
Merged. Thanks!

PR #3315, opened 2015-08-26
Adapted poparsing encoding strategy

claudep:
Try to convert to unicode as soon as possible.

### DIFF
--- a/translate/storage/poparser.py
+++ b/translate/storage/poparser.py
@@ -35,17 +35,22 @@ From the GNU gettext manual:
      msgstr TRANSLATED-STRING
 """
 
-isspace = str.isspace
-find = str.find
-rfind = str.rfind
-startswith = str.startswith
+SINGLE_BYTE_ENCODING = 'iso-8859-1'
+isspace = six.text_type.isspace
+find = six.text_type.find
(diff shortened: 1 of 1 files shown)


translate/storage/poparser.py
```diff
@@ -323,10 +331,14 @@ def set_encoding(parse_state, store, unit):
...
 def decode_list(lst, decode):
-    return [decode(item) for item in lst]
+    return [decode(item.encode(SINGLE_BYTE_ENCODING)) for item in lst]
 
 
 def decode_header(unit, decode):
+    """
+    The header has been arbitrarily decoded with a single-byte encoding. We
+    re-encode it to decode values with the proper encoding defined in the header.
+    """
```
unho: This docstring looks like it belongs to `decode_list` instead.

unho:
Nice cleanup.

translate/storage/poparser.py
claudep: I amended the comment, but still kept it here. I think this explanation makes more sense under the decode_header function, because decode_list is more a sort of utility for decode_header. But if you still think I should move it, I'll do it.

unho:
Merged. Thanks!

PR #3316, opened 2015-08-26
__str__ to serialize()

claudep:
Before going further in this patch, I'd like to get your opinion on it. On Python 3, str(obj) cannot return bytes, so we have to do something at least. The "minimal" way could have been to replace __str__ by __bytes__, and then calling bytes(store). However, I think it would be clearer if we added a specific API (like serialize(), but could be another term) to serialize a store to a file. This is also easier if we have to pass arguments at some point, because adding arguments to Python magic methods is not very clean. Note that I still kept the serialization behavior of str(store) for Python 2, to not break compatibility if external tools count on it. The test suite still passes after the first commit. Not sure if this is worthy or not.

I'm waiting on your input before going further.

### DIFF
--- a/docs/formats/base_classes.rst
+++ b/docs/formats/base_classes.rst
@@ -158,12 +158,12 @@ what we discussed is related to the above.  A quick summary:
     of duplicate code in them). Ideally conversion should be as simple as::
 
       >>> po_store = POStore(filecontent)
-      >>> print(str(po_store))
+      >>> print(po_store.serialize())
       msgid "bleep"
       msgstr "blorp"
        
       >>> xliff_store = XliffStore(po_store)
-      >>> print(str(xliff_store))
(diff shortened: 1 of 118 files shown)


unho:
TTK is used in other places, so please lets keep `__str__()` so people using `str(store)` doesn't have to change their code.

claudep:
That's what I tried to stick to, at least on Python 2. On Python 3, it's not possible to keep this behavior because returning bytes from `__str__` is illegal.

claudep:
I've completed the patch now. Don't consider that as a pressure to accept it. It can also stay some time if you want to think a little bit more about it.

unho:
Will merge after rebase.

claudep:
Rebased.

unho:
Merged. Thanks!

PR #3317, opened 2015-08-27
py3: don't install deps that are only py2 compatible

dwaynebailey:
iniparse and vobject are not Python 3 ready.  So only install them if
we're in a Python 2 environment.

I haven't done anything in terms of not installing ini2po or ical2po on py3 or bailing early if the deps don't exist.  But just wanted to get some thoughts on using the new pip syntax to install these deps only on py2

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -4,8 +4,8 @@
 # Format support #
 ##################
 BeautifulSoup4>=4.3  # Trados TM
-iniparse>=0.3.1      # INI (ini2po)
-vobject>=0.6.6       # iCal (ical2po)
+iniparse>=0.3.1 ;python_version < '3.0'      # INI (ini2po)
+vobject>=0.6.6 ;python_version < '3.0'       # iCal (ical2po)
 
 #aeidon>=0.14        # Subtitles (sub2po)
 # aeidon not available through pip/PyPI,


unho:
I am fine with using environment markers: https://pip.pypa.io/en/latest/reference/pip_install.html#requirement-specifiers

requirements/optional.txt
```diff
@@ -4,8 +4,8 @@
 # Format support #
 ##################
 BeautifulSoup4>=4.3  # Trados TM
-iniparse>=0.3.1      # INI (ini2po)
-vobject>=0.6.6       # iCal (ical2po)
+iniparse>=0.3.1 ;python_version < '2.7'      # INI (ini2po)
+vobject>=0.6.6 ;python_version < '2.7'       # iCal (ical2po)
```
unho: Shouldn't these be `<3.0` instead?

claudep:
The problem with that syntax is that it blows up on pip < 1.6. For example, my pip is 1.5.6 on Debian stable, and here is what I'm getting:
```
ValueError: ("Expected ',' or end-of-list in", "iniparse>=0.3.1 ;python_version < '3.0'", 'at', " ;python_version < '3.0'")
```
So if we keep this (I have no better solution to propose), we should also document somewhere that TTK needs pip >= 1.6 for requirements installation (recommending `pip install --upgrade pip` in this case).

requirements/optional.txt
dwaynebailey: Oops yes or at least <=

dwaynebailey:
@claudep yes we'd need to document for older pip versions.

Only other solution I've been able to think of would be breaking our <3.0 stuff into its our optional-py2.txt but that seems like as much effort in terms of documentation and likely errors.

claudep:
What about finishing/committing this one?

unho:
I think it is ok to merge this.

claudep:
What about a note somewhere in the documentation about the minimal pip version?

unho:
Yes, it is necessary to add notes at least on installation and latest release notes (dev). This was just a quick check, so perhaps I missed some other places where this needs to be highlighted.

claudep:
1.14.0-rc1 release notes now exist, should be easier to finish this patch :-)

unho:
Please do not use 1.14.0-rc1 release notes, but instead dev ones. dev will become whichever next version is at the moment of the release. A commit merging 1.14.0-rc1 release notes into dev has been merged to master.

dwaynebailey:
@claudep @unho feel free to pick this up and land it with docs changes. I'm just not getting any time to look at this at the moment.

unho:
This PR is included in #3344, so closing this.

PR #3318, opened 2015-08-27
Made unquote_plus Python 3 compatible

claudep:
On Python 3 unquote_plus accepts only text (unicode) and returns text.


unho:
Merged. Thanks!

PR #3319, opened 2015-08-27
Removed autoencode class

claudep:
While difficult to port to Python 3, the autoencode class is not useful
any longer, as basically units should stay unicode and encoding now
happens only at file level.


unho:
Merged. Thanks!

PR #3321, opened 2015-08-28
Added encoding handling in base TranslationStore class

claudep:


translate/storage/csvl10n.py
```diff
@@ -351,10 +351,10 @@ class csvfile(base.TranslationStore):
     Mimetypes = ['text/comma-separated-values', 'text/csv']
     Extensions = ["csv"]
 
-    def __init__(self, inputfile=None, fieldnames=None, encoding="auto"):
+    def __init__(self, inputfile=None, fieldnames=None, encoding=None):
         base.TranslationStore.__init__(self, unitclass=self.UnitClass)
         self.units = []
-        self.encoding = encoding or 'utf-8'
+        self.encoding = encoding or 'auto'
```
unho: Is it actually necessary to use `auto`? I mean, we should fallback to `utf-8` and that can be accomplished if encoding is `None`, isn't it?

claudep: The problem is that there is some complex logic in the `detect_encoding` method which is partly based on the presence of `auto`. I admit I didn't try to change that, and I did not check if that behavior was thoroughly tested or not. I chose the "less-change-as-possible" way for that part. I could investigate a bit more on that `auto` value if you think we should do it.

translate/storage/lisa.py
```diff
@@ -343,7 +342,7 @@ def parse(self, xml):
             xml = posrc
         parser = etree.XMLParser(strip_cdata=False)
         self.document = etree.fromstring(xml, parser).getroottree()
-        self._encoding = self.document.docinfo.encoding
+        self.encoding = self.document.docinfo.encoding
```
unho: It seems that in this file we are extracting the encoding but not using it. Not an issue that needs to be addressed in this PR.

translate/storage/pocommon.py
```diff
@@ -205,11 +205,10 @@ class pofile(poheader.poheader, base.TranslationStore):
     # We don't want windows line endings on Windows:
     _binary = True
 
-    def __init__(self, inputfile=None, encoding=None, duplicatestyle=None):
-        super(pofile, self).__init__(unitclass=self.UnitClass)
+    def __init__(self, inputfile=None, duplicatestyle=None, **kwargs):
```
unho: I am worried about these changes in the signatures. The order is altered and the positional `encoding` is now a keyword argument. Please note that this might be used in third party software.

translate/storage/pypo.py
```diff
@@ -864,12 +862,9 @@ def _getoutput(self):
     def encode(self, lines):
         """encode any unicode strings in lines in self._encoding"""
```
unho: This docstring needs a rewrite?

translate/storage/pypo.py
```diff
@@ -864,12 +862,9 @@ def _getoutput(self):
...
         newlines = []
-        encoding = self._encoding
-        if encoding is None or encoding.lower() == "charset":
-            encoding = 'UTF-8'
         for line in lines:
             if isinstance(line, six.text_type):
-                line = line.encode(encoding)
+                line = line.encode(self.encoding)
```
unho: A few lines below, in `decode()`, `self._encoding` is being used but probably is fine to use `self.encoding` and simplify the if condition a bit.

translate/storage/tbx.py
```diff
@@ -35,7 +35,7 @@ class tbxunit(lisa.LISAunit):
     def createlanguageNode(self, lang, text, purpose):
         """returns a langset xml Element setup with given parameters"""
         if isinstance(text, bytes):
-            text = text.decode("utf-8")
+            text = text.decode(self.encoding)
```
unho: Shouldn't it be the encoding for the file?

translate/storage/csvl10n.py
unho: OK. I would probably have done the same.

translate/storage/lisa.py
claudep: LISAfile is used as a parent class in different other locations, I wouldn't bet that self.encoding is never used...

translate/storage/pocommon.py
claudep: Generally, the default values often imply that these should be used as keyword arguments. Moreover, `encoding` was placed at different places in different constructors. It would be a different story if it would always be placed at the same position (as is often `inputfile`).

translate/storage/pypo.py
claudep: Yes, the underscore should be removed.

translate/storage/pypo.py
claudep: Sure, I have probably forgotten to fix that. Will do.

translate/storage/tbx.py
claudep: Good catch. The idea is that at some point, units would never receive encoded text, so we would be able to remove these lines. But for now, it's safer to simply revert that change, not a big deal.

unho:
Will merge after rebase.

claudep:
Done.

unho:
Merged. Thanks!

PR #3324, opened 2015-08-28
Reopen file with open instead of a class constructor

claudep:
Python 3 doesn't seem to like opening files with the file
class. Let's stick to the classical open().


unho:
Merged. Thanks!

PR #3325, opened 2015-08-28
Standardized UnitClass definition in TranslationStore subclasses

claudep:


translate/storage/csvl10n.py
```diff
@@ -352,8 +352,7 @@ class csvfile(base.TranslationStore):
     Extensions = ["csv"]
 
     def __init__(self, inputfile=None, fieldnames=None, encoding=None):
-        base.TranslationStore.__init__(self, unitclass=self.UnitClass)
-        self.units = []
+        super(csvfile, self).__init__()
         self.encoding = encoding or 'auto'
```
unho: Why not passing the encoding to `__init__()`?

translate/storage/subtitles.py
```diff
@@ -79,11 +79,9 @@ class SubtitleFile(base.TranslationStore):
     """A subtitle file"""
     UnitClass = SubtitleUnit
 
-    def __init__(self, inputfile=None, unitclass=UnitClass):
+    def __init__(self, inputfile=None, **kwargs):
         """construct an Subtitle file, optionally reading in from inputfile."""
-        self.UnitClass = unitclass
-        base.TranslationStore.__init__(self, unitclass=unitclass)
-        self.units = []
+        super(SubtitleFile, self).__init__(self, **kwargs)
```
unho: Why `__init__(self, **kwargs)` and not just `__init__(**kwargs)`?

unho:
Nice cleanup. Will merge after all doubts/nits are addressed and PR is rebased.

claudep:
Fixed what you noticed and rebased.

unho:
I am going to merge this despite that perhaps the `unitclass` argument perhaps can also be removed from `base.TranslationStore.__init__()`. But that can be done in a future PR if considered a good idea.

unho:
Merged. Thanks!

claudep:
I don't know either if `unitclass` as a constructor argument is useful or even used. However, it is currently part of the API, so if any third-party is currently calling SomeStoreClass(unitclass=customUnitClass), we would break compatibility if we remove that possibility.

unho:
Sure.

PR #3326, opened 2015-08-28
Accept unicode input for parsestring

claudep:
As text (unicode) becomes ubuquitous on Python 3, it is easier
to just accept and transform any text content to bytes in parsetring
before feeding parse() with it.


unho:
Merged. Thanks!

PR #3327, opened 2015-08-29
Made pypo.pounit encoding agnostic

claudep:
This is a follow-up commit of 0199313bd63.

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -35,7 +35,6 @@ from translate.misc import quote
 from translate.misc.deprecation import deprecated
 from translate.misc.multistring import multistring
 from translate.storage import base, pocommon, poparser
-from translate.storage.pocommon import encodingToUse
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


translate/storage/pypo.py
```diff
@@ -35,7 +35,6 @@
 from translate.misc.deprecation import deprecated
 from translate.misc.multistring import multistring
 from translate.storage import base, pocommon, poparser
-from translate.storage.pocommon import encodingToUse
```
unho: If I am not wrong this is the last usage of `encodingToUse` so perhaps it can also removed from codebase.

claudep:
You're right. I removed that function and rebased the patch.

translate/storage/test_pypo.py
```diff
@@ -261,17 +262,17 @@ def test_output_str_unicode(self):
         assert len(pofile.units) == 1
         print(pofile.serialize())
         thepo = pofile.units[0]
-        assert str(thepo) == posource.encode("UTF-8")
-        # extra test: what if we set the msgid to a unicode? this happens in prop2po etc
-        thepo.source = u"Norwegian Bokm\xe5l"
```
unho: Why is this extra test removed?

claudep: Because with the new strategy, thepo.source is already a unicode string and setting `thepo.source` to an unicode string is not doing anything worth testing, IMHO.

unho:
Merged. Thanks!

PR #3328, opened 2015-08-29
Add minimal test for po to json conversion

claudep:


translate/convert/test_po2json.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def test_basic(self):
+        """test a basic po to json conversion"""
+        json_source = '''{ "text": "A simple string"}'''
+        input_po = u'''#: .text
+msgid "A simple string"
+msgstr "Du texte simple"
+'''
+        json_out = self.po2json(input_po, json_source)
+        assert '"text": "Du texte simple"' in json_out
```
unho: It is preferable to add a variable holding the whole expected JSON output and assert with that variable.

translate/convert/test_po2json.py
```diff
@@ -0,0 +1,26 @@
...
+        input_file = wStringIO.StringIO(po_source)
+        json_file = wStringIO.StringIO(json_template)
+        input_po = po.pofile(input_file)
+        convertor = po2json.rejson(json_template, input_po)
+        output_json = convertor.convertstore()
+        return output_json
+
+    def test_basic(self):
+        """test a basic po to json conversion"""
+        json_source = '''{ "text": "A simple string"}'''
```
unho: I would prefer this to be renamed to `json_template` since it states better its purpose.

unho:
Will merge after raised issues are properly addressed and PR is rebased.

claudep:
Like that (also rebased)?

unho:
Merged. Thanks!

PR #3330, opened 2015-08-29
Output text from TranslationUnit.__str__

claudep:
This is the continuation of deferring serialization to
stores instead of units. Moreover, this makes it compatible
with Python 3.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -42,6 +42,7 @@ class ParseError(Exception):
         return repr(self.inner_exc)
 
 
+@six.python_2_unicode_compatible
 class TranslationUnit(object):
     """Base class for translation units.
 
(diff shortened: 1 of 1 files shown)


unho:
Have you tested if this impacts performance?

claudep:
It should not, as this method is almost never used, because subclasses often override it. I've also not seen any cases where it was used for store serialization. So I don't even know where to start to make a benchmark.
In the test suite, only 6 tests out of 2071 are calling this method. In 5 cases, this is in testing abstract classes (test_base and test_monolingual) and one is for mo files (but the mo store serialization don't use it).

unho:
Would like to have input from @dwaynebailey.

dwaynebailey:
@unho I'm not sure what your concerns might be.  I think this one is gtm

unho:
lgtm

PR #3331, opened 2015-08-31
Improved encoding detection

nijel:
This series of patches fixes handling of unicode files with BOM with defined encoding. Eg. UTF-8 Java properties specify utf-8 encoding, but if the file has BOM, the BOM bytes are currently wrongly parsed as content of first unit.

Changes made to encoding detection:

- converts detected encoding to lower case as that's what is used in rest of the sources (so that we can actually compare these strings)
- removes double adding of `self.encoding` to list of detected encodings
- adds the BOM variant of the encoding if it has been detected
- brings back wrong encoding warning for confidence = 1.0 (because I've misunderstood surrounding code in #67)
- avoids detecting ascii text as utf-8 if it is expected to be iso-8859-1
- added basic unicode detection even without chardet

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -20,6 +20,7 @@
 
 """Base classes for storage interfaces."""
 
+import codecs
 import logging
 import six
 try:
@@ -31,6 +32,17 @@ from translate.misc.multistring import multistring
 from translate.storage.placeables import (StringElem, general,
                                           parse as rich_parse)
 from translate.storage.workflow import StateEnum as states
+
+# Simple BOM based encoding detection
+ENCODING_BOMS = (
+    (codecs.BOM_UTF8, 'utf-8-sig'),
+    (codecs.BOM_UTF16, 'utf-16'),
+    (codecs.BOM_UTF16_BE, 'utf-16-be'),
+    (codecs.BOM_UTF16_LE, 'utf-16-le'),
+    (codecs.BOM_UTF32, 'utf-32'),
+    (codecs.BOM_UTF32_BE, 'utf-32-be'),
+    (codecs.BOM_UTF32_LE, 'utf-32-le'),
+)
 
 
 class ParseError(Exception):
@@ -737,6 +749,18 @@ class TranslationStore(object):
             newstore.parse(storestring)
         return newstore
 
+    def fallback_detection(self, text):
+        """
+        Simple detection based on BOM in case chardet is not available.
+        """
+        for bom, encoding in ENCODING_BOMS:
+            if text.startswith(bom):
+                return {
+                    'encoding': encoding,
+                    'confidence': 1.0
+                }
+        return None
+
     def detect_encoding(self, text, default_encodings=None):
         """
         Try to detect a file encoding from `text`, using either the chardet lib
@@ -755,7 +779,7 @@ class TranslationStore(object):
             else:
                 detected_encoding['encoding'] = detected_encoding['encoding'].lower()
         except ImportError:
-            detected_encoding = None
+            detected_encoding = self.fallback_detection(text)
 
         encodings = []
         # Purposefully accessed the internal _encoding, as encoding is never 'auto'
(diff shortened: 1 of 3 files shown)


nijel:
Looks like something went wrong here - the tests work locally, but fail on Travis. I'll look into that.

tests/cli/data/test_prop2po_files/stderr.txt
```diff
@@ -0,0 +1 @@
+root: WARNING: trying to parse ./data/test_prop2po_files/one.properties with encoding: iso-8859-1 but detected encoding is utf-8 (confidence: 1.0)
```
claudep: Previously the warning was only emitted when confidence was below 1.0. Apparently you removed that condition in the patch. Was that intended?

nijel: That was intended, because in the code path the detected encoding is not used at all. However I have no clue what ever original intentions here...

nijel: Thinking more about this, this warning is really wrong in this case. I've tried to fix that in 548445de92ed5928898cdbd45741c34503667468.

translate/storage/base.py
```diff
@@ -759,7 +759,9 @@ def detect_encoding(self, text, default_encodings=None):
             if detected_encoding['confidence'] < 0.48:
                 detected_encoding = None
             elif detected_encoding['encoding'] == 'ascii':
-                detected_encoding['encoding'] = 'utf-8'
+                detected_encoding['encoding'] = self.encoding
+            else:
+                detected_encoding['encoding'] = detected_encoding['encoding'].lower()
         except ImportError:
             detected_encoding = None
```
claudep: Could we add here:
```if text.startswith(codecs.BOM_UTF8):
    detected_encoding = 'utf-8'
if text.startswith(codecs.BOM_UTF16):
    detected_encoding = 'utf-16'
```
so that the test will pass even without chardet installed.

nijel: Dne Wed, 02 Sep 2015 08:38:22 -0700
Claude Paroz <notifications@github.com> napsal(a):

> > @@ -759,7 +759,9 @@ def detect_encoding(self, text, default_encodings=None):
> >              if detected_encoding['confidence'] < 0.48:
> >                  detected_encoding = None
> >              elif detected_encoding['encoding'] == 'ascii':
> > -                detected_encoding['encoding'] = 'utf-8'
> > +                detected_encoding['encoding'] = self.encoding
> > +            else:
> > +                detected_encoding['encoding'] = detected_encoding['encoding'].lower()
> >          except ImportError:
> >              detected_encoding = None
> 
> Could we add here:
> ```if text.startswith(codecs.BOM_UTF8):
>     detected_encoding = 'utf-8'
> if text.startswith(codecs.BOM_UTF16):
>     detected_encoding = 'utf-16'
> ```
> so that the test will pass even without chardet installed.

Added something like that in 3ed9e62.

-- 
	Michal Čihař | http://cihar.com | http://blog.cihar.com

translate/storage/base.py
```diff
@@ -22,6 +22,7 @@
 
 import logging
 import six
+import codecs
```
claudep: imports should be alphabetically sorted

claudep:
I think commits should now be squashed.

nijel:
I've split them into logical pieces, if you prefer single commit doing all changes I can squash them.

claudep:
That was just a suggestion, as I'm not a committer myself. @unho ?

unho:
It is ok to have all these commits separate since this way each specific change has a detailed explanation on the commit message.

gtm after rebase.

nijel:
Rebased on current master.

claudep:
Merged, thanks.

nijel:
@claudep, I still don't see commits in the repository, is that okay?

claudep:
They are, see https://github.com/translate/translate/commits/master under my "Fixed ... on Python 3" series.

nijel:
Ah, you've rebased it again. Okay then.

PR #3332, opened 2015-09-03
Fixed unit equality issue with monolingual units

### LINKED ISSUE #3322: unit equality issue
As the `__ne__` method is not defined for translation units, there is a bug with some unit comparison, for example txt units return True for both `==` and `!=`.
This patch shows the problem when running the test suite:

```diff
diff --git a/translate/storage/test_base.py b/translate/storage/test_base.py
index b65104d..fcf586b 100644
--- a/translate/storage/test_base.py
+++ b/translate/storage/test_base.py
@@ -109,6 +109,7 @@ class TestTranslationUnit:
         assert unit1 != unit3
         assert unit4 != unit5
         assert unit1 != unit6
+        assert not(unit1 == unit6)
 
     def test_target(self):
         unit = self.unit
```

This unexpectedly fails for `TxtUnit`, `RESXUnit`, and `dtdunit`.

We can define the `__ne__` method:

```diff
diff --git a/translate/storage/base.py b/translate/storage/base.py
index b2123b2..22a2bc8 100644
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -141,6 +141,9 @@ class TranslationUnit(object):
         """
         return self.source == other.source and self.target == other.target
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     def __str__(self):
         """Converts to a string representation that can be parsed back using
         :meth:`~.TranslationStore.parsestring`."""
```

But now `test_eq()` fails, because for some units (mainly monolingual), the source is equal to the target so unit1 and unit6 are indeed identical. I tried to special case the comparison for tests subclassing `TestMonolingualUnit`, but this doesn't seem to be sufficient, as monolingual units are not all behaving the same.

Any pointer from a more experienced dev?

claudep:
Fixes #3322.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -78,6 +78,7 @@ class TranslationUnit(object):
 
     """
 
+    monolingual = False
     rich_parsers = []
     """A list of functions to use for parsing a string into a rich string
     tree."""
@@ -138,6 +139,9 @@ class TranslationUnit(object):
                  equals this unit.
         """
         return self.source == other.source and self.target == other.target
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
 
     def __hash__(self):
         return None
--- a/translate/storage/dtd.py
+++ b/translate/storage/dtd.py
@@ -225,6 +225,7 @@ def removeinvalidamps(name, value):
 @six.python_2_unicode_compatible
 class dtdunit(base.TranslationUnit):
     """An entity definition from a DTD file (and any associated comments)."""
+    monolingual = True
 
     def __init__(self, source="", android=False):
         """construct the dtdunit, prepare it for parsing"""
--- a/translate/storage/resx.py
+++ b/translate/storage/resx.py
@@ -29,6 +29,7 @@ from translate.storage.placeables import general
 class RESXUnit(lisa.LISAunit):
     """A single term in the RESX file."""
 
+    monolingual = True
     rootNode = "data"
     languageNode = "value"
     textNode = ""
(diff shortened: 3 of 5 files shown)


unho:
Just curious why only those monolingual formats needed to be adjusted and all the other monolingual formats don't.

Will merge after some clarification is provided and PR is rebased.

claudep:
Is there somewhere a list of monolingual store types?

unho:
> Is there somewhere a list of monolingual store types?

No, there is not such a list.

- Monolingual formats are: android, catkeys, csv, dtd, html, ical, idml, ini, json, mozilla lang, odf, php, properties, rc, resx, subtitles, txt.
- Not sure about: omegat, oo, symbian, tiki, trados, wordfast, utx.
- Definitely bilingual or multilingual: po, qm, qph, tbx, tmx, ts, xliff.

claudep:
It's either a real problem in the various implementations, or missing tests. However, I'm afraid to be still an inexperienced ttk dev to be able to say much more about this issue.

unho:
I am not sure if there is anybody with enough experience and knowledge of TTK to provide an answer to that question.

claudep:
We could also simply special-case 'RESXUnit', 'dtdunit' and 'TxtUnit' unit classes in the tests. Tests will not pass on Python 3 unless we are fixing this, so someone has to decide something at some point... I don't feel like reimplementing unit classes now.

unho:
Ok, then just merge this PR. Future adjustments, if any, can be proposed in a new PR.

claudep:
Obsoleted by #3342

PR #3334, opened 2015-09-03
Removed stdiotell.StdIOWrapper

claudep:
I may have missed something, but I didn't find any useful usage for the stdiotell.StdIOWrapper wrapper class in the current version of the ttk.


translate/convert/tiki2po.py
```diff
@@ -80,8 +80,6 @@ def converttiki(inputfile, outputfile, template=None, includeunused=False):
 def main(argv=None):
     """Converts tiki .php files to .po."""
     from translate.convert import convert
-    from translate.misc import stdiotell
-    sys.stdout = stdiotell.StdIOWrapper(sys.stdout)
```
unho: You forgot to remove `sys` import.

unho:
I am a bit concerned that we might be breaking something by removing `stdiotell.StdIOWrapper`, but it seems harmless and tests seem to pass. So will merge after the nit is addressed and the PR is rebased.

claudep:
I'm almost certain we don't break anything in TTK itself, as I've audited all `tell()` calls. Of course, we never know about third-party code, but I don't see the real added value in that utility. It was certainly useful in the past for some usage.
Removed the leftover import and rebased.

unho:
Merged. Thanks!

PR #3335, opened 2015-09-09
Fix tools tests on Python 3

claudep:


claudep:
Just tell me if you'd prefer having commits squashed.

translate/tools/test_podebug.py
```diff
@@ -28,8 +28,8 @@ class TestPODebug:
     debug = podebug.podebug()
 
     def setup_method(self, method):
-        self.postore = po.pofile(PO_DOC)
-        self.xliffstore = xliff.xlifffile(XLIFF_DOC)
+        self.postore = po.pofile(PO_DOC.encode())
```
unho: Encode to which coding? I suppose it is `utf-8` but would be great if it is explicit.

translate/tools/test_phppo2pypo.py
```diff
@@ -21,10 +22,10 @@ def test_single_po(self):
 msgid "I have %2$s apples and %1$s oranges"
 msgstr "I have %2$s apples and %1$s oranges"
         """
-        outputfile = wStringIO.StringIO()
+        outputfile = BytesIO()
         phppo2pypo.convertphp2py(inputfile, outputfile)
 
-        output = outputfile.getvalue()
+        output = outputfile.getvalue().decode()
```
unho: Is it possible to be explicit about the encoding we are decoding from?

translate/tools/test_podebug.py
claudep: I didn't bother specifying the encoding when only ascii content was at work, but I can review the patch to always specify an encoding if you wish.

translate/tools/test_pomerge.py
```diff
@@ -488,17 +488,17 @@ def test_merging_header_entries(self):
...
         locations that we don't go fuzzy (bug 1583)"""
 
-        templatepo = r'''#, fuzzy
+        templatepo = u'''#, fuzzy
 msgid ""
 msgstr ""
-"Content-Type: text/plain; charset=utf-8\n"
-"Content-Transfer-Encoding: 8bit\n"
+"Content-Type: text/plain; charset=utf-8\\n"
+"Content-Transfer-Encoding: 8bit\\n"
```
unho: Here it goes our multiline string that contained valid PO source :(

translate/tools/pocount.py
```diff
@@ -55,14 +55,14 @@ def calcstats_old(filename):
     except ValueError as e:
         logger.warning(e)
         return {}
-    units = filter(lambda unit: unit.istranslatable(), store.units)
+    units = list(filter(lambda unit: unit.istranslatable(), store.units))
```
unho: Why not use list comprehension? I don't particularly like keeping `filter`.

unho: Actually I hate more `lambda` than `filter`, but that is just some personal preference of mine.

translate/tools/pocount.py
```diff
@@ -55,14 +55,14 @@ def calcstats_old(filename):
...
     translated = translatedmessages(units)
     fuzzy = fuzzymessages(units)
-    review = filter(lambda unit: unit.isreview(), units)
+    review = list(filter(lambda unit: unit.isreview(), units))
     untranslated = untranslatedmessages(units)
-    wordcounts = dict(map(lambda unit: (unit, statsdb.wordsinunit(unit)), units))
-    sourcewords = lambda elementlist: sum(map(lambda unit: wordcounts[unit][0], elementlist))
-    targetwords = lambda elementlist: sum(map(lambda unit: wordcounts[unit][1], elementlist))
+    wordcounts = dict(map(lambda unit: (id(unit), statsdb.wordsinunit(unit)), units))
```
unho: Is it not cleaner the following?

```python
    wordcounts = dict((id(unit), statsdb.wordsinunit(unit)) for unit in units)
```

translate/tools/test_podebug.py
unho: I would prefer that. Tests are something that should use clean and readable code so it is easy to understand and spot errors.

unho:
No need to squash the commits. Actually it kinda makes it easier to understand each single set of changes.

translate/tools/test_pomerge.py
claudep: Do you suggest another approach (considering that ur'' prefix is not valid in Python 3)?

unho: No, I was just wailing for its loss :)

claudep: Finally found that with unicode_literals import, we could use raw strings together with unicode content! You can wipe your tears...

unho:
lgtm

PR #3338, opened 2015-09-10
Removed unneeded 'bin/env python' preambles

claudep:


unho:
Just reviewed. Not sure what that error is about, but if it is because 'bin/env python' preambles are being removed then obviously this PR is not going to be merged. After all those preambles are harmless.

claudep:
Don't worry, I mistakenly removed a different line. Should be fixed.

unho:
lgtm

PR #3339, opened 2015-09-14
Fix convert tests on Python 3

claudep:


translate/storage/csvl10n.py
```diff
@@ -314,9 +314,11 @@ def try_dialects(inputfile, fieldnames, dialect):
 
 
 def valid_fieldnames(fieldnames):
-    """check if fieldnames are valid"""
+    """Check if fieldnames are valid, that is at least one field is identified
+       as the source.
+    """
     for fieldname in fieldnames:
-        if fieldname in canonical_field_names and fieldname == 'source':
+        if fieldname == 'source':
```
unho: `canonical_field_names` is no longer used, so it can be dropped.

translate/convert/test_tiki2po.py
```diff
@@ -4,6 +4,8 @@
 # Author: Wil Clouser <wclouser@mozilla.com>
 # Date: 2008-12-01
 
+from io import BytesIO
+
 from translate.convert import test_convert, tiki2po
 from translate.misc import wStringIO
```
unho: You can drop this `wStringIO` import.

translate/convert/xliff2po.py
```diff
@@ -24,6 +24,8 @@
 for examples and usage instructions.
 """
 
+from io import IOBase
```
unho: This import is not used.

translate/convert/xliff2po.py
```diff
@@ -24,6 +24,8 @@
...
+
 from translate.misc import wStringIO
```
unho: This is no longer used neither.

translate/convert/xliff2po.py
```diff
@@ -67,11 +69,6 @@ def converttransunit(self, transunit):
 
     def convertstore(self, inputfile, duplicatestyle="msgctxt"):
         """Converts a .xliff file to .po format"""
-        # XXX: The inputfile is converted to string because Pootle supplies
-        # XXX: a PootleFile object as input which cannot be sent to PoXliffFile
-        # XXX: The better way would be to have a consistent conversion API.
-        if not isinstance(inputfile, (file, wStringIO.StringIO)):
-            inputfile = str(inputfile)
```
unho: Are you sure this change is harmless?

claudep: No, I'm not absolutely sure. However, it's old code (~2007) and no tests are failing. Maybe we can make sure by running Pootle tests against ttk master?

unho: Pootle tests are not as extensive as TTK ones and the use mostly PO files, so I don't think testing on Pootle will help.

I see you are replacing `StringIO` with `BytesIO` in other places, so perhaps it is ok to keep this hack with that minor change.

claudep: On Python 3, the problem is `file`, not `StringIO`. I can find a workaround, but keeping old code around when you don't even know what it's supposed to fix is not very sane. Sometimes, it's better to break code and reintroduce a fix later with proper tests rather than keeping `magic` code forever.

unho: Just did some bisect and `PootleFile` object for which this hack was introduced seems to be gone from Pootle since 2009.

So I say lets move this specific change to a separate commit with a detailed commit message including the relevant parts of this chat. That way if we have to investigate it will be easier to find a small change with a detailed description and that is easily revertable.

claudep: I searched pootle code for xliff2po/convertstore/convertxliff occurrences but found nothing. What did I miss?

unho: That probably was used in old Pootle and now is gone. Please craft the commit I asked for in https://github.com/translate/translate/pull/3339/files#r39437681

translate/storage/csvl10n.py
unho: Done

unho:
lgtm

PR #3341, opened 2015-09-14
Removed obsolete hack in xliff2po.convertstore

claudep:
This workaround was introduced for a Pootle use case, and the
PootleFile object it was supposed to support is gone since 2009.
Moreover, the `file` type object is no more on Python 3.


unho:
lgtm

PR #3342, opened 2015-09-15
Fixed unit equality issue with some monolingual units

### LINKED ISSUE #3322: unit equality issue

claudep:
Fixes #3322.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -139,6 +139,9 @@ class TranslationUnit(object):
         """
         return self.source == other.source and self.target == other.target
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     def __hash__(self):
         return None
 
--- a/translate/storage/test_base.py
+++ b/translate/storage/test_base.py
@@ -82,7 +82,19 @@ class TestTranslationUnit:
         assert unit1 == unit2
         assert unit1 != unit3
         assert unit4 != unit5
-        assert unit1 != unit6
+        if unit1.__class__.__name__ in ('RESXUnit', 'dtdunit', 'TxtUnit'):
+            # unit1 will generally equal unit6 for monolingual formats (resx, dtd, txt)
+            # with the default comparison method which compare units by their
+            # target and source properties only.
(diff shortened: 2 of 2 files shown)


claudep:
Here's an alternative way to "fix" the current issue. It's still hacky, but I prefer handling the differences in the tests, rather than put a `monolingual` attribute to arbitrary classes. Would replace #3332.

unho:
I prefer this to #3332.

lgtm

PR #3344, opened 2015-09-15
Activate test runs on Python 3

claudep:
Just to see where we are...

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -4,8 +4,8 @@
 # Format support #
 ##################
 BeautifulSoup4>=4.3  # Trados TM
-iniparse>=0.3.1      # INI (ini2po)
-vobject>=0.6.6       # iCal (ical2po)
+iniparse>=0.3.1 ;python_version < '3.0'      # INI (ini2po)
+vobject>=0.6.6 ;python_version < '3.0'       # iCal (ical2po)
 
 #aeidon>=0.14        # Subtitles (sub2po)
 # aeidon not available through pip/PyPI,
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ language: python
 python:
   - 2.6
   - 2.7
+  - 3.3
+  - 3.4
 env:
   matrix:
     - USECPO=0 MINVERSION=1 META=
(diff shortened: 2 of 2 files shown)


unho:
It seems that some tests are not passing. Is that expected?

claudep:
Yes, sure. This pull request is not meant to be merged now, it's just a way to see the remaining work. But I think (and hope) we are rather close!

claudep:
pip installation currently fails because python-Levenshtein==0.10.2 (Python 3 support has been added in 0.11)

unho:
Current requirement is:

python-Levenshtein>=0.10.2,!=0.11.0

So it should be ok to use instead:

python-Levenshtein>=0.11.1

unho:
I see that Travis now runs 7 jobs each time is fired. Perhaps all those are necessary, but please check if you can reduce the number of jobs.

julen:
Congrats @claudep, well done :dancer:

claudep:
@unho also did a great job at reviewing...

unho:
lgtm. Congratulations.

claudep:
Let's still note that with ~65% test coverage, there are probably several remaining issues with Python 3. I'll try to stay around for fixing those, at least.

unho:
That is much appreciated.

KevinEtienne:
Do you think this is ready to be included in the next release? It might worth adding support for `3.5` and maybe dropping `3.3`?

requirements/optional.txt
```diff
@@ -4,8 +4,8 @@
 # Format support #
 ##################
 BeautifulSoup4>=4.3  # Trados TM
-iniparse>=0.3.1      # INI (ini2po)
-vobject>=0.6.6       # iCal (ical2po)
+iniparse>=0.3.1 ;python_version < '3.0'      # INI (ini2po)
+vobject>=0.6.6 ;python_version < '3.0'       # iCal (ical2po)
```
claudep: Looks like vobject could be replaced by https://pypi.python.org/pypi/python-card-me (to be tested)

unho: Yes, but it seems it doesn't support Python 2.6.

unho:
@claudep It would be great if you can rebase this PR too. Also it seems there are some changes (for example "Fixed order of progress bar option names" commit) that seem to be ok to be merged to master.

claudep:
Rebased. The remaining (random) failure is due to a header ordering issue between X-Accelerator-Marker and X-Merge-On. They are not in `poheader.header_order` so their ordering is undefined.

unho:
@claudep Thanks. I would also merge the "Upgraded lxml minimal requirement for Python 3" commit as it won't harm.

unho:
@claudep Just merged some fix for random order in 2fa69bb38 and followup commit. Would be great if you can rebase again and merge the "Upgraded lxml minimal requirement for Python 3" commit.

unho:
This was merge to master. The support for INI and iCalendar is disabled when using Python 3.

claudep:
\o/

PR #3345, opened 2015-09-15
Fixed compiling all code on Python 3

claudep:
Even if some files might not be used with Python 3, the
'python -m compileall' command may choke on syntax issues.


unho:
lgtm

PR #3346, opened 2015-09-16
Drop obsolete PyLuceneIndexer1

claudep:
Latest 1.x PyLucene release was on March 2006. PyLucene 2.0 was
released on May 2006.


dwaynebailey:
lgtm

julen:
As a related note, `translate.search.indexing` could potentially go away: if I'm not mistake it's only been used for older Pootle's indexing support, which now makes use of ElasticSearch no more.

PR #3349, opened 2015-09-17
Write to output directly inside file.serialize()

claudep:
I'd like to take advantage of the fact that the serialize() API is still not part of a released version to adapt it to accept a file-like objects as first mandatory parameter. This allows us to write to that file directly inside the method, incrementally for those storages that support it. This should be more memory-friendly.

### DIFF
--- a/translate/storage/dtd.py
+++ b/translate/storage/dtd.py
@@ -93,6 +93,7 @@ except ImportError:
     etree = None
 
 from translate.misc import quote
+from translate.misc.deprecation import deprecated
 from translate.storage import base
 
 
@@ -561,20 +562,21 @@ class dtdfile(base.TranslationStore):
                     warnings.warn("%s\nError occured between lines %d and %d:\n%s" % (e, start + 1, end, b"\n".join(lines[start:end])))
                 start += linesprocessed
 
-    def serialize(self):
-        """convert to bytes. double check that unicode is handled somehow here"""
-        source = self.getoutput()
-        if not self._valid_store():
+    def serialize(self, out):
+        """Write content to file"""
+        content = b''
+        for dtd in self.units:
+            unit_str = six.text_type(dtd).encode(self.encoding)
+            out.write(unit_str)
+            content += unit_str
+        if not self._valid_store(content):
             warnings.warn("DTD file '%s' does not validate" % self.filename)
-            return None
-        if isinstance(source, six.text_type):
-            return source.encode(self.encoding)
-        return source
+            out.truncate(0)
 
+    # Deprecated on 1.14
+    @deprecated("Use bytes(dtdfile) instead")
     def getoutput(self):
-        """convert the units back to source"""
-        sources = [six.text_type(dtd) for dtd in self.units]
-        return "".join(sources)
+        return bytes(self)
 
     def makeindex(self):
         """makes self.id_index dictionary keyed on entities"""
@@ -583,7 +585,7 @@ class dtdfile(base.TranslationStore):
             if not dtd.isnull():
                 self.id_index[dtd.entity] = dtd
 
-    def _valid_store(self):
+    def _valid_store(self, content):
         """Validate the store to determine if it is valid
 
         This uses ElementTree to parse the DTD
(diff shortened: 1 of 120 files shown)


translate/tools/poconflicts.py
```diff
@@ -181,7 +181,7 @@ def str_len(x):
                 unit.othercomments.append("# (poconflicts) %s\n" % filename)
                 conflictfile.units.append(unit)
             with open(fulloutputpath, "wb") as fh:
-                fh.write(conflictfile.serialize())
+                conflictfile.serialize(fh)
```
unho: Shouldn't this be `fh.write(conflictfile)`?

unho: The same goes for other similar changes.

claudep: No, the idea of this patch is to pass the file handle to the serialize call which will directly write into the file. However, this code would be clearer if `conflictfile` was better renamed to `conflict_store`.

translate/storage/dtd.py
```diff
@@ -561,20 +561,18 @@ def parse(self, dtdsrc):
...
-        return "".join(sources)
+    def serialize(self, out):
+        """Write content to file"""
+        content = b''
+        for dtd in self.units:
+            unit_str = six.text_type(dtd).encode(self.encoding)
+            out.write(unit_str)
+            content += unit_str
+        # Android files are invalid DTDs
+        if etree is not None and not self.android:
```
unho: What was the reason to pull this out of `_valid_store()`?

claudep: I don't think that there is a good reason. I just moved code back and forth until I was satisfied and the test was finally in serialize. I *think* it could be re-moved inside  `_valid_store`, would this be your preferred solution?

translate/storage/txt.py
```diff
@@ -145,14 +145,8 @@ def parse(self, lines):
             unit = self.addsourceunit("\n".join(block))
             unit.addlocation("%s:%d" % (self.filename, startline + 1))
 
-    def serialize(self):
-        source = self.getoutput()
-        if isinstance(source, six.text_type):
-            return source.encode(self.encoding)
-        return source
-
-    def getoutput(self):
```
unho: I agree with normalizing the API, but this wasn't a private method, so I am not sure if just removing it is the best approach.

unho:
Review complete.

translate/storage/dtd.py
unho: I think it is the right approach to have it on `_valid_store`.

translate/tools/poconflicts.py
unho: ok

claudep:
I pushed fixes as separate commits so you can better review them. Will merge them if all is well.

translate/storage/txt.py
```diff
@@ -145,14 +146,13 @@ def parse(self, lines):
...
-            return source.encode(self.encoding)
-        return source
+    def serialize(self, out):
+        for idx, unit in enumerate(self.units):
+            if idx > 0:
+                out.write(b'\n\n')
+            out.write(six.text_type(unit).encode(self.encoding))
 
+    # Deprecated on 1.14
+    @deprecated("Use TxtFile.serialize(out_file) instead")
```
unho: Please record the deprecation on the release notes.

translate/storage/test_txt.py
```diff
@@ -33,7 +33,13 @@ def test_multipleblocks(self):
...
         print(txtsource)
-        print(txtfile.serialize())
+        print(bytes(txtfile))
         print("*%s*" % txtfile.units[0])
-        assert txtfile.serialize().decode('utf-8') == txtsource
+        assert bytes(txtfile).decode('utf-8') == txtsource
         assert self.txtregen(txtsource) == txtsource
+
+    def test_getoutput(self):
+        # Test deprecated method still works
```
unho: pytest allows to ensure that this is deprecated. See for an example https://github.com/translate/translate/blob/b512b51d64ab3e3c1a022c5541ac29b0abb91e9a/translate/storage/test_properties.py#L52

Related to that example you might want to drop the now unused `deprecated_call` import in that file.

translate/storage/txt.py
unho: To be clear, in the `dev` release notes.

claudep:
Pushed fixes in two new commits.

translate/storage/dtd.py
```diff
@@ -561,20 +561,16 @@ def parse(self, dtdsrc):
...
+            out.write(unit_str)
+            content += unit_str
+        if not self._valid_store(content):
             warnings.warn("DTD file '%s' does not validate" % self.filename)
-            return None
-        if isinstance(source, six.text_type):
-            return source.encode(self.encoding)
-        return source
-
-    def getoutput(self):
```
unho: If possible this also should be kept, marking it as deprecated.

claudep:
Done (and rebased).

unho:
lgtm

PR #3350, opened 2015-09-17
Transformed most of properties to the decorator syntax

claudep:


translate/storage/aresource.py
```diff
@@ -247,16 +247,16 @@ def escape(self, text, quote_wrapping_whitespaces=True):
...
+    def source(self):
         if (super(AndroidResourceUnit, self).source is None):
             return self.target
         else:
             return super(AndroidResourceUnit, self).source
 
-    source = property(getsource, setsource)
+    @source.setter
+    def source(self, source):
+        base.TranslationUnit.source.__set__(self, source)
```
unho: Why is `base.TranslationUnit` instead of `super(AndroidResourceUnit, self)`?

unho:
Stopped review because I noticed that the API is being changed by the removal of `gettarget` and such.

I personally agree with this removal, since the new properties syntax is much more clear and sane, but I have to raise the warning flag because we are breaking the API. So I suggest to keep the `gettarget` and such, but marking them as deprecated, and moving the logic to the new saner properties syntax (basically converting `gettarget` and such to one liners).

I would also like to ask to put in a separate commit the changes to get rid of direct usage of `gettarget` and such, because that can be landed without breaking anything. Prior review, of course, so we make sure non-wanted changes don't make through.

unho:
I would even suggest to move to a separate PR the changes that get rid of direct usage of `gettarget` and such, since that is a separate cleanup that doesn't depend on the changes that transform the properties to the decorator syntax.

claudep:
That's a good idea. I've now pushed the first part as a separate PR: #3358

unho:
@claudep Why was this closed?

claudep:
Sorry, this was a mistake on my side. I'll resubmit new PRs based on your comments, with deprecation paths.

unho:
Ok, thanks a lot!

claudep:
I am struggling with decorating lambdas with the deprecated decorator. I find it a bit heavy to let both forms for each method. Could you maybe give me an example of how you would solve this issue with deprecation?

unho:
I was thinking on having the new properties syntax while you keep the old `gettarget` (and such) marking them as deprecated (which will be dropped in two versions according to our deprecation policy, and leaving as with just the saner and cleaner new properties syntax). So **no lambdas needed**. I think that the following would just work while keeping previous API:

```python
from translate.misc.deprecation import deprecated


class WhateverUnit(base.TranslationUnit):

    @property
    def source(self):
        # Getter code. Picked from old 'getsource'.

    @property.setter
    def source(self, source):
        # Setter code. Picked from old 'setsource'.

    # Deprecated on 1.14.
    @deprecated("Use unit.source instead")
    def setsource(self, source):
        self.source = source

    # Deprecated on 1.14.
    @deprecated("Use unit.source = value instead")
    def getsource(self, lang=None):
        return self.source
```

unho:
Fixed some minor issues with the sample code in the previous comment.

claudep:
Thanks for the example. Do you want tests/docs for each one?

unho:
All these deprecated `gettarget` (and such) must be listed on release notes with the appropriate replacement. But IMHO no tests are needed since we removed all the usages of this deprecated API in the tests.

translate/storage/lisa.py
```diff
@@ -164,15 +173,7 @@ def settarget(self, text, lang='xx', append=False):
...
         else:
-            self.set_target_dom(None, False)
-
-    def gettarget(self, lang=None):
-        """retrieves the "target" text (second entry), or the entry in the
-        specified language, if it exists"""
-        return self.getNodeText(self.get_target_dom(lang),
-                                getXMLspace(self.xmlelement,
-                                            self._default_xml_space))
-    target = property(gettarget, settarget)
```
claudep: Leandro, I have another question about the lang parameter (and the append for settarget) to several property accessors in this file. In this patch, I removed them because I checked that they were not used inside TTK. Now I realize that some third party may use them. What would you suggest? Ignore those parameters in deprecated shims, or create a new and separate method (not a property), like `target_in_lang(lang)`, or simply keeping the file as is with the `property(getter, setter)` syntax?

unho: IMHO `property(getter, setter)` syntax is an aberration that we should try to get rid of. So lets replace that with the new properties syntax.

As I see we should keep both `gettarget` and the properties (using new syntax), since in this particular scenario we cannot blindly replace the functions with the properties. To my understanding the properties don't allow providing the `lang`, but as you say since we cannot be sure if any third party is using the functions by passing that `lang`, so we cannot drop them.

This is my first impression, and I think that addressing this this way is safe. Eventually I would like to remove those functions, but I am not sure if they provide some key functionality.

unho: As a clarification, this `gettarget` should be kept, and not deprecated, since they have this extra `lang` argument. But the `gettarget` and such with no arguments that are present in the rest of TTK must be marked as deprecated in favor of their properties.

Issue #3275, opened 2015-05-28
prop2po with personality=strings fails on keys with "=" character


jakub-szczepaniak:
I have noticed a problem when converting .strings files. If key and value contain equal sign, then the resulting PO file has part of the key as a text for translation.

In .strings file I have a key 

```properties
/* No comment provided by engineer. */
"Sample key with = sign" = "Sample key with = sign"
```

This string is converted to below message:

```po
#: Sample+key+with
msgid "sign\" = \"Sample key with = sign"
msgstr ""
```

Please also check attached sample files.
version used: 1.9.0

jakub-szczepaniak:
Screenshot from .strings file:
![image](https://cloud.githubusercontent.com/assets/7815188/7857180/d62c0934-052e-11e5-92a5-60f5ebd8bf2e.png)
Screenshot from .po file:
![image](https://cloud.githubusercontent.com/assets/7815188/7857192/e89c423c-052e-11e5-912e-98d593d35349.png)

PR #3351, opened 2015-09-18
Fixing issue #3275

claudep:


translate/storage/properties.py
```diff
@@ -136,66 +136,6 @@
...
-
-
-@deprecated("Use Dialect.find_delimiter instead")
-def find_delimeter(line):
-    """Misspelled function that is kept around in case someone relies on it.
-
-    .. deprecated:: 1.7.0
-       Use :func:`find_delimiter` instead
-    """
-    return _find_delimiter(line, DialectJava.delimiters)
```
unho: Please don't drop this. We have decided to keep the deprecated stuff for a while so people sees the warnings and moves to the recommended way. You can instead change it so it actually uses `Dialect.find_delimiter`.

julen: Wasn't 1.7.0 ages ago? It should be fine to drop [according to what I read](http://docs.translatehouse.org/projects/translate-toolkit/en/stable-1.13.0/developers/deprecation.html#period-of-maintenance).

unho: Also bring back the test for this deprecated function.

unho: Sure, @julen is right.

unho:
lgtm

unho:
But please record the removal of the deprecated function in the release notes.

translate/storage/properties.py
unho: Just for the record, it was deprecated on 1.7.0, but the deprecation warning was added on 1.12. Since we are now on 1.13 and the next release will be 1.14 we have matched the requirement pointed out by @julen.

claudep:
Please review the various release notes additions.

unho:
Release notes reviewed.

lgtm

PR #3352, opened 2015-09-18
OS X Strings (UTF-16) format

pawel-furmaniak:
As outlined at https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html, apple recommends to use UTF-16 for *.strings files

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -423,6 +423,12 @@ class DialectStringsUtf8(DialectStrings):
     default_encoding = "utf-8"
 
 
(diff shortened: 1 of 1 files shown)


claudep:
`DialectStrings` has already an 'utf-16' encoding by default. Did you miss that?

unho:
@claudep is right.

Thanks for your effort. Closing this.

claudep:
Maybe the documentation is not explicit enough about that?

PR #3354, opened 2015-09-21
[WIP][RFC]Add Romanian-specific filters and tests

Jobava:
I am experimenting with adding support for Romanian-specific filters and tests.

This is very preliminary and not ready to merge.

Any kind of pointers are welcome to turn this into a workable changeset.

These changes were added on top of fr.py and test_fr.py in translate/lang

### DIFF
--- a/translate/lang/ro.py
+++ b/translate/lang/ro.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2007-2009 Zuza Software Foundation
+#
+# This file is part of the Translate Toolkit.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>.
+
+"""This module represents the Romanian language.
+
+.. seealso:: https://en.wikipedia.org/wiki/Romanian_language
+"""
+
+import re
+
+from translate.lang import common
+
+
+def ghilimele(text):
+
+    def convertquotation(match):
+        prefix = match.group(1)
+        # Let's see that we didn't perhaps match an XML tag property like
+        # <a href="something">
+        if prefix == u"=":
+            return match.group(0)
+        return u"%s„%s”" % (prefix, match.group(2))
+
+    # Check that there is an even number of double quotes, otherwise it is
+    # probably not safe to convert them.
+    if text.count(u'"') % 2 == 0:
+        text = re.sub('(.|^)"([^"]+)"', convertquotation, text)
+    singlecount = text.count(u"'")
+    if singlecount:
+        if singlecount == text.count(u'`'):
+            text = re.sub("(.|^)`([^']+)'", convertquotation, text)
+        elif singlecount % 2 == 0:
+            text = re.sub("(.|^)'([^']+)'", convertquotation, text)
+    text = re.sub(u'(.|^)“([^”]+)”', convertquotation, text)
+    return text
+
+
+class RoChecker(TranslationChecker):
+
+    @critical
+    def scedilla(self, str1, str2):
+        """Checks whether target text contains a ş or Ş (s-cedilla).
+
+           glyphs incorrectly used instead of the Romanian Șș (s-comma)
+           they are not valid characters for Romanian (only Turkish) 
+           ignore this test if there is a Turkish name or placename
+        """
+        if ("ş" in str2) or ("Ş" in str2):
+            raise SeriousFilterFailure(u"S-cedilla detected: ş or Ş, please"
+                                       "use T-comma instead: ș or Ș")
+        else:
+            return True
+
+
+    @critical
+    def tcedilla(self, str1, str2):
+        """Checks whether target text contains a ţ or Ţ (t-cedilla).
+
+           glyphs incorrectly used instead of the Romanian Țț (t-comma)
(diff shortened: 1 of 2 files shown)


dwaynebailey:
@Jobava if you can fix the issues picked up by [Travis](https://travis-ci.org/translate/translate/builds/81365013) that would be a great first step.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+            return True
+
+
+class ro(common.Common):
+    """This class represents Romanian."""
+
+    validaccel = u"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" + \
+                 u"1234567890"
+    specialchars = u"ȘșȚțĂăÂâÎî«»„”"
+    checker = RoChecker
```
dwaynebailey: So I'm not 100% sure this will work.  Does it follow a pattern for another language?

translate/lang/test_ro.py
```diff
@@ -0,0 +1,57 @@
...
+    language = factory.getlanguage('ro')
+    sentences = language.sentences(u"")
+    assert sentences == []
+
+    sentences = language.sentences(u"Normal case. Nothing interesting.")
+    assert sentences == [u"Normal case.", u"Nothing interesting."]
+    sentences = language.sentences(u"Is that the case ? Sounds interesting !")
+    assert sentences == [u"Is that the case ?", u"Sounds interesting !"]
+
+def test_scedilla():
```
dwaynebailey: You'll want tests here. Have a look in filter/test_checks.py for some examples of how to do tests and how to do them against a specific checker.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+        elif singlecount % 2 == 0:
+            text = re.sub("(.|^)'([^']+)'", convertquotation, text)
+    text = re.sub(u'(.|^)“([^”]+)”', convertquotation, text)
+    return text
+
+
+class RoChecker(TranslationChecker):
+
+    @critical
+    def scedilla(self, str1, str2):
```
dwaynebailey: These might be similar to the validchars check.

An option might be to expand that to be more generic.  I could see individual tests that could:

1. Have a list of chars that are not allowed in the target
2. Have a list of chars that should be different in the target (like your cedilla checks here)

The problem is always foreign words creating false positives.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+        """Checks whether target text contains a ş or Ş (s-cedilla).
+
+           glyphs incorrectly used instead of the Romanian Șș (s-comma)
+           they are not valid characters for Romanian (only Turkish) 
+           ignore this test if there is a Turkish name or placename
+        """
+        if ("ş" in str2) or ("Ş" in str2):
+            raise SeriousFilterFailure(u"S-cedilla detected: ş or Ş, please"
```
dwaynebailey: We reserve SeriousFilterFailures to ones that will break things.  I'd leave this as a FilterFailure, thus it can be ignored as required.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+            raise SeriousFilterFailure(u"T-cedilla detected: ţ or Ţ, please"
+                                       "use T-comma instead: ț or Ț")
+        else:
+            return True
+
+
+class ro(common.Common):
+    """This class represents Romanian."""
+
+    validaccel = u"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" + \
```
dwaynebailey: This can include any characters that would be in the keyboard and could be used as an accelerator key.  Your list is probably correct, just wanted to make sure you understand that if characters like ț and ș as available in one keypress then you can add those also.

translate/lang/ro.py
unho: This seems to come from https://github.com/translate/translate/blob/master/translate/lang/common.py#L194-L197 Anyway there is no single language config setting a custom checker.

translate/lang/ro.py
unho: I guess the same goes for `tcedilla` below.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+
+import re
+
+from translate.lang import common
+
+
+def ghilimele(text):
+
+    def convertquotation(match):
```
unho: Please add docstrings here explaining what does this.

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+
+    validaccel = u"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" + \
+                 u"1234567890"
+    specialchars = u"ȘșȚțĂăÂâÎî«»„”"
+    checker = RoChecker
+
+    @classmethod
+    def characters(cls, text):
+        """Returns a list of characters in text."""
+        return [c for c in cls.character_iter(text)]
```
unho: This looks exactly the same as `Common.characters` so why keep it here?

translate/lang/ro.py
```diff
@@ -0,0 +1,110 @@
...
+        """Implement some extra features for quotation marks.
+
+        Known shortcomings:
+            - % and $ are not touched yet for fear of variables
+            - Double spaces might be introduced
+        """
+        text = super(cls, cls).punctranslate(text)
+        # We might get problems where we got a space in URIs such as
+        # http ://
+        text = text.replace(u"\u00a0://", "://")
```
unho: I would say this is an issue for all languages, so consider proposing this as a fix for `Common` base class.

unho:
@Jobava If you finish this PR soon it probably will make to next release.

unho:
@Jobava Just pinging you again to encourage you to rebase this PR on latest master and address the raised concerns. It would be nice to have this in the next release!

unho:
@Jobava It would be great if you can update this PR. We are approaching the a newer Translate Toolkit release.

Jobava:
Closing this in favour of https://github.com/translate/translate/pull/3543 and others like it (coming soon :) )

Thanks for the patience!

PR #3359, opened 2015-10-09
Fixes for CSV and UTF-8 on Python 2

nijel:
After recent changes the CSV storage doesn't handle UTF-8 input anymore. The problem is in detect_header which now accepted unicode in Python 2, however the csv reader wasn't able to consume unicode input leading to unicode errors.

While working with that, I've also removed BOM logic from CSV as it's no longer necessary here (since #3331 has been merged).

### DIFF
--- a/translate/storage/csvl10n.py
+++ b/translate/storage/csvl10n.py
@@ -21,7 +21,6 @@
 or entire files (csvfile) for use with localisation
 """
 
-import codecs
 import csv
 import six
 from io import StringIO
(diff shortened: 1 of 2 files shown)


claudep:
There is unfortunately a problem with the tests on Python 3, even if it might not be directly related to your changes.
Basically, the issue is with chardet:
```
chardet.detect(b'foo.c:1;test;zkou\xc5\xa1ka')
{'confidence': 0.7183645168039828, 'encoding': 'ISO-8859-2'}
```
Python 2 tests are passing because it does not use the result of `TranslationStore.detect_encoding`, the original bytestring is passed to `csv.StringIO`, while on Python 3 the decoded content is passed.

I wonder if we should always try to decode content with `utf-8` in the first place, then use heuristics detection only when receiving UnicodeDecodeError. What do you think?

nijel:
Hmm, then the Python 3 side also needs some fixes. I've only focused on breakage I had in Python 2 compared to latest release (in 1.13.0 utf-8 CSV files worked fine, with current master they fails with Unicode error). Also adding test with unicode makes these problems more visible in future.

nijel:
Anyway, this seems to be corner case in chardet (which probably should be fixed rather there than in translate-toolkit), which won't happen with more chars, so I've adjusted the testcase to not fire it.

translate/storage/csvl10n.py
```diff
@@ -21,7 +21,6 @@
 or entire files (csvfile) for use with localisation
 """
 
-import codecs
 import csv
 import six
 from io import StringIO
```
claudep: It appears that this import is not used any longer.

claudep:
Merged, thanks!

PR #3361, opened 2015-10-15
Packaging python scripts via `console_scripts` entry point.

layzerar:
`console_scripts` entry point is better supported than regular `scripts`, including windows. So we don't need `py2exe` anymore.

http://python-packaging.readthedocs.org/en/latest/command-line-scripts.html#the-console-scripts-entry-point


layzerar:
Sorry, I should try again.

PR #3362, opened 2015-10-15
Packaging python scripts via `console_scripts` entry point.

layzerar:
`console_scripts` entry point is better supported than regular `scripts`, including windows. So we don't need `py2exe` anymore.

http://python-packaging.readthedocs.org/en/latest/command-line-scripts.html#the-console-scripts-entry-point


layzerar:
It's better to install the package before testing it.

layzerar:
The Travis CI build error seems not concern with this pull request.
```
E   XMLSchemaParseError: Element '{http://www.w3.org/2001/XMLSchema}import': Failed to parse the XML resource 'http://www.w3.org/2001/xml.xsd'., line 4
```

claudep:
I cannot explain the failure either. You might try to force-push to the PR with some minor change to re-trigger a build and see if the error was only a transient one.

layzerar:
It seems a temporary failure.

setup.py
```diff
@@ -428,16 +228,14 @@ def standardsetup(name, version, custompackages=[], customdatafiles=[]):
         packages.append("translate.%s" % subpackage)
 
     datafiles = getdatafiles()
-    dosetup(name, version, packages + custompackages, datafiles + customdatafiles,
-            translatescripts + translatebashscripts)
+    console_scripts = build_console_scripts(translatescripts)
```
unho: Please ensure that `translatescripts` list all the commands that you have removed. For example `idml2po` is missing.

layzerar:
@unho I added all the missing commands just now.

layzerar:
This pull request also fixes the conflict of multiple python.

Scenario:
1. Have `python2.6` and `python2.7`, and `python` is symbol link to `python2.7`
2. Install `traslate-toolkit` on `python2.6`.
3. Run `po2csv`, the script `/usr/bin/env python` will try import  `traslate-toolkit` on `python2.7`.

claudep:
I'm not an expert on packaging, but this PR looks like a nice cleanup. Hopefully merged soon...

unho:
Didn't manage to test it yet. @claudep I would appreciate very much if can you test this PR.

claudep:
@unho sorry, but windows is not my friend. I have not a single windows box in my house.

friedelwolff:
If I look correctly, this removes support for the innosetup (-setup.exe) type builds. Is that intended? If so, I guess some documentation might also need to be updated, although I might simply be out of touch with things.

unho:
@layzerar  The removal of `py2exe` is independent of moving to `console_scripts`, isn't it?

unho:
Merged most of this PR in 40d5dd2d after testing it worked OK.

The part about removing support for py2exe was left out. For now we are encouraging to deploy using pip, but I would appreciate if anyone can shed some light on if py2exe is necessary to deploy on all platforms. This is not my area of expertise.

unho:
I am closing this. If you consider that removing `py2exe` usage is necessary, please create a new PR.

layzerar:
@unho Sorry, I submit a new PR to fix the the innosetup (-setup.exe) type builds.
https://github.com/translate/translate/pull/3378

unho:
@layzerar Thanks. Already merged #3378 too.

I am still unsure about if `py2exe` is still necessary. I just unsuccessfully tried to figure out how to build an installer. It looks simpler to check if installing using `pip` works fine on Windows, which I will try in a moment.

unho:
Just checked that latest translate-toolkit can be installed on Windows using pip and that all commands are accessible. The latest fix is creating the executables for the scripts (converters, tools) so now we don't have to do anything weird, just `pip install package`.

If I am not wrong `py2exe` purpose is to build an installer. I am not sure if that is necessary anymore since now we can recommend users to install Python 2.7 that comes with bundled `pip`. But I guess that using `pip` on a black screen is perhaps too difficult for Windows users.

PR #3365, opened 2015-11-02
Fix up pip install on windows and py2exe with recent setuptools

lambacck:


setup.py
```diff
@@ -20,6 +20,7 @@
 import os
 import re
 import sys
+from setuptools import setup
```
unho: Please put this a few lines below so imports are ordered alphabetically:

```python
from distutils.sysconfig import get_python_lib
from os.path import dirname, isfile, join
from setuptools import setup
```

setup.py
```diff
@@ -456,6 +456,11 @@ def dosetup(name, version, packages, datafiles, scripts, ext_modules=[]):
           ext_modules=ext_modules,
           cmdclass=cmdclass,
           install_requires=parse_requirements('requirements/required.txt'),
+          entry_points={
+              'console_scripts': [
+                  os.path.basename(x) + ' = ' + x.replace(os.path.sep, '.') + ':main' for x in translatescripts
```
unho: Please break this line:

```python
                  os.path.basename(x) + ' = ' + x.replace(os.path.sep, '.') + ':main'
                  for x in translatescripts
```

claudep:
Isn't this patch conflicting with #3362?

setup.py
unho: Shouldn't `translatebashscripts` be included here as well?

unho:
> Isn't this patch conflicting with #3362?

I think so. #3662 looks like it is a nicer cleanup.

lambacck:
#3662 is a nicer cleanup but it means you loose the inslaller and py2exe. That means on Windows users need to install python. That's totally fine for my use case where I'm using it with another python application.  Right now there is no way to install the tool on Windows without knowing to go to download version 1.9 from sourceforge.

unho:
Closed in favor of #3662.

Issue #3374, opened 2016-01-21
pofilter -tprintf problems


asyschikov:
1) printf test passes this pair:

``` po
msgid "Your question % s"
msgstr "Jouw vraag"
```

While "% s" is actually a valid printf formatter, example: https://ideone.com/AZxWh5
From specification: https://en.wikipedia.org/wiki/Printf_format_string#Flags_field 

2) printf passes this pair

``` po
msgid "Your question %"
msgstr "Jouw vraag"
```

Which will break python formatting, Python doesn't allow single % (while it is valid printf). Literal % should be screened as %% in Python. This is not strictly related to printf because it is valid printf but there is no test to validate it.

asyschikov:
2) I looked more into it, it is not only python issue. Literal % chars must be screened as %% in C as well.

asyschikov:
Basically [this regex](https://github.com/translate/translate/blob/61782aa8fcc31bda13ee4897ec2ef4887b5d45e8/translate/filters/checks.py#L54) should include space in flags section.

unho:
I don't think this is a critical failure since IMHO it is a good practice to place "%s" together and not separate like "% s". But if you are willing to contribute a PR with the appropriate changes and append some tests for it in https://github.com/translate/translate/blob/61782aa8fcc31bda13ee4897ec2ef4887b5d45e8/translate/filters/test_checks.py#L584-L636 we would review it.

asyschikov:
@unho it is not good practice but it is critical failure, which I recently witnessed. The text was `Value is %(some_var)s%% something`, translator translated into another language (I will use English) as `Value is %(some_var)s% something`. It passed print_f test, but in this piece there are 2 parameters introduced, and new one is `% s`. And this resulted in error because there was no value provided for second parameter. I looked into print_f standard deeper and it doesn't allow any % signs that are not 1) part of *valid* parameter definition 2) part of %% literal. Self standing % sign is invalid. I will write a PR.

asyschikov:
@unho https://github.com/translate/translate/pull/3375

unho:
As commented in https://github.com/translate/translate/pull/3375#issuecomment-209298996 I am postponing it since the proposed fix introduces a 20% increase in the time it takes to run the printf check. After 2.0 release we can revisit this.

PR #3375, opened 2016-01-24
Bug fix 3374

asyschikov:
1. Add pipe brackets to printf regex to match boost format specification
2. Add space to list of flags in prinf regex to match classical printf specification so that check will catch '% s'
3. Check that all percent signs are either paired (%%) or part of printf placeholder, otherwise fail check.
4. Tests to cover added changes

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -34,7 +34,6 @@ import logging
 import re
 import six
 
-from itertools import izip_longest
 from translate.filters import decoration, helpers, prefilters, spelling
 from translate.filters.decorators import (cosmetic, critical, extraction,
                                           functional)
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -34,6 +34,7 @@ import logging
 import re
 import six
 
+from itertools import izip_longest
 from translate.filters import decoration, helpers, prefilters, spelling
 from translate.filters.decorators import (cosmetic, critical, extraction,
                                           functional)
@@ -55,14 +56,16 @@ printf_pat = re.compile('''
         %(                          # initial %
         (?P<boost_ord>\d+)%         # boost::format style variable order, like %1%
         |
+        \|?                         # boost allows to enclose params in pipes, %|spec|
               (?:(?P<ord>\d+)\$|    # variable order, like %1$s
               \((?P<key>\w+)\))?    # Python style variables, like %(var)s
         (?P<fullvar>
-            [+#-]*                  # flags
+            [+#\- ]*                # flags
             (?:\d+)?                # width
             (?:\.\d+)?              # precision
             (hh\|h\|l\|ll)?         # length formatting
             (?P<type>[\w@]))        # type (%s, %d, etc.)
+        \|?
         )''', re.VERBOSE)
 
 # The name of the XML tag
@@ -977,6 +980,48 @@ class StandardChecker(TranslationChecker):
 
         if (count1 or count2) and (count1 != count2):
             raise FilterFailure(u"Different number of printf variables")
+
+        # Find percent signs that are not followed by even number of percent signs
+        def find_lone_percents(string):
+            count_signs = 0
+            for pos, c in enumerate(string):
+                if not c == '%':
+                    continue
+
+                if pos + 1 >= len(string) or string[pos + 1] != '%':
+                    if count_signs % 2 == 0:
+                        yield pos
+                    count_signs = 0
+                else:
+                    count_signs += 1
+
+        # check that all odd percent signs are parts of placeholders
+        def has_unmatched_percents(string):
+            lone_signs = find_lone_percents(string)
+            placeholders = [m.span() for m in printf_pat.finditer(string)]
+            ph_ind = 0
+            ph = placeholders[0] if placeholders else None
+            for sign_pos in lone_signs:
+                if not placeholders:
+                    return True
+
+                if ph[0] <= sign_pos < ph[1]:
+                    continue
+
+                while sign_pos > ph[0]:
+                    ph_ind += 1
+                    if ph_ind >= len(placeholders):
+                        return True
+
+                    ph = placeholders[ph_ind]
+
+                if not (ph[0] <= sign_pos < ph[1]):
+                    return True
+
+            return False
+
+        if has_unmatched_percents(str1) or has_unmatched_percents(str2):
+            raise FilterFailure(u"One of the strings contains percent sign that is neither part of placeholder or %% literal")
 
         return 1
 
--- a/translate/filters/test_checks.py
+++ b/translate/filters/test_checks.py
@@ -592,6 +592,10 @@ def test_printf():
     assert fails(stdchecker.printf, "I am %#100s", "Ek is %10s")
     assert fails(stdchecker.printf, "... for user %.100s on %.100s:", "... lomuntu osebenzisa i-%. I-100s e-100s:")
     assert passes(stdchecker.printf, "%dMB", "%d MG")
+    assert fails(stdchecker.printf, "Discount 10% something", "Korting tien procent")
+    assert fails(stdchecker.printf, "Discount 10%%", "Koring 10%")
+    assert fails(stdchecker.printf, "%%$", "%$")
+    assert passes(stdchecker.printf, "%%$", "%%$")
     # Reordering
     assert passes(stdchecker.printf, "String %s and number %d", "String %1$s en nommer %2$d")
     assert passes(stdchecker.printf, "String %1$s and number %2$d", "String %1$s en nommer %2$d")


translate/filters/checks.py
```diff
@@ -55,14 +55,16 @@
         %(                          # initial %
         (?P<boost_ord>\d+)%         # boost::format style variable order, like %1%
         |
+        \|?                         # boost::format allows to enclose params in pipes, %|spec|
```
unho: Can you please add tests for enclosing params in pipes?

unho: If no tests are added for this feature I prefer to leave it out for now. If tests are added I prefer tests and this particular change in the regex to be put in a separate PR.

translate/filters/checks.py
```diff
@@ -55,14 +55,16 @@
...
               (?:(?P<ord>\d+)\$|    # variable order, like %1$s
               \((?P<key>\w+)\))?    # Python style variables, like %(var)s
         (?P<fullvar>
-            [+#-]*                  # flags
+            [+#\- ]*                # flags
             (?:\d+)?                # width
             (?:\.\d+)?              # precision
             (hh\|h\|l\|ll)?         # length formatting
             (?P<type>[\w@]))        # type (%s, %d, etc.)
+        \|?
```
unho: Is this related to boost:format change above? If yes, please add a proper comment in this line. Also this change should preferably be part of a separate PR.

translate/filters/checks.py
```diff
@@ -55,14 +55,16 @@
         %(                          # initial %
         (?P<boost_ord>\d+)%         # boost::format style variable order, like %1%
         |
+        \|?                         # boost::format allows to enclose params in pipes, %|spec|
               (?:(?P<ord>\d+)\$|    # variable order, like %1$s
               \((?P<key>\w+)\))?    # Python style variables, like %(var)s
         (?P<fullvar>
-            [+#-]*                  # flags
+            [+#\- ]*                # flags
```
unho: This looks like an unrelated fix. Please put this on a commit of its own, and preferably on its own so we can quickly merge it.

translate/filters/checks.py
```diff
@@ -978,6 +980,48 @@ def printf(self, str1, str2):
...
+
+                    ph = placeholders[ph_ind]
+
+                if not (ph[0] <= sign_pos < ph[1]):
+                    return True
+
+            return False
+
+        if has_unmatched_percents(str1) or has_unmatched_percents(str2):
+            raise FilterFailure(u"One of the strings contains percent sign that is neither part of placeholder or %% literal")
```
unho: This is a very long line (it exceeds 80 chars). Can you please wrap it like the following:

```python
            raise FilterFailure(u"One of the strings contains a percent sign "
                                u"that is neither part of placeholder or %% "
                                u"literal")
```

translate/filters/test_checks.py
```diff
@@ -592,6 +592,10 @@ def test_printf():
     assert fails(stdchecker.printf, "I am %#100s", "Ek is %10s")
     assert fails(stdchecker.printf, "... for user %.100s on %.100s:", "... lomuntu osebenzisa i-%. I-100s e-100s:")
     assert passes(stdchecker.printf, "%dMB", "%d MG")
+    assert fails(stdchecker.printf, "Discount 10% something", "Korting tien procent")
+    assert fails(stdchecker.printf, "Discount 10%%", "Koring 10%")
+    assert fails(stdchecker.printf, "%%$", "%$")
```
unho: Perhaps I am looking at this the wrong way, but when a `%%` literal is present the `printf` should just skip it and don't consider it a failure. So this two tests above should be passing.

unho:
Please put a reference on the commit message about it fixing the issue, like `Fixes #3374`, so the issue is autoclosed when we merge the PR.

unho:
I am wondering if "unmatched percent sign" should actually be a separate check. If I am not wrong current code for "printf" check doesn't detect any error on strings like "10% something", so it would be ok to convert this in a different check.

@dwaynebailey Any input on this?

translate/filters/checks.py
asyschikov: @unho [there was actually a test for it](https://github.com/translate/translate/blob/master/translate/filters/test_checks.py#L629):
```python
assert passes(stdchecker.printf, "(x,y) = (%|+5|,%|+5|)", "(x,y) = (%|+5|,%|+5|)")`
```
 although that test tested nothing because pipe-brackets were not incorporated in regex, so according to current version of checker this string doesn't have any params. After I implemented a rule "no lone percent signs" this test started to rightfully fail that's why I had to add pipe brackets. So that's why this can't be separated into two PRs. I just added test for this feature: https://github.com/asyschikov/translate/commit/bbd102e5a0e6c8b639c3c26882449db61cc7a437

translate/filters/checks.py
asyschikov: @unho this may look like unrelated fix but they are all related, this is first thing I did and then I had to fix others otherwise it won't work properly, because printf test will fail this string: `Down payment is %d%% something`. It will match '% s' as placeholder and fail, while it is actually part of `%%` and it is a valid string. I wanted to fix everything related to that percent things at once, because this sole fix will introduce a problem immediately (we have a string that is like that for example).

asyschikov:
@unho it is up to you of course, but printf function (and python analog) will treat `%d% something` as 2 placeholders:

```python
>>> '%d% something' % (10, )
  File "<stdin>", line 1, in <module>
TypeError: not enough arguments for format string
>>> '%d% something' % (10, 'asd')
'10asdomething'
```

translate/filters/checks.py
unho: Thanks.

unho:
> @unho it is up to you of course, but printf function (and python analog) will treat %d% something as 2 placeholders:

As I said the error is not some printf placeholder being malformed, it is about lone percent signs since it can be either an unescaped percent or a malformed printf placeholder. Thus my suggestion of adding a separate check for it since that way we can report the specific problem "The percent sign is either incorrectly escaped or it is part of a malformed printf placeholder".

Please do fix the other raised issues.

translate/filters/test_checks.py
asyschikov: Goal of this test `"%%$", "%$"` was to check that it will fail on second string because `%` is not part of `%%` pair and not part of valid placeholder. I consider such string malformed (so does python):

```python
>>> "%d %$" % (1,)
  File "<stdin>", line 1, in <module>
TypeError: not enough arguments for format string
>>> "%d %$" % (1, 2)
  File "<stdin>", line 1, in <module>
ValueError: unsupported format character '$' (0x24) at index 4
```

One above in same vein but malformation is caused by `%` followed by EOL:

```python
>>> "Koring 10%" % ()
  File "<stdin>", line 1, in <module>
ValueError: incomplete format
>>> "Koring 10%" % (1,)
  File "<stdin>", line 1, in <module>
ValueError: incomplete format
```

asyschikov:
@unho right - error is not with particular placeholder, problem with the string as a whole because it will break printf/python formatting. And I added specific error explaining the problem (and made line shorter) https://github.com/translate/translate/commit/cb527b91d8d829fe19dcc9435a5fa9c10a27445f

translate/filters/test_checks.py
unho: You are right about me looking at it in the wrong way. But I just run this specific example and it returns `False`, so it is not working properly.

unho: mmm, forget about that, double checking.

unho: It is working properly. My most sincere apologies.

translate/filters/checks.py
unho: Ok. I wonder now if we should make this flexible by using `\s` instead of ` ` (whitespace).

unho:
@friedelwolff Do you mind create an issue for your valuable input?

phlax:
@friedelwolff github is defo good for discussing specific issues - but for more immediate discussion we are generally on https://gitter.im/translate/dev

unho:
@friedelwolff Created issue #3412 to track what you suggested in your last comment.

unho:
I did some research and here are some timings:

| Before | After|
| --- | --- |
| 26,10	| 31,69|
| 26,58	| 31,30|
| 26,49	| 31,37|
| 26,39	| 31,92|
| 26,43	| 31,70|

So average times are:

| Before | After|
| --- | --- |
|26,40 | 31,60|

Which means that with these changes the `printf` check takes 19.7% more time with respect to master.

Given that I wonder if we should consider adding instead a separate check to begin work towards #3412. I am putting this on hold for now until we figure out which is the best way to proceed given that we are working towards releasing a final 2.0 TTK release.

PR #3376, opened 2016-02-02
Update jsonl10n.py

EmilTemirov:
Added ability to parse ES6 based js translation.

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -215,7 +215,7 @@ class JsonFile(base.TranslationStore):
         if isinstance(input, bytes):
             input = input.decode('utf-8')
         try:
-            input = re.sub(r"\s*?export default\s*?\{", "{", input)
+            input = re.sub(r"^\s*?export default\s*?\{", "{", input)
             self._file = json.loads(input)
         except ValueError as e:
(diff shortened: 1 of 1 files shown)


unho:
Please add some tests in `translate/storage/test_jsonl10n.py`. Also please squash the commits and rebase on top of latest master.

unho:
@EmilTemirov Why did you close this pull request?

EmilTemirov:
@unho I figured out that ES6 based js translation is fundamentally different than JSON file. I wanted to integrate Translate Toolkit with https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Defining-Translations but in this case, you have no restrictions on the format of your translations.js file. As you can see in examples, there are used single quotes for keys and values that are not correctly for JSON format. 
I found another solution for my project (http://i18next.com/translate/) that supports JSON for translations.

unho:
Actually that is not a translation format, but JavaScript code. The translations are defined in a way that looks like JSON. Probably TTK would work with your patch, so did you actually tried to use it?

EmilTemirov:
Yes, I did. It works but you should remember about double quotes... Otherwise, you will get a JSON Parse exception.

PR #3377, opened 2016-02-04
Do not fail on unknown locale

nijel:
There are locales which Babel doesn't know but are valid and crashing on
these is not nice. The code now falls back to using plurals as for
English.

Signed-off-by: Michal Čihař <michal@cihar.com>


unho:
Merged. Thanks.

PR #3378, opened 2016-02-17
Fix innosetup (-setup.exe) type builds

layzerar:


unho:
Merged in c264cea6. Thanks!

Note that I haven't tested it.

unho:
Just tested installing this on Windows using pip and it works fine. I can finally install translate-toolkit on Windows without any kind of magic (I am guessing that in the past I installed Python 2.7 which came with pip bundled). I suppose that `py2exe` is still necessary, but at least now it is possible to get newer Translate Toolkit installable on Windows again.

So thank you very much @layzerar!

layzerar:
I'm glad to see it help. Mayby innosetup builds is useful for someone who don't have python.

unho:
Yes, Inno Setup is useful for people who don't have Python and pip installed, and that don't know how to use command line. In other words, most Windows users. In the other hand Translate Toolkit is a set of command line tools, so I am not sure installing from command line with pip is such a big issue.

unho:
Anyway this PR actually helped when installing using pip, since now we get actual `.exe` executables for Translate Toolkit commands.

PR #3379, opened 2016-02-18
Replace optparse by argparse

claudep:
Updates and replaces PR #90.

PR #3380, opened 2016-02-18
Replace optparse by argparse

claudep:
Replaces #3379


claudep:
@dwaynebailey @unho There are some test errors with CLI tests related to the change of INPUT/OUTPUT being now purely positional (and no more -i / -o). Before fixing them, I'd like to be sure that it's an approved change.

unho:
@claudep Sorry for the delay. Yes, tests need to pass, so adjust what needs adjusting in separate commits in this PR so we can review those changes separately.

dwaynebailey:
@claudep from my side we need to move to argparse, so whatever is needed to be changed is fine by me.  I'm really not sure how far I got along the optparse -> argparse migration, I fear there is still a lot of cruft in optrecurse (which itself likely needs refactoring based on what argparse can provide, but that's likely a separate PR)

translate/misc/optrecurse.py
```diff
@@ -32,26 +32,28 @@
...
-        if action == "manpage":
-            parser.print_manpage()
-            sys.exit(0)
-        return super(ManPageOption, self).take_action(action, dest, opt, value,
-                                                      values, parser)
+    def __init__(self,
+                 option_strings,
+                 dest,
+                 help=None):
+        super(ManPageAction, self).__init__(option_strings, dest, help=help)
```
unho: I wonder if it is actually necessary to `__init__` here.

translate/convert/test_convert.py
```diff
@@ -113,20 +113,14 @@ def test_help(self):
...
+        options = self.help_check(options, "--progress PROGRESS")
         options = self.help_check(options, "--version")
         options = self.help_check(options, "-h, --help")
         options = self.help_check(options, "--manpage")
-        options = self.help_check(options, "--errorlevel=ERRORLEVEL")
-        options = self.help_check(options, "-i INPUT, --input=INPUT")
-        options = self.help_check(options, "-x EXCLUDE, --exclude=EXCLUDE")
-        options = self.help_check(options, "-o OUTPUT, --output=OUTPUT")
+        options = self.help_check(options, "--errorlevel ERRORLEVEL")
+        options = self.help_check(options, "-x EXCLUDE, --exclude EXCLUDE")
```
unho: This kind of changes might require adjusting the documentation as well.

translate/tools/pretranslate.py
```diff
@@ -210,15 +210,15 @@ def main(argv=None):
...
+    parser.add_argument("--tm", dest="tm", default=None,
                       help="The file to use as translation memory when fuzzy matching")
     parser.passthrough.append("tm")
     defaultsimilarity = 75
-    parser.add_option("-s", "--similarity", dest="min_similarity",
-                      default=defaultsimilarity, type="float",
-                      help="The minimum similarity for inclusion (default: %d%%)" % defaultsimilarity)
+    parser.add_argument("-s", "--similarity", dest="min_similarity",
+                      default=defaultsimilarity, type=float,
+                      help="The minimum similarity for inclusion (default: %(default)s)")
```
unho: Perhaps it is better to keep the `%%`. Otherwise remove it in the other appearance where `%%` was kept.

translate/misc/optrecurse.py
```diff
@@ -131,12 +136,18 @@ def format_manpage(self):
             result.append('\n\n'.join([re.sub('\.\. note::', 'Note:', l)
                                               for l in description_lines]))
         result.append('.SH OPTIONS\n')
-        ManHelpFormatter().store_option_strings(self)
+        #ManHelpFormatter().store_option_strings(self)
```
unho: Please don't keep commented code.

unho:
I am a bit reluctant to land this before the next release because:

- how invasive these changes are,
- perhaps there are still [additional changes to be done](https://github.com/translate/translate/pull/3380#issuecomment-185894441),
- perhaps there are not enough tests to exercise all command line arguments,
- maybe docs need to be altered as well.

claudep:
After having a hard time digging into remaining issues, I can only give you reason. I'd like to clean some stuff in option parsing before continuing this work. #3381 might be a first step.

unho:
@claudep It seems that progress on this has stagnated. Please do not think of my comments about discouraging on moving forward with this, but rather encouraging on getting this landed on smaller chunks if possible.

claudep:
I'm a fan of smaller chunks when possible. Buts as soon as you switch optparse by argparse, you are forced to update a whole bunch of other stuff to keep things working.

unho:
@claudep Please do the required cleanups and update this PR when you have time.

PR #3381, opened 2016-02-18
Clean inputformats/outputoptions

claudep:


translate/tools/poconflicts.py
```diff
@@ -65,13 +65,6 @@ def set_usage(self, usage=None):
         else:
             super(ConflictOptionParser, self).set_usage(usage)
 
-    def run(self):
```
unho: Is it ok to remove the whole method?

claudep: I see no differences with the same method of the parent class.

unho: You are right.

unho:
I don't see anything wrong. But I can't understand the purpose of the change and I don't feel confident to provide a good to merge. Also this conflicts with work in #3380. So I would appreciate some explanation.

claudep:
Sure. No offense intended, but currently the code that parses arguments and prepares input/output/template files is in bad shape, spaghetti-like. When working on #3380 where suddenly `options.input` becomes a list with one element (if nargs=1), I found it very difficult to adapt the parsing code for that. My intent is to incrementally simplify/refactor that code (mainly `translate/misc/optrecurse.py` and `translate/convert/convert.py`) to be able to more easily adapt and fix things. For example, just the fact that `options` is a mix of option parsing result and a mutable structure updated by side-effects in a bunch of different methods is unfortunate at least. In my eyes, this part of the toolkit is almost unmaintainable currently. That's a proposal and my personal view, you can disagree, of course.

claudep:
Oh, and of course I volunteer to rebase #3380 after that process :-)

unho:
@claudep lgtm. Thanks a lot!

dwaynebailey:
/me polishes medal of valour for @claudep

dwaynebailey:
@claudep I got lost/overwhelmed when looking at the optrecurse part of things.  It's trying to do what argparse does with positional arguments as its pre argparse days.

So it does need a knife taken to it.  To rather use argparse in these situations and remove the custom unmaintained logic.

claudep:
Thanks for the medal :-)

The problem is that optrecurse is not only handling arguments, it's also setting up things like opening files and defining a global environment for the commands. I will try to continue doing incremental changes like this one until we have separated things more cleanly.

PR #3386, opened 2016-03-01
Command docs cleanups

unho:
There is still much more cleanup to do, but this is a first stage.

### DIFF
--- a/docs/commands/csv2po.rst
+++ b/docs/commands/csv2po.rst
@@ -1,4 +1,3 @@
-
 .. _csv2po:
 .. _po2csv:
 
@@ -14,15 +13,17 @@ your PO translations.
 If you are starting out with your own CSV files (not created by po2csv), take
 note of the assumptions of the column layout explained below.
 
+
 .. _csv2po#usage:
 
 Usage
 =====
 
-::
+.. code-block:: console
 
   csv2po [options] <csv> <po>
   po2csv [options] <po> <csv>
+
 
 Where:
 
@@ -31,6 +32,7 @@ Where:
 +--------+----------------------------------------------+
 | <po>   | is a file or directory containing PO files   |
 +--------+----------------------------------------------+
+
 
 Options (csv2po):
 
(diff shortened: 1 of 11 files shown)


docs/commands/csv2po.rst
```diff
@@ -88,34 +92,45 @@ The resultant CSV file has the following layout
...
+These examples demonstrate the use of csv2po:
+
+.. code-block:: console
 
   po2csv -P pot csv
 
-We use the :opt:`-P` option to recognise POT files found in *pot* and convert
-them to CSV files placed in *csv*::
+
+We use the :option:`-P` option to recognise POT files found in :file:`pot` and
```
dwaynebailey: Using `:option:` without the associated `.. option::` breaks latest Sphinx.  You're only not seeing that because Sphinx is currently limited to not use the latest version. Sadly that commit that limits it provides no clarity on why.  Personally with the current task list this doesn't seem like a great fix considering that it actually is making things worse in terms of a proper fix.

phlax: not pinning the upstream apps causes ongoing problems. Often devs want to get on with working on their code not worry about what has changed in upstream versions. When things do break - the easiest fix is to pin it to the known working version.

If the known working upstream has no known security issues *this is a perfectly valid strategy*

I am not sure i follow what you are saying about commits - i figure it would have been a lot clearer if you could have provided a link or some kind of reference.

unho: I will undo the `:opt:` changes.

docs/commands/csv2po.rst
```diff
@@ -14,16 +13,18 @@ your PO translations.
...
 note of the assumptions of the column layout explained below.
 
+
 .. _csv2po#usage:
 
 Usage
 =====
 
-::
+.. code-block:: console
```
dwaynebailey: Can you point to a rendering of this.  ``::`` was meant to make this fixed width.  Using `console` code block is for command line code.  This isn't that.  So I'd like to see that the rendering is actually better, not worse.  I'm happy either way but really the simple question is this: "To the reader of the docs is this more helpful", I don't see that question even asked in this PR to motivate the change.

docs/commands/csv2po.rst
```diff
@@ -88,34 +92,45 @@ The resultant CSV file has the following layout
...
+
 .. _csv2po#examples:
 
 Examples
 ========
 
-These examples demonstrate the use of csv2po::
+These examples demonstrate the use of csv2po:
+
+.. code-block:: console
```
dwaynebailey: Can we rather use ``.. hightlight:: console`` to set the highlighting for the file.  Adding all these codeblocks actually makes it harder to read the raw text.

docs/commands/csv2po.rst
```diff
@@ -88,34 +92,45 @@ The resultant CSV file has the following layout
...
 
   po2csv -P pot csv
 
-We use the :opt:`-P` option to recognise POT files found in *pot* and convert
-them to CSV files placed in *csv*::
+
+We use the :opt:`-P` option to recognise POT files found in :file:`pot` and
+convert them to CSV files placed in :file:`csv`:
```
dwaynebailey: I'm concerned about using things like ``:file:``, not that they're wrong, but just that `_` and `*` are maybe a little clearer to document writers.  Since these aren't even files but are just extensions.  Try read that sentence quickly and you spend more time parsing the markup then reading the content.

dwaynebailey:
LGTM - for no other reason than I don't think this should block the release.

General observation.

* Some of these changes do not increase readability
  * .. code-block - is not doing what is wanted for those sections and making it harder to read the code
  * In line formatting removed for `:xxx:` markers makes actually reading the raw docs much much harder.

So I'm not sure most of the changed are achieving what is set out.  Plus not sure that most of the issues I have above are in any way blockers to a ttk release.

docs/commands/csv2po.rst
unho: These are not extensions but filenames, `pot` is the output POT file and `csv` is the input CSV file.

dwaynebailey: Ah true.  But my central point is readability. I have to now parse that string more slowly and carefully because of those markers.  So I don't agree that they are adding any value.

unho:
Directly landed the command option changes after checking with Dwayne.

docs/commands/csv2po.rst
unho: The point was to use `code-block` instead of less visible `::`.

docs/commands/csv2po.rst
unho: Actually it is not clear if those are filenames or directories, but in any way we can consider those to be filepaths and highlight them with `:file:` so readers notice that.

docs/commands/csv2po.rst
unho: I prefer to be explicit.

docs/commands/csv2po.rst
unho: Just checked and I don't see any rendering difference. As I said I find `code-block` to be more visible and explicit than `::` and since this is listing the way to run the commands the *console* highlight seems appropriate.

docs/commands/csv2po.rst
unho: I wanted to say that from the reading these are clearly directories.

unho:
Landed some typos and minor cleanups. There is no agreement regarding the rest of the changes so I am closing the PR.

PR #3393, opened 2016-03-10
Add tbx2po

dwaynebailey:
This adds a basic tbx2po converter.  It's aimed at Microsoft TBX files conversion and nothing much else.

### DIFF
--- a/translate/convert/tbx2po.py
+++ b/translate/convert/tbx2po.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2006-2007 Zuza Software Foundation
+#
+# This file is part of translate.
+#
+# translate is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# translate is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>.
+
+"""Convert TermBase eXchange (.tbx) glossary file into a Gettext PO file
+
+See: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/tbx2po.html
+for examples and usage instructions
+"""
+
+from translate.storage import po, tbx
+
+
+class tbx2po(object):
+    """A class that takes translations from a .tbx file and puts them in a .po
+    file
+    """
+
+    def convertfile(self, tbxfile):
+        """Converts a tbxfile to a tbxfile, and returns it. uses templatepo if
+        given at construction
+        """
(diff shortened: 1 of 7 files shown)


unho:
Please directly land the reflow commit.

translate/convert/tbx2po.py
```diff
@@ -0,0 +1,76 @@
...
+"""Convert TermBase eXchange (.tbx) glossary file into a Gettext PO file
+
+See: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/tbx2po.html
+for examples and usage instructions
+"""
+
+from translate.storage import po, tbx
+
+
+class tbx2po:
```
unho: Inherit from `object`.

translate/convert/tbx2po.py
```diff
@@ -0,0 +1,76 @@
...
+            term.addnote("Part of speech: %s" % tbxunit.getnotes('pos'), 'developer')
+            self.pofile.addunit(term)
+        self.pofile.removeduplicates()
+        return self.pofile
+
+
+def converttbx(inputfile, outputfile, templatefile, charset=None,
+               columnorder=None):
+    """Reads in inputfile using csvl10n, converts using csv2tbx, writes to
+    outputfile
```
unho: Adjust docstring to actual behavior.

translate/convert/tbx2po.py
unho: Probably in this case it doesn't make any sense to use a class with just one method. Looks simpler to just have a function.

dwaynebailey: It's a pattern inherited from all other converters.  I'm not going to deviate from that,

unho:
It would be great to have tests for the notes functionality and the converter, but given how faulty TBX support is in TTK I don't want to block on this.

I just tested this with 40 TBX glossaries I have and I have to say it works, but not always in the way I would like to. Again this is because faulty TBX support. But there are two things that bother me and perhaps are issues you wanna look at:

- When there are more than 2 languages you are not able to tell which ones should be extracted to the bilingual PO.
- When there are more than one translation (term) for the same entry (concept) in the same language only one of them is extracted. A good example of that is an entry about keyboard keys, for example Escape which might have two terms: `Esc` and `Escape` (abbreviation and full form).

translate/convert/tbx2po.py
unho: I am ok with keeping the pattern as long as we don't introduce a class that doesn't inherit from `object`, which is not a good coding practice.

dwaynebailey:
@unho noting your observations

* Yes there is no way to indicate a language in extraction.  When we find files in the wild that need that we can add it.
* Disambiguating is hard. Thus if there are more than one term for the exact item we need to find other disambiguating information that we can use in the msgctxt to make the PO valid.  If that doesn't exist, then there is nothing we can actually do.

unho:
> * Disambiguating is hard. Thus if there are more than one term for the exact item we need to find other disambiguating information that we can use in the msgctxt to make the PO valid.  If that doesn't exist, then there is nothing we can actually do.

Disambiguating is not necessary. In current terminology you do this when you have to provide several translations for the same concept (entry): https://mozilla.locamotion.org/af/terminology/translate/#search=%3B&sfields=source,target

unho:
gtm provided issues are created for the concerns.

PR #3395, opened 2016-03-11
JSON: Use OrderedDict to maintain unit order

### LINKED ISSUE #3394: json2po output strings in wrong order
Running `json2po` outputs a valid PO files, but the strings are in the wrong order. I’m not sure exactly *which* order it is, but is *neither* the original order, the reverse order or alphabetically. Here’s an example JSON file:

```json
    {
        "foo": "foo",
        "bar": "bar",
        "baz": "baz"
    }
```

Running `json2po` on this file results in the following (with a PO file header):

```po
    #: .baz
    msgid "baz"
    msgstr ""
    
    #: .foo
    msgid "foo"
    msgstr ""
    
    #: .bar
    msgid "bar"
    msgstr ""
```

Having the strings in a different order than in the original file makes it much harder to correctly translate the file, as one loses context.

I observe this bug both in version 1.13.0 and 1.14.0-rc1.

dwaynebailey:
By default JSON will use a normal dict for objects.  Overridding this
and using OrderedDict means we can maintain order and thus translator
context.

Thanks http://stackoverflow.com/a/6921760/2167584

Fixes #3394

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -70,6 +70,12 @@ TODO:
 
 import json
 import os
+try:
+    from collections import OrderedDict
+except ImportError:
+    # Python 2.6 does not have OrderedDict and also can't use it in
+    # json.loads()
+    OrderedDict = None
 import six
 
 from translate.storage import base
@@ -214,7 +220,11 @@ class JsonFile(base.TranslationStore):
         if isinstance(input, bytes):
             input = input.decode('utf-8')
         try:
-            self._file = json.loads(input)
+            if OrderedDict is not None:
+                self._file = json.loads(input, object_pairs_hook=OrderedDict)
+            else:
+                # object_pairs_hook is not present in Python 2.6
+                self._file = json.loads(input)
         except ValueError as e:
             raise base.ParseError(e.message)
 
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -67,6 +67,7 @@ Formats and Converters
 - JSON
 
   - Output now includes a trailing newline.
+  - Unit ordering is maintaned (:issue:`3394`).
 
 
 Filters and Checks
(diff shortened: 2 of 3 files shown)


requirements/required.txt
```diff
@@ -2,6 +2,7 @@
 # standard library in Python >= 2.7
 argparse
 six>=1.9
+ordereddict ;python_version <= '2.6'       # JSON
```
unho: I would remove the JSON comment. This package might be necessary to address #3396 in a proper way.

dwaynebailey: That can happen when that issue is addressed.

unho: I am ok with that.

translate/storage/jsonl10n.py
```diff
@@ -70,6 +70,11 @@
 
 import json
 import os
+try:
+    from collections import OrderedDict
+except ImportError:
+    # Python 2.6 does not have OrdereDict and also can't use it in json.loads()
```
unho: Typo: OrderedDict

unho: Also json.loads() can't use OrderedDict in 2.6, even if it were available.

setup.py
```diff
@@ -398,8 +398,9 @@ def parse_requirements(file_name):
     requirements = []
     with open(file_name, 'r') as fh:
         for line in fh:
-            # Ignore comments, blank lines and included requirements files
-            if re.match(r'(\s*#)|(\s*$)|(-r .*$)', line):
+            # Ignore comments, blank lines, included requirements files,
+            # extended requirements syntax
+            if re.match(r'^(\s*#|\s*$|-r .*$|.*;)', line):
```
unho: Is this change related to the issue. If not please put in separate commit.

dwaynebailey: Nah I can lose this now

unho:
lgtm if Travis passes.

phlax:
lgtm

PR #3398, opened 2016-03-14
Docs: pot2po can be used to update against templates for other formats

unho:


dwaynebailey:
While this is fine to merge.  I think the better solution is to address the core issue.  The command name is confusing.  Just renaming/aliasing the command would prevent the confusion that you are trying to document away.

I always says that the best documentation is no documentation.  A more self explanatory command wouldn't require us documenting that other formats are supported.

I had hoped to address this in a future ttk that used subcommands.

unho:
@dwaynebailey It is indeed a good idea to provide a command alias for `pot2po`, but I couldn't find any good name. If you find one I can look into it.

Merging this PR.

PR #3400, opened 2016-03-14
JSON: keep order when serializing

### LINKED ISSUE #3394: json2po output strings in wrong order

dwaynebailey:
Followon form 219708fa2d1c44c57aa5eebc8e3edb533d3bdf3e for back conversion.

Fixes #3394

### DIFF
--- a/translate/convert/test_po2json.py
+++ b/translate/convert/test_po2json.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+import sys
+import pytest
+
 from translate.convert import po2json
 from translate.misc import wStringIO
 from translate.storage import po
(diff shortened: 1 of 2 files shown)


unho:
lgtm

unho:
Would be nice to document this on dev release notes.

dwaynebailey:
@unho "Unit ordering is maintained" is already in the release notes.  Unless you had some other aspect you felt needed documenting.  I think that is enough.

unho:
@dwaynebailey No, that might be enough. Anyway that will be checked again at release time, as usual. As I said earlier, lgtm.

PR #3403, opened 2016-03-15
Cleanup: class inherit from object

dwaynebailey:
Changs `class X:` -> `class X(object):`


claudep:
Note that it is good practice on Python 2, but Python 3 doesn't need that because inheriting from object is implicit. This could then be reverted when Python 2 support is dropped.

unho:
As long it is not a problem I prefer explicit to implicit.

lgtm

PR #3405, opened 2016-03-16
Fixed comparison regression with multistrings

### LINKED ISSUE #3404: Backwards incompatible change in multistring
The behavior of the `multistring` class with respect to string comparisons has changed in 1.14-rc1, more specifically when the `autoencode` class was removed in 4f9ee7edfc9. At least this is the case in Python 2.7.

I have experienced this in some custom Pootle code when testing out TTK from master — a multistring with non-ASCII characters was being compared against the empty byte string `''`, and it resulted in a encoding error. I can reproduce this in a test case:

```python
from translate.misc import multistring

def test_cmp():
    t = multistring.multistring
    s1 = t("test")
    assert s1.__cmp__("test") == 0

    s2 = t(u"téßt")
    assert s2.__cmp__("test") > 0
```
(Note that for 1.13.x the multistring constructor needs an `encoding='UTF-8'` parameter in the test).

I'm not sure if this could be considered a bug, after all my understanding is apps/libraries should internally work with code points and not bytes, however this is a breaking change that might potentially affect existing consumer code so I consider it's worth bringing this up to hear people's thoughts.

claudep:
Fixes #3404.
Thanks Julen Ruiz Aizpuru for the report and the tests.

### DIFF
--- a/translate/misc/multistring.py
+++ b/translate/misc/multistring.py
@@ -55,7 +55,7 @@ class multistring(six.text_type):
         elif isinstance(otherstring, six.text_type):
             return cmp_compat(six.text_type(self), otherstring)
         elif isinstance(otherstring, bytes):
-            return cmp_compat(bytes(self), otherstring)
+            return cmp_compat(self.encode('utf-8'), otherstring)
         elif isinstance(otherstring, list) and otherstring:
             return cmp_compat(self, multistring(otherstring))
         else:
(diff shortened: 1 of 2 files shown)


unho:
Looks good. Would like to hear the input from @julen.

unho:
@claudep As pointed out in https://github.com/translate/translate/issues/3404#issuecomment-197503936 this works. So lgtm from my side.

PR #3414, opened 2016-03-30
Placeables: Skip non-placeholders with whitespace

### LINKED ISSUE #3368: False positive in FormattingPlaceable
In the following example, "% s" is detected as a placeable:
"xx% support a Web" -> "xx<mark>% s</mark>upport a Web"

The regex should be able to detect the whitespace between % and s and abort.

More specifically, the bug is probably in:
https://github.com/translate/translate/blob/master/translate/storage/placeables/general.py#L150

unho:
This fix prevents from detecting strings like "% s" to be format placeholders.

Fixes #3368.


dwaynebailey:
@unho mmm you mean except that space in that place is actually valid :(

unho:
@dwaynebailey I mean that we don't want to mark as placeable strings with stuff that are not format placeholders. For example "%s" is a placeholder, but "% s" is not.

dwaynebailey:
This placeholder is implementing printf as in `man 3 printf`, please give reasons for departing from that if you must.  I'd be happy if I'd actually get valid arguments from people.  And this fix essentially makes it work by making it not work for the space example.

The docs state:
 `' '` (space)  A blank should be left before a positive number produced by a signed conversion (a, A, d, e, E, f, F, g, G, or i).

I agree its a corner case.  But until I actually hear arguments about why and documents in the change I convinced that nobody has even bothered to look at the comments in the function or the manual page.

PR #3415, opened 2016-03-30
Cleanup docstring to more closely conform to PEP257

dwaynebailey:
* Does PEP257 cleanups
* Sets initial pep257 command configuration
* Turn `__all__` from list to tuple (may be controversial but not technically wrong) to aid pep257

Unfortunately we can't turn this on by default because of https://github.com/PyCQA/pydocstyle/issues/182

### DIFF
--- a/translate/storage/xml_extract/generate.py
+++ b/translate/storage/xml_extract/generate.py
@@ -232,9 +232,9 @@ def replace_dom_text(make_parse_state,
 
           action: etree_Element x base.TranslationUnit -> None
 
-      which takes a dom_node and a translation unit. The dom_node is rearranged
-      according to rearrangement of placeables in unit.target (relative to their
-      positions in unit.source).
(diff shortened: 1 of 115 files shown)


translate/storage/xml_extract/generate.py
```diff
@@ -106,7 +106,7 @@ def set_dom_root_for_unit_node(parent_unit_node, unit_node, dom_tree_roots):
 
 
 def _map_source_dom_to_doc_dom(unit_node, source_dom_node):
-    """Creating a mapping from the DOM nodes in source_dom_node which correspond to
+    r"""Creating a mapping from the DOM nodes in source_dom_node which correspond to
```
unho: Why prefixing it with `r`?

unho:
Other than the nit, lgtm.

translate/storage/xml_extract/generate.py
dwaynebailey: `\ `

PR #3417, opened 2016-04-01
Skip dtd real newlines

unho:


translate/convert/test_po2dtd.py
```diff
@@ -435,7 +435,7 @@ def test_roundtrip_amp(self):
     def test_merging_entries_with_spaces_removed(self):
         """dtd2po removes pretty printed spaces, this tests that we can merge this back into the pretty printed dtd"""
         posource = '''#: simple.label\nmsgid "First line then "\n"next lines."\nmsgstr "Eerste lyne en dan volgende lyne."\n'''
-        dtdtemplate = ('<!ENTITY simple.label "First line then\n'
+        dtdtemplate = ('<!ENTITY simple.label "First line then\\n'
```
dwaynebailey: This doesn't look right.  Why are we escaping this newline?

dwaynebailey: Interestingly this test was never right in that the PO file would be wrong.

But this fix is also wrong.  The \n in the DTD is not \\n.  The PO file itself would be wrong based on our new output.

translate/storage/dtd.py
```diff
@@ -105,6 +105,9 @@
 ending in :attr:`.labelsuffixes` into accelerator notation"""
 
 
+#TODO convert this to Dialect like for properties.
```
dwaynebailey: `# TODO` please

unho: Forgot about this one. Removing and creating issue.

Issue #3396, opened 2016-03-11
Get rid of translate.misc.dictutils.ordereddict


unho:
`translate.misc.dictutils.ordereddict` defines a custom ordered dict class. Check if we can get rid of it and replace it with `collections.OrderedDict` (and `ordereddict.OrderedDict` for Python 2.6).

unho:
After #3401 this might be changed to:

- Replace usage usage of `translate.misc.dictutils.ordereddict` by `collections.OrderedDict`.
- Drop `translate.misc.dictutils.ordereddict`.

PR #3422, opened 2016-04-07
Drop Python 2.6 support

### LINKED ISSUE #3401: Drop Python 2.6 support in ttk
We should drop Python 2.6 support because of the maintenance burden.  The reasons and options available to users is [clearly outlined in this](http://www.snarky.ca/stop-using-python-2-6) and similar articles.

Summary would be:

1. There are alternate options for people to get at least Python 2.7 on older platforms like RHEL.
2. It's a massive burden to support Python 2.6 at no cost to the user but at great expense to the ttk devs. Especially considering point 1.

As a note to point 2. Landing the json2po changes to maintain ordering was a 10min fix.  Coping with making it run on Python 2.6 lead to new dependencies, try except blocks and fallbacks to non-optimal behaviour for Python 2.6.  And that took half a day to develop.

So my motivation, so that we can focus on localisation, is to drop a very very very old release which we're maintaining for a teeny tiny user base (if they exist at all) who already have a many many many options to get Python 2.7 which we do support.

unho:
Fixes #3401, #3396.


dwaynebailey:
@unho I don't see the actual removal of misc.orderdict in this PR.

unho:
@dwaynebailey It is dropped now.

dwaynebailey:
@unho lgtm

PR #3425, opened 2016-04-08
Grab bag of fixes

dwaynebailey:
This is a grab bag of fixes from mlo.  I've tried to commit them as distinct things without trying to make PRs for everything.  Would appreciate some review though.

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -353,7 +353,7 @@ class pounit(pocommon.pounit):
             commentlist = self.automaticcomments
             linestart = "#. "
         text = text.split("\n")
-        newcomments = [linestart + line + "\n" for line in text]
+        newcomments = [linestart + line[:-1] + line[-1:].replace("\\", "\\ ") + "\n" for line in text]
         if position == "append":
             newcomments = commentlist + newcomments
         elif position == "prepend":
--- a/translate/convert/po2dtd.py
+++ b/translate/convert/po2dtd.py
@@ -140,7 +140,8 @@ class po2dtd(object):
             unquoted = None
         else:
             unquoted = inputunit.source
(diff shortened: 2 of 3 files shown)


translate/convert/accesskey.py
```diff
@@ -135,8 +135,8 @@ def extract(string, accesskey_marker=DEFAULT_ACCESSKEY_MARKER):
             if (accesskey_marker == '&' and
                 XMLEntityPlaceable.regex.match(string[marker_pos-1:])):
                 continue
-            label = string[:marker_pos-1] + string[marker_pos:]
-            if string[marker_pos] != " ":  # FIXME weak filtering
+            if string[marker_pos] != " ":  # FIXME week filtering
```
unho: week filtering? What does this mean?

unho:
Some of the changes in po2dtd seem to require further tests tweaks for Travis to pass.

translate/convert/accesskey.py
dwaynebailey: Mmm that doesn't make sense at all :/

unho:
Perhaps it is a good idea to try to break this into smaller chunks so some stuff can be landed.

translate/storage/pypo.py
```diff
@@ -353,7 +353,7 @@ def addnote(self, text, origin=None, position="append"):
             commentlist = self.automaticcomments
             linestart = "#. "
         text = text.split("\n")
-        newcomments = [linestart + line + "\n" for line in text]
+        newcomments = [linestart + line[:-1] + line[-1:].replace("\\", "\\ ") + "\n" for line in text]
```
unho: Some test would be good to have.

dwaynebailey:
We've landed all but the `\` at end of line in PO files into `master`. So closing this

PR #3429, opened 2016-04-19
Re-added str method for multistring type

### LINKED ISSUE #3428: Multistring doesnt work as a python2 string
Previously you were able to so ~~`str(multistring("foo"))`~~ but it seems this no longer works.

EDIT: apologies that was a bad example it should have read something like `str(multistring("é"))`

Im wondering if we should add py2 compat similar to the one in django (https://docs.djangoproject.com/en/1.9/_modules/django/utils/six/#python_2_unicode_compatible)

@claudep any thoughts on this?

claudep:
Fixes #3428.


unho:
looks ok.

PR #3430, opened 2016-04-22
multistring: restore decoding bytes

julen:
We've hit an issue in Pootle which can only be reproduced with TTK master (https://github.com/translate/pootle/issues/4594). The real cause of this is unclear (@phlax mentioned this is originating somewhere in `Store.update()`), and it seems like coercing `multistring`s with non-ASCII data to unicode fails.

In that sense, I'm not sure whether `multistring` should accept bytes in its constructor or not, however previously this [used to work via the `autoencode` class](https://github.com/translate/pootle/blob/02c993c59fadde9f495934ce0a34b87d523ad5f1/pootle/apps/pootle_store/fields.py#L59-L79) (note the `UTF-8` argument to the constructor) which is now removed in master.

I have added a test and a patch for this, which when creating the object would decode bytes from UTF-8 if that was passed. This seems to align with the previous behavior, not sure if we should aim for that though...

@claudep we'd appreciate to hear your thoughts on this one, TIA :)

### DIFF
--- a/translate/misc/multistring.py
+++ b/translate/misc/multistring.py
@@ -21,19 +21,39 @@
 strings in the strings attribute
 """
 
+import warnings
+
 import six
+
+from .deprecation import RemovedInTTK2Warning
+
+
+def _create_text_type(newtype, string, encoding):
+    """Helper to construct a text type out of characters or bytes. Required to
+    temporarily preserve backwards compatibility. Must be removed in TTK2.
+    """
+    if isinstance(string, six.text_type):
+        return six.text_type.__new__(newtype, string)
+
+    warnings.warn(
+        'Passing non-ASCII bytes as well as the `encoding` argument to '
+        '`multistring` is deprecated. Always pass unicode characters instead.',
+        RemovedInTTK2Warning, stacklevel=2,
+    )
+    return six.text_type.__new__(newtype, string or six.binary_type(), encoding)
 
 
 class multistring(six.text_type):
 
     def __new__(newtype, string=u"", *args, **kwargs):
+        encoding = kwargs.pop('encoding', 'utf-8')
         if isinstance(string, list):
             if not string:
                 raise ValueError("multistring must contain at least one string")
-            newstring = multistring.__new__(newtype, string[0])
+            newstring = _create_text_type(newtype, string[0], encoding)
             newstring.strings = [newstring] + [multistring.__new__(newtype, altstring) for altstring in string[1:]]
         else:
-            newstring = six.text_type.__new__(newtype, string)
+            newstring = _create_text_type(newtype, string, encoding)
             newstring.strings = [newstring]
         return newstring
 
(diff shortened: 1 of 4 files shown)


claudep:
My opinion is that if non-ascii non-unicode content is reaching the multistring constructor, there is a bug somewhere in the stack. You could choose to be permissive and elude the issue by this patch, or you could be stricter and decide that non-ascii bytestrings are forbidden at this stage. Up to you.

phlax:
@claudep i think we should fix the issue in Pootle, but the issue of not having backwards compatibility means that old versions of Pootle cant use the newer ttk.

Im also concerned that other software that uses ttk may also have similar issues.

Im not clear the best way forward really.

claudep:
I understand. I won't blame you if you choose the permissive solution.

phlax:
Im thinking perhaps we land the patch for backwards compatibility, but add a deprecation warning, and deprecate ~soon after

translate/misc/multistring.py
```diff
@@ -24,16 +24,22 @@
 import six
 
 
+def _create_text_type(newtype, string):
+    if isinstance(string, unicode):
```
claudep: Also use `six.text_type` here instead of `unicode`, same in tests.

unho:
> Im thinking perhaps we land the patch for backwards compatibility, but add a deprecation warning, and deprecate ~soon after

Sounds like a sensible way to don't break without notice, but looking forward. We can't obviously prevent any existing Pootle or third party to avoid using newer TTK once we drop the deprecation warning in the future to embrace the stricter way. But don't let that stopping us from moving forward.

translate/misc/multistring.py
julen: Thank you 👍, living in Py2 world this was easy to miss.

phlax:
lgtm, nice one!

julen:
Thanks for the comments, folks. @claudep your comment reinforces my thinking, although as mentioned here we are abruptly breaking backwards compatibility and creating some trouble, hence the middle-ground solution proposed by @phlax is probably the best bet. I have updated the PR to take this into account.

As an orthogonal thought, I'm a bit concerned there's no clear path to API deprecation, and we should probably formalize how deprecation/removal is handled in TTK; e.g. unless it becomes absolutely necessary to break backwards compatibility, generally speaking deprecate in version N, remove in version N+1. I believe this would align with common practice out there.

unho:
lgtm

julen:
@claudep any objections to have this PR merged?

julen:
In line with my previous comment, I'd also be interested to hear more on how are we willing to handle deprecation in the future. Should I bring that up on the mailing list? Or do people prefer to discuss these things in issues these days?

claudep:
I don't want to take position on the deprecation strategy, but apart from that, I support the change.

PR #3431, opened 2016-04-25
multistring: fix list coercion to text

julen:
At the moment code such as `'%s' % [multistring(u'føø')]` fails: the
representation of `multistring` is used when coercing the list to text, however
there are decoding errors.

This commit adjusts the representation of `multistring`s so that a) coercing
lists containing `multistring`s is fixed, and b) the representation allows to
re-create the original object.

On b), this is what the official docs on `repr()` suggest at
https://docs.python.org/2/library/functions.html#repr:
>  For many types, this function makes an attempt to return a string that would
>  yield an object with the same value when passed to eval(), otherwise the
>  representation is a string enclosed in angle brackets that contains the name
>  of the type of the object together with additional information often
>  including the name and address of the object.

### DIFF
--- a/translate/misc/multistring.py
+++ b/translate/misc/multistring.py
@@ -91,7 +91,9 @@ class multistring(six.text_type):
         return self.__cmp__(otherstring) == 0
 
     def __repr__(self):
-        _repr = "multistring([" + u",".join(self.strings) + "])"
+        _repr = u"multistring(%r)" % (
+            [six.text_type(item) for item in self.strings]
+        )
         return _repr.encode('utf-8') if six.PY2 else _repr
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


julen:
Does anyone know of a solution to have the `repr()` test pass on Python 3? I can conditionally test for alternate representations based on the Python version, however I'm figuring there might be a more elegant solution.

claudep:
I'm not aware of a more elegant solution. You can define a str_prefix as `'' if six.PY3 else 'u'`, then include that str_prefix in tested strings.

julen:
Thanks for the quick feedback @claudep; I've updated the test accordingly.

julen:
Oh, the representation itself is different too 😎.

unho:
lgtm

PR #3433, opened 2016-04-25
Add setcontext to mo file

nijel:
It works without problem, just the method is missing.

### DIFF
--- a/translate/storage/mo.py
+++ b/translate/storage/mo.py
@@ -118,6 +118,9 @@ class mounit(base.TranslationUnit):
             return None
         return "".join(self.msgctxt)
 
+    def setcontext(self, context):
+        self.msgctxt = [context]
+
     def isheader(self):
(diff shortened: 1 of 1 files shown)


unho:
The method is unnecessary. The inherited implementation is just fine. Won't land, but thanks for your contribution. Closing this.

nijel:
Have you tested it? Why do you think that following code (which is inherited) will work? It doesn't store the context at all...

```
    def setcontext(self, context):
        """Set the message context"""
        pass
```

unho:
No, it doesn't store the context, but I don't think it is necessary to.

nijel:
So it's okay to have setcontext which does nothing and fiddle with msgctx directly?

nijel:
In such case I'd rather suggest to remove setcontext from the base class if it's not intended as generic API.

Anyway I've updated my branch to be based on current master and added testcase for this.

PR #3435, opened 2016-04-26
Multistring hash (for plurals)

phlax:
Previously you were able to do `str in dict_with_plural_multistring_key` and because the str and multistring hashes matched it would find the dict element.

This PR adds a test to demonstrate how ttk behaviour version has changed, and a fix to restore old beviour


unho:
lgtm from what I can tell.

PR #3436, opened 2016-04-30
fix LibreOffice validxml check re self-closing tags

cloph:
self-closing tags were not treated properly, the regular expression did
extend the match over the closing /> until the next tags, hence breaking
the check/causing false positives.

Fix the regex and also further extend the checks so it is able to decide
whether empty tag is self-closed as it should be, or wehther non-empty
ones are errornously self-closing.

add quotes around the error message to make identifying the match
easier.

changes to the tag-regex:
[/]?? → /?
no reason for character class if it is just a single character, also
non-greedy modifier doesn't make sense for 0 or 1 match of single character

[a-z_\-]+? → [a-z_-]+
- at the end of list doesn't need to be escaped, and as we limit the
match to valid characters only, no need for lazy modifier, we want to
match all characters from the class, also encapsulate in a named group

(?:| +[a-z]+?=".*?") → (?: +[a-z]+="[^"]+")*
there can be more than one attribute, so add * multiplier (and that way
also get dir of the empty alterative in the group) once again remove
non-greedy modifiers, as we want to match all from the characterclass,
replace the dot accordingly

> → /?>
the important bugfix, to also account for self-closing tags.


unho:
Landed in b2c13630. Thanks a lot for your contribution!

Did some tests to see the performance impact. Roughly this improvement only imposes a less than 1% penalty in performance:

| Before        | After         |
| ------------- | ------------- |
| 4,2578005791  | 4,3593928814  |
| 4,2627990246  | 4,3861579895  |
| 4,4204928875  | 4,2902569771  |
| 4,2738559246  | 4,3098728657  |
| 4,3039479256  | 4,3436949253  |

So the average time before is 4,3037792683 and with the changes is 4,3378751278, resulting in an increase of 0,00792% in the time used. With larger files the proposed changes even reduce the time compared to previous state.

PR #3437, opened 2016-05-04
2.0.0 beta release

unho:


docs/releases/2.0.0-beta1.rst
```diff
@@ -1,31 +1,34 @@
 Translate Toolkit 1.14.0
```
dwaynebailey: Version needs changing

unho: Changing it.

docs/releases/2.0.0-beta1.rst
```diff
@@ -1,31 +1,34 @@
...
 
-
-1.14.0 vs 1.14.0-rc1
-====================
+2.0.0-beta1 vs 1.14.0-rc1
+=========================
 
 - Dropped support for Python 2.6.
 - Dropped ``translate.misc.dictutils.ordereddict``.
+- Added support for Python 3.5.
```
dwaynebailey: Expaned Python 3 support to include Python 3.5

docs/releases/2.0.0-beta1.rst
```diff
@@ -46,7 +49,7 @@ Python 3 support
 - Translate Toolkit went through a massive code cleanup looking forward Python
   3 compatibility. There might be quirks that need to be fixed, so please test
   and `report any issue <https://github.com/translate/translate/issues/new>`_
-  you might find.
+  you might find. Python 3.5 is now supported.
```
dwaynebailey: That addition doesn't make any sense.

Just add a bullet pint: Python 3.3-3.5 is now supported

dwaynebailey: Can we specifically credit Claude for the Python 3 support within the major changes listed above.

unho:
Addressed the raised concerns.

claudep:
Thanks for the credits :-)

translate/__version__.py
```diff
@@ -19,16 +19,16 @@
...
 code touching stats or quality checks.  An increased build number will force a
 toolkit user, like Pootle, to regenerate it's stored stats and check
 results."""
 
-sver = "1.14.0-rc1"
+sver = "2.0.0-beta1"
 """Human readable version number. Used for version number display."""
 
-ver = (1, 14, 0)
+ver = (2, 0, 0)
```
dwaynebailey: This does remind me that we really should duplicate the versioning stuff from Pootle into ttk.  Just not now.

translate/__version__.py
```diff
@@ -19,16 +19,16 @@
...
-build = 14001
+build = 20000
 """The build number is used by external users of the Translate Toolkit to
 trigger refreshes.  Thus increase the build number whenever changes are made to
 code touching stats or quality checks.  An increased build number will force a
 toolkit user, like Pootle, to regenerate it's stored stats and check
 results."""
 
-sver = "1.14.0-rc1"
+sver = "2.0.0-beta1"
```
dwaynebailey: i wonder if we should rather follow semantic versioning where this would be 2.0.0b1 IIRC

unho: Will change tomorrow.

translate/__version__.py
unho: Sure.

docs/releases/2.0.0-beta1.rst
```diff
@@ -47,6 +50,7 @@ Python 3 support
   3 compatibility. There might be quirks that need to be fixed, so please test
   and `report any issue <https://github.com/translate/translate/issues/new>`_
   you might find.
+- Python 3.3-3.5 is now supported. Thanks to Claude Paroz.
```
dwaynebailey: Drop the Claude bit here, above was enough

unho:
Updated. Will land and release this morning.

unho:
It seems @ta2-1 found some issue on Pootle that might stem to TTK, so holding release until he works out the issue.

unho:
The issue seems to be Pootle only, so releasing.

PR #3442, opened 2016-05-20
Simplify newlines handling

### LINKED ISSUE #3262: Parse EOL as newline on Android strings
Contrary to what @nijel implemented on [Android resources support #2](https://github.com/translate/translate/pull/2#issuecomment-8185814) it is (now?) possible to [use a real EOL instead of an explicit `\n`](http://stackoverflow.com/a/17288710) on Android strings' files. This makes the strings' files easier to read.

Thus translate-toolkit should parse an EOL as a newline instead of a whitespace.

nijel:
The format now handles \n and newline equally, so we can remove specific
handling for newlines.

Fixes #3262

### DIFF
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -34,8 +34,8 @@ from translate.misc.multistring import multistring
 from babel.core import Locale, UnknownLocaleError
 
 EOF = None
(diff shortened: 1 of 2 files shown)


nijel:
Rebased on current master....

dwaynebailey:
@nijel thanks again

PR #3443, opened 2016-05-21
[ts] Reduce format differences with files saved from Qt Linguist

bjorn:
Normally our .ts files are edited in Qt Linguist. When they are however saved by for example Weblate, the files contain many formatting differences that make it hard to see what actual edits have been made to the translations (basically you first have to open and re-save the file in Qt Linguist, and then make a diff again).

With this change, the difference is much smaller, but improvements could still be made regarding how ' and " are exported for example (Qt Linguist writes `&apos;` and `&quot;`), as well as the header format:
```diff
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version='1.0' encoding='utf-8'?>
```

### DIFF
--- a/docs/formats/ts.rst
+++ b/docs/formats/ts.rst
@@ -16,9 +16,9 @@ References
 The format is XML and seems to only have been documented properly since Qt 4.3
 
 * `Current DTD Specification
-  <http://qt-project.org/doc/qt-5.0/qtlinguist/linguist-ts-file-format.html>`_ for Wt 5.0,
+  <http://doc.qt.io/qt-5/linguist-ts-file-format.html>`_ for Qt 5,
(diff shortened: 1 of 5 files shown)


dwaynebailey:
@bjorn thanks for the PR.

Would you be able to:

- Add tests in `test_ts2.py` that test this change
- Fix the doc failure

bjorn:
Alright, I pushed a small fixup that should take care of the doc failure (one ` too many). I'll look into adding a test soon.

bjorn:
@dwaynebailey, @dg0yt helped with the test and to correct further small differences.

bjorn:
Closing this PR since it was superceded by #3465.

PR #3455, opened 2016-06-22
Fix parsing of Gettext po files with first entry in unicode

### LINKED ISSUE #3450: poparser wrongly parses first unit source
When msgid of first unit of po file is unicode, it's wrongly parsed (it's decoded using iso-8859-1 instead of charset specified in the header).

File which shows this behavior: [Author_Group.po.txt](https://github.com/translate/translate/files/322881/Author_Group.po.txt)

Testcase patch to see the problem: [test_po.patch.txt](https://github.com/translate/translate/files/322885/test_po.patch.txt)

nijel:
The headers is first parsed with iso-8859-1 and the parser always
parses next line using this encoding. This needs to be parsed again once
we know proper encoding of the file.

This fixes #3450

### DIFF
--- a/translate/storage/poparser.py
+++ b/translate/storage/poparser.py
@@ -356,6 +356,12 @@ def parse_header(parse_state, store):
         return None
     set_encoding(parse_state, store, first_unit)
     decode_header(first_unit, parse_state.decode)
+    # Fix encoding of next line in parser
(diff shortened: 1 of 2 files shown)


dwaynebailey:
Landed as 2f4e298d842d7f902fd8d4d3864a4c237c50be40

Thanks @nijel nicely spotted and thanks for the tests.

PR #3465, opened 2016-07-23
Reduce format differences with files saved from Qt Linguist

dg0yt:
Writes the XML declaration with double quotes.
Writes an empty location element.

### DIFF
--- a/docs/formats/ts.rst
+++ b/docs/formats/ts.rst
@@ -16,9 +16,9 @@ References
 The format is XML and seems to only have been documented properly since Qt 4.3
 
 * `Current DTD Specification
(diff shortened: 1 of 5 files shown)


dg0yt:
This is a new pull request of previous work in #3443. It turned out that I need to use the integration checks...

The difference which is left open is the handling of '/" vs. &apos;/&quot;.

dg0yt:
Shall I touch the quotes sub-issue, or leave this PR as is? I would use regexps applied on the string from the etree.

dwaynebailey:
@dg0yt feel free to touch, but it might make it easier to PR elsewhere as we're prioritising reviewing your great work here to land.

dg0yt:
I guess I can't easily create another PR because this one might get updated as soon as I commit to my fork - and the new change must depend on those already collected.

unho:
Landed in master. Thanks for the contributions.

PR #3467, opened 2016-07-24
Fix parsing of Gettext locations with percent char

### LINKED ISSUE #3466: Unicode error on % char in PO location
When location in PO file contains location, you end up with unicode error when parsing this:

UnicodeDecodeError: 'utf8' codec can't decode byte 0xab in position 20: invalid start byte

Sample to reproduce:

```po
#: /foo/bar/%%var%%www%%about.html:44
msgid "test"
msgstr "rest"
```

I will create PR to fix this.

nijel:
Fixes #3466

Signed-off-by: Michal Čihař <michal@cihar.com>

### DIFF
--- a/translate/storage/test_po.py
+++ b/translate/storage/test_po.py
@@ -322,6 +322,17 @@ msgstr[1] "Kóeie"
         assert locations[0] == u"programming/C/programming.xml:44(para)"
         assert isinstance(locations[0], six.text_type)
 
+    def test_percent_location(self):
+        """test that we correctly handle a location with percent chars"""
+        posource = '#: /foo/bar/%%var%%www%%about.html:44\nmsgid "test"\nmsgstr "rest"\n'
+        pofile = self.poparse(posource)
+        u = pofile.units[-1]
+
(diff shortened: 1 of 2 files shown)


translate/storage/test_po.py
```diff
@@ -322,6 +322,17 @@ def test_nongettext_location(self):
         assert locations[0] == u"programming/C/programming.xml:44(para)"
         assert isinstance(locations[0], six.text_type)
 
+    def test_percent_location(self):
+        """test that we correctly handle a location with percent chars"""
+        posource = '#: /foo/bar/%%var%%www%%about.html:44\nmsgid "test"\nmsgstr "rest"\n'
```
unho: Next times please use a multiline string since it makes easier to copy paste actual examples without need to escape chars. Also it is easier to read by humans.

unho:
Landed in 4d94eb66. Thanks for your contribution!

translate/storage/test_po.py
nijel: Will do, I've just followed style used in the test above...

PR #3468, opened 2016-07-26
Move all namespaces to <resources> tag

nijel:
All standard namespaces need to be defined on <resources> tag rather
than on lower elements in Android string resources.

### DIFF
--- a/min-required.txt
+++ b/min-required.txt
@@ -1,6 +1,6 @@
-lxml==2.1.0  # >=1.3.4 should work
+lxml==3.5.0
 python-Levenshtein==0.10.2
-# If Python 
+# If Python
 iniparse==0.3.1
 vobject==0.6.6
 BeautifulSoup==3.2
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -508,5 +508,5 @@ class AndroidResourceFile(lisa.LISAfile):
         super(AndroidResourceFile, self).addunit(unit, new)
         # Move aliased namespaces to the <resources> tag
         # The top_nsmap was introduced in LXML 3.5.0
-        if do_cleanup and etree.LXML_VERSION >= (3, 5, 0):
+        if do_cleanup:
             etree.cleanup_namespaces(self.body, top_nsmap=newns)
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -489,3 +489,24 @@ class AndroidResourceFile(lisa.LISAfile):
             self.settargetlanguage(target_lang)
 
         return target_lang
+
+    def addunit(self, unit, new=True):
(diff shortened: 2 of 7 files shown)


translate/storage/aresource.py
```diff
@@ -489,3 +489,24 @@ def gettargetlanguage(self):
...
+            newns = self.body.nsmap
+            for ns in unit.xmlelement.nsmap:
+                if ns not in newns:
+                    do_cleanup = True
+                    newns[ns] = unit.xmlelement.nsmap[ns]
+
+        super(AndroidResourceFile, self).addunit(unit, new)
+        # Move aliased namespaces to the <resources> tag
+        # The top_nsmap was introduced in LXML 3.5.0
+        if do_cleanup and etree.LXML_VERSION >= (3, 5, 0):
```
unho: Please bump the lxml requirements to match this version. If possible it is preferable to have a consistent behavior independently of the installed dependencies.

nijel: Okay, I've raised it in separate commit.

min-required.txt
```diff
@@ -1,6 +1,6 @@
-lxml==2.1.0  # >=1.3.4 should work
+lxml==3.5.0
```
dwaynebailey: @unho @nijel I'm all for bumping this to prevent the `if<3.5` logic.  But I'd like to know more ito how this impacts users.  Are we actually excluding some platforms now?

phlax: @dwaynebailey see the `requirements` section here http://lxml.de/installation.html

we need to check those against debian wheezy,wheezy-backports,jessie and some flavours of ubuntu i reckon

nijel: 3.5.0 is neither in jessie nor in wheezy, 3.6.0 is in sid/strech.

phlax: we used debian as reference - because it gives a good indication of support in other distros - particularly ubuntu.

If the deps have to be compiled to install the py libs, i would say that was a major blocker

phlax: @nijel just to clarify - its the libxml dependencies that im concerned about - not so much the lxml version

nijel: Okay, then you're fine - everything since wheezy (oldstable) has new enough libxml and libxslt.

unho: Ubuntu 12.04 has accepted versions for the required lxml dependencies, and 14.04 comes with the recommended minimum, and of course newer releases do too. Jessie comes with acceptable libxml2 2.9.1 where the recommended minimum is 2.9.2 and bundles recommended minimum libxslt. I would say that we are safe, so I am landing.

phlax: thanks @unho for taking the time to follow this up

unho:
Landed to master. Thanks for your contribution.

PR #3470, opened 2016-07-26
ts: Write quotes as entities

dg0yt:
Removes final format difference with files saved from Qt Linguist.

### DIFF
--- a/translate/storage/ts2.py
+++ b/translate/storage/ts2.py
@@ -488,6 +488,19 @@ class tsfile(lisa.LISAfile):
         # Qt Linguist does self-close the "location" elements though
         for e in root.xpath("//*[not(./node()) and not(text()) and not(name() = 'location')]"):
             e.text = ""
+        # For conformance with Qt output, write XML declaration with double quotes
         out.write(str('<?xml version="1.0" encoding="utf-8"?>\n').encode('utf-8'))
-        out.write(etree.tostring(root, doctype=doctype, pretty_print=True,
-                                 xml_declaration=None, encoding='utf-8'))
+        # For conformance with Qt output, post-process etree.tostring output,
+        # replacing ' with &apos; and " with &quot; in text elements
+        treestring = etree.tostring(root, doctype=doctype, pretty_print=True,
+                                    xml_declaration=False, encoding='unicode')
+        pos = 0
+        while pos >= 0:
+            nextpos = treestring.find('<', pos)
+            value = treestring[pos:nextpos].replace("'", "&apos;").replace('"', "&quot;")
+            out.write(value.encode('utf-8'))
+            pos = nextpos
+            nextpos = treestring.find('>', pos)
+            out.write(treestring[pos:nextpos].encode('utf-8'))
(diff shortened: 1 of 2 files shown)


dg0yt:
This might deserve some more testing. But I only use hosted weblate...
Unfortunately, this slows down serialize significantly (tested with 1000 iterations in test_backnforth).

dg0yt:
Now completely without regexps. The slowdow is still noticable, but no longer by factor 10...

translate/storage/ts2.py
```diff
@@ -488,6 +489,20 @@ def serialize(self, out):
...
             e.text = ""
+        # For conformance with Qt output, write XML declaration with double quotes
         out.write(str('<?xml version="1.0" encoding="utf-8"?>\n').encode('utf-8'))
-        out.write(etree.tostring(root, doctype=doctype, pretty_print=True,
-                                 xml_declaration=None, encoding='utf-8'))
+        # For conformance with Qt output, post-process etree.tostring output,
+        # replacing ' with &apos; and " with &quot; in text elements
+        treestring = etree.tostring(root, doctype=doctype, pretty_print=True,
+                                    xml_declaration=False, encoding='utf-8')
+        treestring = codecs.decode(treestring, 'utf-8')
```
nijel: You should use encoding='unicode' for etree.tostring(), that saves need for the additional utf-8 decoding.

dg0yt: Done.

dg0yt:
Ping. I would like to remove nasty git noise workarounds from our project's build system.

unho:
As you said this brings a noticeable performance drop. Can you explore using regexes to speed it up?

dg0yt:
Hm. I tried regexes first and it was slower by one order of magnitude.

translate/storage/ts2.py
```diff
@@ -488,6 +488,19 @@ def serialize(self, out):
...
         for e in root.xpath("//*[not(./node()) and not(text()) and not(name() = 'location')]"):
             e.text = ""
+        # For conformance with Qt output, write XML declaration with double quotes
         out.write(str('<?xml version="1.0" encoding="utf-8"?>\n').encode('utf-8'))
-        out.write(etree.tostring(root, doctype=doctype, pretty_print=True,
-                                 xml_declaration=None, encoding='utf-8'))
+        # For conformance with Qt output, post-process etree.tostring output,
+        # replacing ' with &apos; and " with &quot; in text elements
+        treestring = etree.tostring(root, doctype=doctype, pretty_print=True,
+                                    xml_declaration=False, encoding='unicode')
```
dg0yt: Maybe one could do `encode('utf-8')` the whole string here, rather calling `encode` many times in the loop. But I'm not a python expert. I have hardly any knowledge about dealing with encoding of python strings.

nijel: Indeed it can be done simpler, see https://github.com/translate/translate/pull/3528

unho:
Closing in favor of https://github.com/translate/translate/pull/3528

PR #3472, opened 2016-07-28
Fix for issue 3463

unho:


dwaynebailey:
@unho just to flag that becuase .strings are handled here using .properties we really need to make sure that these changes, which seems quite invasive, aren't impacting our pretty stable .properties support.

unho:
> @unho just to flag that becuase .strings are handled here using .properties we really need to make sure that these changes, which seems quite invasive, aren't impacting our pretty stable .properties support.

The skip first entry commit doesn't not introduce any change in the output regarding current master. The other commit (Do not process blanklines) does indeed remove all blank lines from output.

We can safely drop the later at the cost that users dealing with `.strings` will have to remove the blanklines in their files before using `po2prop`.

PR #3475, opened 2016-08-01
po2prop: remove full untranslated units if required

dwaynebailey:
Previously partial lines remained

### DIFF
--- a/translate/convert/test_po2prop.py
+++ b/translate/convert/test_po2prop.py
@@ -236,10 +236,28 @@ msgstr "translated"
     def test_merging_untranslated_multiline(self):
         """check merging untranslated entries with multiline values"""
         posource = '''#: prop\nmsgid "value1 value2"\nmsgstr ""\n'''
-        proptemplate = '''prop = value1 \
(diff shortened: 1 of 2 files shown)


translate/convert/test_po2prop.py
```diff
@@ -247,6 +247,20 @@ def test_merging_untranslated_multiline(self):
         print(propfile)
         assert propfile == ''  # We drop the key
 
+    def test_merging_untranslated_multiline2(self):
+        """check merging untranslated entries with multiline values"""
+        posource = '''#: legal_text_and_links3\nmsgid "By using {{clientShortname}} you agree to the {{terms_of_use}} and {{privacy_notice}}."\nmsgstr ""\n'''
```
unho: IMHO it is readable to do use of multiline strings and avoid having everything on a single line which uses `\n`. So:

```python
        posource = '''
#: legal_text_and_links3
msgid "By using {{clientShortname}} you agree to the {{terms_of_use}} and {{privacy_notice}}."
msgstr ""
'''
```

unho:
Nit aside, lgtm.

PR #3476, opened 2016-08-01
Skip some checks for some languages for some checker

unho:
This is necessary in order to disable accelerator checks for Indic
languages in Mozilla, and to disable checks that yield too many
false positives for some languages.


unho:
The changes do work, but it is still to be done disabling "accelerators" check for all the languages we want it to be disabled in Mozilla checker.

PR #3485, opened 2016-08-17
mozlang: last unit is followed by trailing newlines

dwaynebailey:
Based on files in the wild the final unit is, like all units, followed
by newlines.

### DIFF
--- a/docs/formats/index.rst
+++ b/docs/formats/index.rst
@@ -45,6 +45,7 @@ Other translation formats
    catkeys
    android
    resx
+   mozilla_lang
 
 * :doc:`csv`
 * :doc:`ini` (including Inno Setup .isl dialect)
@@ -60,6 +61,7 @@ Other translation formats
 * Haiku :doc:`catkeys` (from version 1.8)
 * :doc:`android` (supports storage, not conversion)
 * :doc:`resx` .NET Resource files (.resx)
+* Mozilla :doc:`.lang <mozilla_lang>` files
 
 .. _formats#translation_memory_formats:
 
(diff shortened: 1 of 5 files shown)


translate/storage/test_mozilla_lang.py
```diff
@@ -65,7 +65,7 @@ def test_active_flag(self):
         """Test the ## active ## flag"""
         lang = ("## active ##\n"
```
unho: In all examples I have checked the file metadata is separated by two empty lines (like the strings), so here there are two missing newlines. It is kinda unrelated to the PR, but it would be great to handle.

translate/storage/test_mozilla_lang.py
```diff
@@ -77,14 +77,14 @@ def test_multiline_comments(self):
...
         assert bytes(store).decode('utf-8') == lang
 
     def test_template(self):
         """A template should have source == target, though it could be blank"""
         lang = (";Source\n"
-                "Source\n")
+                "Source\n\n\n")
         store = self.StoreClass.parsestring(lang)
         unit = store.units[0]
         assert unit.source == "Source"
```
unho: In this very same test there is a lang source that doesn't comply with the format specs since units are not separated by two blank lines.

unho: Also source for next test doesn't include the two trailing blank lines after the string.

docs/formats/index.rst
```diff
@@ -60,6 +61,7 @@ Other translation formats
 * Haiku :doc:`catkeys` (from version 1.8)
 * :doc:`android` (supports storage, not conversion)
 * :doc:`resx` .NET Resource files (.resx)
+* Mozilla :doc:`mozilla_lang <.lang>` files
```
unho: This is triggering some error on docs build.

translate/storage/test_mozilla_lang.py
dwaynebailey: Lets deal with that elsewhere please

dwaynebailey: Not true for `## active ##`, seems to be true for other meta data though

unho:
gtm

PR #3486, opened 2016-08-23
Mozilla noaccelerators check

unho:
Some languages, mainly Indic, do not use accelerators, so instead of checking that the accelerator is present it checks for its absence.

By altering the accelerators check behavior there is no need to disable it anymore for some languages. Anyway the changes introduced to be able to disable some check for some language in a specific checker are kept since it is possible it is needed in the future


translate/filters/checks.py
```diff
@@ -413,10 +417,19 @@ def run_test(self, test, unit):
         """
         return test(unit)
 
-    def get_ignored_filters(self):
+    def get_checker_name(self):
+        """Extract checker name, for example 'mozilla' from MozillaChecker."""
+        return str(self.__class__.__name__).lower()[:-len("checker")]
```
dwaynebailey: Shouldn't we actually just get each checker to define that name instead of trying to mangle its extraction

translate/filters/checks.py
```diff
@@ -309,6 +309,10 @@ def __init__(self, checkerconfig=None, excludefilters=None,
         else:
             self.setconfig(checkerconfig)
 
+        # Exclude extra tests from being treated as tests. This is used to
+        # include some of these extra tests based on language config.
```
dwaynebailey: Not 100% sure I understand what this part is trying to achieve.  If they're included based on language why not call it `lang_specific_tests` or similar? Or `extra_optional_checks` Or `extra_conditional_checks`

translate/filters/checks.py
dwaynebailey: OK I see you just factored that out

translate/filters/checks.py
```diff
@@ -2156,6 +2175,28 @@ def accelerators(self, str1, str2):
         """
         return super(MozillaChecker, self).accelerators(str1, str2)
 
+    @cosmetic
+    def noaccelerators(self, str1, str2):
+        """Checks whether accelerators are absent in the target string.
+
+        Even if they are present in source string.
```
dwaynebailey: YOu probably need more descriptions here

translate/filters/test_checks.py
```diff
@@ -1374,3 +1374,124 @@ def test_skip_checks_per_language_in_some_checkers():
...
+    str1, str2, __ = strprep(u"&Check for updates", u"আপডেইটসমূহৰ বাবে নিৰীক্ষণ কৰক")
+    unit = base.TranslationUnit(str1)
+    unit.target = str2
+
+    failures = asmozillachecker.run_filters(unit)
+    assert 'noaccelerators' not in failures.keys()
+    failures = glmozillachecker.run_filters(unit)
+    assert 'noaccelerators' not in failures.keys()
+    failures = stdchecker.run_filters(unit)
+    assert 'noaccelerators' not in failures.keys()
```
dwaynebailey: just avoid the temporary `failures` variable it serves no purpose and makes it slighlty harder to read

translate/filters/test_checks.py
```diff
@@ -1374,3 +1374,124 @@ def test_skip_checks_per_language_in_some_checkers():
...
+
+    # No-accelerators check passes.
+    # No-accelerators check not in StandardChecker so it passes.
+    str1, str2, __ = strprep(u"&Check for updates", u"আপডেইটসমূহৰ বাবে নিৰীক্ষণ কৰক")
+    unit = base.TranslationUnit(str1)
+    unit.target = str2
+
+    failures = asmozillachecker.run_filters(unit)
+    assert 'noaccelerators' not in failures.keys()
+    failures = glmozillachecker.run_filters(unit)
```
dwaynebailey: But... the above two aren't using the standard checker?

dwaynebailey:
Some general comments.

* I really don't think that this kind of `noxxxx` check is the correct approach ever.  You'll see an old escaped one that made the same mistake.  We check for accelerators and have such a test already.  The only difference is that, based on the language and the style we either want to see accelerators or we don't.  So for me this should just have been an extension of the accelerators check, no a new check.
* I'm uncomfortable that we're still having to add each of these languages to a config.  It's the maintenance nightmare that I raised when we first blocked accelerators test.

I don't see anything technically wrong with the implementation.  I can't quite understand the checks they're not very clear.

I don't feel comfortable giving a signoff as we're just perpetuating the issues I've flagged before.

translate/filters/checks.py
unho: Yes, you have a point. But that feels like another PR.

translate/filters/checks.py
unho: I will rename it to `language_extra_tests`.

translate/filters/checks.py
unho: Will add a more detailed description.

translate/filters/test_checks.py
unho: Will remove it.

translate/filters/test_checks.py
unho: No, because here we are testing that `noaccelerators` in only in `MozillaChecker`, and only for some languages. So we test for:

- `MozillaChecker` with language that uses `noaccelerators`
- `MozillaChecker` with language that does not use `noaccelerators`
- `StandardChecker` which never uses `noaccelerators`

translate/filters/checks.py
unho: Done

translate/filters/checks.py
unho: Done

unho:
> * I really don't think that this kind of `noxxxx` check is the correct approach ever.  You'll see an old escaped one that made the same mistake.  We check for accelerators and have such a test already.  The only difference is that, based on the language and the style we either want to see accelerators or we don't.  So for me this should just have been an extension of the accelerators check, no a new check.

If this this is included as an extension of the accelerators check, then in Pootle it will communicate the wrong thing since we don't have the ability for a check to communicate different error messages. Adding a completely separate check will allow instead to provide specific error message, thus not misleading the users.

> * I'm uncomfortable that we're still having to add each of these languages to a config.  It's the maintenance nightmare that I raised when we first blocked accelerators test.

Since some checks are configured per language we have to put that config somewhere. If you have any suggestion on where this fits better I would be happy to do the required adjustments, provided your suggestion is ok.

dwaynebailey:
1. Sorry @unho but `pofilter` already does this, it reports the description of the failure which is raised in FilterFailure and SeriousFilterFailure.  So making the check broken because Pootle doesn't do it is a shortcut.  I'd rather see this fixed properly i.e. Pootle is able to expose the reason for the failure.  Did you look at that? Se https://github.com/translate/translate/blob/master/translate/filters/checks.py#L1138-L1150 as to why I have an issue with a check that should actually be one.  This never worked because there is no such special thing, its just a broken variable.

2. This is the same with the accelerators check switched off.  I raised it there and let it through for expediency.  These are being enabled/disabled mostly not because of the language, but because of the languages script.  That would be the correct approach.  That way if any new language arrives and it matches that script it can result in the correct behaviour.  What this change creates is a massive maintenance nightmare.

That said the ability to enable specific checks for a language is a good thing.  I just think this is the completely wrong approach here.

unho:
Discussed this with @dwaynebailey and we agreed on refactoring in this way:

- Drop current approach in this PR
- Add some language script list in a single file with a list of languages for each script (less maintenance nightmare)
  - This looks like more performant than detecting script in the text being filtered
- Alter existing `accelerators` filter, based on the script of the language, to either check for accelerators as previously, or to ensure they are absent (this is `noaccelerators`), effectively undoing most of #3476 
- Since Pootle doesn't currently have the ability to display specific error messages for each failed filter, we need to raise that for further discussion.

dwaynebailey:
@unho sounds good.  My only concern would be that e.g. in some projects we don't want accelerators.  But in others even when the language can't do it we want them.  E.g. expecting users to do something like "File (&f)".  So not sure how we'll handle that distinction.

unho:
> My only concern would be that e.g. in some projects we don't want accelerators. But in others even when the language can't do it we want them. E.g. expecting users to do something like "File (&f)". So not sure how we'll handle that distinction.

For projects we can pass a list of disabled checks when instantiating the check.

Regarding that second scenario it is not possible since when using the script to tell whether to run the filter or not we are loosing the granularity we had when setting which filter to skip for which checker for which language. Is this a real scenario?

dwaynebailey:
The scenario is real.  Note sure what GNOME does now.  But a language in moz that shouldn't have accelerators would address it like I highlighted when translating GNOME.

unho:
Discussed this with @dwaynebailey and the conclusion is that the second scenario can be handled if the alter-filter-according-to-script can be triggered only in some specific checkers, like MozillaChecker.

translate/filters/checks.py
```diff
@@ -2155,10 +2161,31 @@ def accelerators(self, str1, str2):
         """Checks whether accelerators are consistent between the
         two strings.
 
-        For Mozilla we lower the severity to cosmetic.
+        For Mozilla we lower the severity to cosmetic, and for some languages
```
dwaynebailey: With this change we could now bring back severe failure.  But lets do that outside of this PR

translate/filters/test_checks.py
```diff
@@ -1365,23 +1365,160 @@ def test_skip_checks_per_language_in_some_checkers():
...
+        checkerconfig=checks.CheckerConfig(accelmarkers="&",
+                                           targetlanguage="as")
+    )
+
     str1, str2, __ = strprep(u"&Check for updates", u"আপডেইটসমূহৰ বাবে নিৰীক্ষণ কৰক")
     unit = base.TranslationUnit(str1)
     unit.target = str2
 
+    # Accelerators check is disabled for Assamese in MozillaChecker.
+    assert 'accelerators' not in mozillachecker.run_filters(unit).keys()
```
dwaynebailey: mozillachecker == 'gl' as far as I can see. What is the mention of Assamse here?

translate/filters/test_checks.py
```diff
@@ -1365,23 +1365,160 @@ def test_skip_checks_per_language_in_some_checkers():
...
-    failures = mozillachecker.run_filters(unit)
-    # Accelerators check is disabled for Assamese in MozillaChecker.
-    assert 'accelerators' not in failures.keys()
+    assert fails(mozillachecker.accelerators, "&File", "&Fayile")
+    assert fails(mozillachecker.accelerators, "My add&-ons", "&Byvoengs mit")
+    assert passes(mozillachecker.accelerators, "&File", "Fayile")
+    assert fails(mozillachecker.accelerators, "File", "&Fayile")
+    assert passes(mozillachecker.accelerators, "Mail &amp; News", "Po en Nuus")
+    assert fails(mozillachecker.accelerators, "Mail &amp; News", "Po en &Nuus")
+    assert passes(mozillachecker.accelerators, "&File", "Fayile")
```
dwaynebailey: I suggest adding a sinple unlaced & as in "Mail & News" -> "Pos & Nuus", hopefully it works

translate/lang/data.py
```diff
@@ -239,6 +239,75 @@
 }
 
 
+scripts = {
+    'devanagari': [
```
dwaynebailey: Since we can make these be any name.  I'd suggest using [ISO 15924](http://unicode.org/iso15924/iso15924-codes.html) so that we can easily follow the standard.

translate/lang/data.py
```diff
@@ -239,6 +239,75 @@
...
+        'kfy',
+        'kok',
+        'mai',
+        'mr',
+        'sa',
+        'sat',
+    ],
+    'bengali': [
+        'bn_BD',
+        'bn_IN',
```
dwaynebailey: Since we simplify in script detection, can't we just specify `bn`.  The distinction in Mozillaland has more to do with politics than actual differences.

translate/lang/data.py
```diff
@@ -239,6 +239,75 @@
...
+        'gu',
+    ],
+    'khmer': [
+        'km',
+    ],
+    'kannada': [
+        'kn',
+    ],
+    'persian': [
+        'ks',
```
dwaynebailey: This could be one of three possible scripts https://en.wikipedia.org/wiki/Kashmiri_language and there isn't a Persian script, I'm pretty sure it would fall under Arabic.

translate/lang/data.py
```diff
@@ -239,6 +239,75 @@
 }
 
 
+scripts = {
```
dwaynebailey: Please document what you're using as a reference for these.

translate/filters/checks.py
```diff
@@ -2155,10 +2161,31 @@ def accelerators(self, str1, str2):
...
+        it also ensures accelerators are absent in the target string since some
+        languages do not use accelerators, for example Indic languages.
         """
+        if self.config.language_script:  # no-accelerators check
```
dwaynebailey: The check isn't meant to not run if a script is set.  It is meant to not run when a script does not support accelerators.  The way that this is configured, any scripts defined for languages would cause accelerators not to work.  The idea of scripts is to be able to bulk configure operations.  I.e. accelerator check does not work for Devengari.  That way add a language code to Devengari and we automatically know that we don't need to run checks.

This seems like a fundamental oversite here ito implementation.

translate/filters/checks.py
unho: Ok.

translate/filters/test_checks.py
unho: It is a leftover. Will remove reference to Assamese.

translate/filters/test_checks.py
unho: Will do. All of these were copied, and some were trimmed below 80 chars. Will put that new example you suggest in the place these examples were copied from too.

translate/lang/data.py
unho: Ok. I used Wikipedia. Will put comments in the code here.

translate/lang/data.py
unho: Will do. I wasn't aware such an standard existed. Nice.

translate/lang/data.py
unho: Sure. Might be worth adding some test to ensure it works.

translate/lang/data.py
unho: Will adjust.

translate/filters/checks.py
unho: Will adjust by adding a list of scripts the MozillaChecker skips in the accelerators check. Will keep this list though to figure out if the language uses a script that should be skipped.

unho:
Updated with the requested changes.

translate/lang/data.py
```diff
@@ -239,6 +239,78 @@
...
+    ],
+    # Unable to find the codes for the following scripts.
+    'assamese': [
+        'as',  # https://en.wikipedia.org/wiki/Assamese_language
+    ],
+    'perso-arabic': [
+        'ks',  # https://en.wikipedia.org/wiki/Kashmiri_language
+    ],
+    'chinese': [
+        'yue',  # https://en.wikipedia.org/wiki/Yue_Chinese
```
dwaynebailey: That's because these and Hans and Hant (not actually Chinese here), but does depend on whether they've chosen Simplified and Complex scripts.  But I think this is OK here.

dwaynebailey:
lgtm

PR #3488, opened 2016-08-24
Bump and pin some requirements

unho:


claudep:
Looks like vobject 0.9.2 supports Python 3. I can have a look if you want.

unho:
@claudep Thanks but it is already on this PR, but somehow messed commits. I am fixing it now.

unho:
No, it seems it is ok.

unho:
> Looks like vobject 0.9.2 supports Python 3. I can have a look if you want.

@claudep It would be great if you can look at the failing tests.

claudep:
vobject upgrade proposed in #3490

lnussel:
Is there a specific reason why six>=1.10.0 is required instead of 1.9 as before?

unho:
This specific requirement was bumped to 1.10.0 only in order to use the latest six version available.

lnussel:
ic. So we will remove that requirement for openSUSE Leap 42.3 as we have six 1.9 there

PR #3490, opened 2016-08-24
Bump vobject version and add Python 3 ical support

claudep:


unho:
gtm

unho:
@claudep Thanks a lot!

PR #3492, opened 2016-08-25
[WIP] Stop ttk from continually trying to import nonsense

phlax:
Hack: cache reimported modules

### DIFF
--- a/translate/lang/factory.py
+++ b/translate/lang/factory.py
@@ -25,7 +25,7 @@ from translate.lang import common, data
 prefix = "code_"
 
 
-def getlanguage(code):
+def _getlanguage(code):
     """This returns a language class.
 
     :param code: The ISO 639 language code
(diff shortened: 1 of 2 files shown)


claudep:
An alternative could be to use the https://docs.python.org/3/library/functools.html#functools.lru_cache decorator. Django does that, but as it is Python 3-only, a compat file has been added to support Python 2.7 (https://github.com/django/django/blob/master/django/utils/lru_cache.py).

phlax:
@claudep yep - i probably will  - this was a quick hack just to see how it affected timings in pootle tests

I am going to use some kind of registry tho - because this dynamic __import__ pattern doesnt look so good to me

nijel:
As Python 3 is now required, I've implemented it using lru_cache, see https://github.com/translate/translate/pull/4067

phlax:
great, closing in favour of your PR

PR #3497, opened 2016-09-14
Moz lang headers

phlax:
Adds support for moz_lang headers, and tag comments

### DIFF
--- a/translate/storage/mozilla_lang.py
+++ b/translate/storage/mozilla_lang.py
@@ -49,7 +49,12 @@ class LangUnit(base.TranslationUnit):
         else:
             target = self.target
         if self.getnotes():
-            notes = ('\n').join(["# %s" % note for note in self.getnotes('developer').split("\n")])
+            notes = ('\n').join(
+                [("#%s" % note
+                  if note.startswith("#")
+                  else "# %s" % note)
+                 for note
+                 in self.getnotes('developer').split("\n")])
             return u"%s\n;%s\n%s%s" % (notes, self.source, target, unchanged)
         return u";%s\n%s%s" % (self.source, target, unchanged)
 
@@ -71,6 +76,7 @@ class LangStore(txt.TxtFile):
     def __init__(self, inputfile=None, mark_active=False, **kwargs):
         self.is_active = False
         self.mark_active = mark_active
+        self._headers = []
         super(LangStore, self).__init__(inputfile, **kwargs)
 
     def parse(self, lines):
@@ -88,6 +94,10 @@ class LangStore(txt.TxtFile):
                 self.is_active = True
                 continue
 
+            if line.startswith("## ") and not line.startswith('## TAG'):
+                self._headers.append(line)
+                continue
+
             if len(line) == 0 and not readyTrans:  # Skip blank lines
                 continue
 
@@ -102,8 +112,12 @@ class LangStore(txt.TxtFile):
                 readyTrans = False  # We already have our translation
                 continue
 
-            if line.startswith('#') and not line.startswith('##'):
-                # Read comments, but not meta tags (e.g. '## TAG')
+            is_comment = (
+                line.startswith('#')
+                and (not line.startswith("##")
+                     or line.startswith('## TAG')))
+            if is_comment:
+                # Read comments, *including* meta tags (i.e. '## TAG')
                 comment += line[1:].strip() + "\n"
 
             if line.startswith(';'):
@@ -117,6 +131,13 @@ class LangStore(txt.TxtFile):
     def serialize(self, out):
         if self.is_active or self.mark_active:
             out.write(b"## active ##\n")
+        for header in self._headers:
+            out.write(six.text_type("%s\n" % header).encode('utf-8'))
+        if self._headers:
+            out.write(b"\n\n")
         for unit in self.units:
             out.write(six.text_type(unit).encode('utf-8'))
             out.write(b"\n\n\n")
+
+    def getlangheaders(self):
+        return self._headers
--- a/translate/storage/mozilla_lang.py
+++ b/translate/storage/mozilla_lang.py
@@ -80,12 +80,10 @@ class LangStore(txt.TxtFile):
         super(LangStore, self).__init__(inputfile, **kwargs)
 
     def parse(self, lines):
-        # Have we just seen a ';' line, and so are ready for a translation
-        readyTrans = False
+        source_unit = None
         comment = ""
         if not isinstance(lines, list):
             lines = lines.split(b"\n")
-        unit = None
 
         for lineoffset, line in enumerate(lines):
             line = line.decode(self.encoding).rstrip("\n").rstrip("\r")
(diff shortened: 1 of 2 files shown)


translate/storage/test_mozilla_lang.py
```diff
@@ -121,3 +127,59 @@ def test_ok_translations(self, ok, target, istranslated):
...
+                "## NOTE: foo\n"
+                "## TAG: fooled_you ##\n"
+                ";Source\n"
+                "Target\n"
+                "\n\n")
+        store = self.StoreClass.parsestring(lang)
+        assert "## TAG: fooled_you ##" not in store.getlangheaders()
+
+    def test_tag_comments(self):
+        """Ensure we can handle and preserve miltiline comments"""
```
phlax: need to change these docstrings

translate/storage/test_mozilla_lang.py
```diff
@@ -121,3 +127,59 @@ def test_ok_translations(self, ok, target, istranslated):
...
+        lang = ("## active ##\n"
+                "# First comment\n"
+                "## TAG: important_tag\n"
+                "# Second comment\n"
+                "# Third comment\n"
+                "## TAG: another_important_tag\n"
+                ";Source\n"
+                "Target\n"
+                "\n\n")
```
unho: When it comes to snippets of any format I find preferable to use triple quoted strings so we can copy-paste verbatim snippets.

translate/storage/mozilla_lang.py
```diff
@@ -116,6 +130,13 @@ def parse(self, lines):
...
             out.write(b"## active ##\n")
+        for header in self._headers:
+            out.write(bytes("%s\n" % header))
+        if self._headers:
+            out.write(b"\n\n")
         for unit in self.units:
             out.write(six.text_type(unit).encode('utf-8'))
             out.write(b"\n\n\n")
+
+    def getlangheaders(self):
```
unho: I wonder if instead of `self._headers` it is better to just have `self.headers`. That way we can skip this extra method.

translate/storage/mozilla_lang.py
```diff
@@ -116,6 +130,13 @@ def parse(self, lines):
     def serialize(self, out):
         if self.is_active or self.mark_active:
             out.write(b"## active ##\n")
+        for header in self._headers:
```
unho: I would use `line` and `self.header`. It is not like you have multiple headers at the beginning of the file, just 1 header with multiple lines.

translate/storage/mozilla_lang.py
phlax: i want to explicitly make this the public method - afaict the only other format that has headers implemented is po - and in that case its called getpoheaders (iirc) - because they are format-specific

phlax: we might want to think about a generic approach to handling headers - but unless we do i think it should be explicit

translate/storage/mozilla_lang.py
phlax: im not really following your point here

translate/storage/test_mozilla_lang.py
```diff
@@ -121,3 +127,59 @@ def test_ok_translations(self, ok, target, istranslated):
...
+        lang = ("## active ##\n"
+                "## some_tag ##\n"
+                "## another_tag ##\n"
+                "## NOTE: foo\n"
+                "## TAG: fooled_you ##\n"
+                ";Source\n"
+                "Target\n"
+                "\n\n")
+        store = self.StoreClass.parsestring(lang)
+        assert "## TAG: fooled_you ##" not in store.getlangheaders()
```
unho: It would be nice to test too `MAX_LENGTH`

unho: I guess it doesn't belong to this PR.

translate/storage/test_mozilla_lang.py
```diff
@@ -121,3 +127,59 @@ def test_ok_translations(self, ok, target, istranslated):
...
+            == ("## active ##\n"
+                "## some_tag ##\n"
+                "## another_tag ##\n"
+                "## NOTE: foo\n"
+                "\n\n"
+                ";Source\n"
+                "Target\n"
+                "\n\n"))
+
+    def test_not_headers(self):
```
unho: I would add another test to ensure that

```
lang = """## active ##
## some_tag ##
## another_tag ##
## NOTE: foo
## TAG: fooled_you ##
# Note for string
;Source
Target


"""
```

correctly parses the unit comment.

translate/storage/mozilla_lang.py
unho: I am against this specific things. They will just make more difficult to get to a common API. I would say have `headers` always, and if the format doesn't support them just have empty list and output empty string. After all it is what this implementation is doing.

translate/storage/mozilla_lang.py
unho: The file has a single header, not multiple headers. That single header can have multiple lines. So I suggest to alter the variables names to reflect that.

phlax: thats not correct i dont believe

translate/storage/mozilla_lang.py
phlax: great - please raise a PR to address these more fundamental concerns and we can think about it then

translate/storage/test_mozilla_lang.py
phlax: personally i really dislike triple quotes in methods etc- but can we focus on the important things please

unho:
Despite I would love the concerns addressed, lgtm given that this is just a patch to keep going.

translate/storage/mozilla_lang.py
dwaynebailey: I guess the distinction is headerline (a single line in a header) vs header (a group of items making up a header) vs headers (a number of header blocks).  Tbh I'm not that fussed

translate/storage/test_mozilla_lang.py
```diff
@@ -38,6 +40,10 @@ def test_comments(self):
         unit.addnote("Hello")
         assert str(unit).find("Hello") == 2
         assert str(unit).find("# Hello") == 0
+        unit.addnote("# TAG: goodbye")
```
dwaynebailey: Ideally I'd test this on its own or adapt the comment

dwaynebailey:
Some nits, otherwise lgtm

translate/storage/test_mozilla_lang.py
phlax: this is addressed

translate/storage/test_mozilla_lang.py
unho: I dislike them too, but in this specific scenario the advantages are greater than the drawbacks and my personal opinion against them.

Specifically using triple quotes you don't have to escape everything, so you can copy-paste easily and you don't lose plenty of time escaping. Also the snippet looks like like in real life so it is considerably easier for humans to understand it, and therefore understand what the test is about. Just my opinion.

translate/storage/test_mozilla_lang.py
phlax: i put together because the behaviour changed - and i felt like it should go here - ive updated the comment

translate/storage/test_mozilla_lang.py
```diff
@@ -121,3 +127,87 @@ def test_ok_translations(self, ok, target, istranslated):
...
+                "## TAG: another_important_tag\n"
+                "# Another comment\n"
+                ";Source\n"
+                "Target\n"
+                "\n\n")
+        store = self.StoreClass.parsestring(lang)
+        assert not store.getlangheaders()
+        assert (
+            "First comment"
+            in store.units[0].getnotes(origin="developer").split("\n"))
```
unho: Is it not better to get all the dev notes for the first units and store in variable? Considerable shorter asserts I would say.

phlax: i think this is more explicit

PR #3500, opened 2016-09-21
Added YAML support

### LINKED ISSUE #3248: Support ruby/rails i18n yml files
I really missed the support for ruby/rails i18n yml format. Do you have any milestone planned to support it?

Example file: https://github.com/noosfero/noosfero/blob/master/config/locales/en-US.yml

Issue first opened at https://github.com/nijel/weblate/issues/645

nijel:
There are two classes to support two different YAML formats:

YAMLFile : generic key/value format
RubyYAMLFile : Ruby format with target language as root node

Fixes #3248

### DIFF
--- a/translate/storage/test_yaml.py
+++ b/translate/storage/test_yaml.py
@@ -1,0 +1,120 @@
+# -*- coding: utf-8 -*-
+
+import sys
+from io import BytesIO
+from translate.storage import yaml, test_monolingual
+
+
+class TestYAMLResourceUnit(test_monolingual.TestMonolingualUnit):
+    UnitClass = yaml.YAMLUnit
+
+
+class TestYAMLResourceStore(test_monolingual.TestMonolingualStore):
+    StoreClass = yaml.YAMLFile
+
+    def test_serialize(self):
+        store = yaml.YAMLFile()
+        store.parse('key: value')
+        out = BytesIO()
+        store.serialize(out)
+
+        assert out.getvalue() == b'key: value\n'
+
+    def test_edit(self):
+        store = yaml.YAMLFile()
+        store.parse('key: value')
+
+        store.units[0].settarget('second')
+
+        out = BytesIO()
+        store.serialize(out)
+
+        assert out.getvalue() == b'key: second\n'
+
+    def test_edit_unicode(self):
+        store = yaml.YAMLFile()
+        store.parse('key: value')
+
+        store.units[0].settarget(u'zkouška')
+
+        out = BytesIO()
+        store.serialize(out)
+
+        assert out.getvalue() == u'key: zkouška\n'.encode('utf-8')
+
+    def test_ordering(self):
+        store = yaml.YAMLFile()
+        store.parse('''
+foo: foo
+bar: bar
+baz: baz
+''')
+
+        assert store.units[0].source == 'foo'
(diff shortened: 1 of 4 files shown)


translate/storage/test_yaml.py
```diff
@@ -0,0 +1,120 @@
...
+import sys
+from io import BytesIO
+from translate.storage import yaml, test_monolingual
+
+
+class TestYAMLResourceUnit(test_monolingual.TestMonolingualUnit):
+    UnitClass = yaml.YAMLUnit
+
+
+class TestYAMLResourceStore(test_monolingual.TestMonolingualUnit):
```
unho: I guess this should be `test_monolingual.TestMonolingualStore`.

translate/storage/yaml.py
```diff
@@ -0,0 +1,230 @@
...
+
+from collections import OrderedDict
+import os
+
+import six
+import yaml
+
+from translate.storage import base
+
+import yaml.constructor
```
unho: This import should be before translate toolkit one.

translate/storage/test_yaml.py
```diff
@@ -0,0 +1,120 @@
...
+        store.serialize(out)
+
+        assert out.getvalue() == b'''foo:
+  bar: bar
+  baz:
+    boo: booo
+'''
+
+
+class TestRubyYAMLResourceStore(test_monolingual.TestMonolingualUnit):
```
unho: This one should be `test_monolingual.TestMonolingualStore` too.

translate/storage/test_yaml.py
unho: If you change this you will see that several tests fail. Please fix them.

unho:
Very nice work, but if you correct the two misspelled classes in tests it unveils several failing tests. Please fix that.

unho:
On a related note I was able to very quickly code some converters on top of this.

unho:
It would be great too if you can further adjust this so comments are extracted. Here you have an example file with comments https://github.com/openstreetmap/openstreetmap-website/blob/master/config/locales/en.yml#L44-L45

translate/storage/test_yaml.py
nijel: Oops, I've based the tests on resx and the same issue is there (translate/storage/test_resx.py).

nijel:
Indeed supporting comments would be great, but I've decided to ignore this part for initial implementation as PyYAML support for this is poor. There is PyYAML fork which should be better (https://pypi.python.org/pypi/ruamel.yaml), but I have no experience with it.

Anyway I've addressed all issues you've pointed out (+ rebased changes on current master).

translate/storage/test_yaml.py
nijel: See #3504 and #3505.

unho: Yes, thanks a lot for those PRs.

unho:
We can live without comments for now. Landed in master.

rvanlaak:
@unho awesome! Any news about when the next beta will be released? And @nijel how can I let Weblate install that specific beta version of `translation-toolkit`? 👍

nijel:
You can use ``pip install 'translate-toolkit>=2.0.0b4'`` (or something similar).

PR #3505, opened 2016-10-03
Fix JSON tests

nijel:
Fixes wrongly used base class in JSON tests and resulting test failures.

I've just spotted this bug in code as I've made same mistake in #3500 and the same problem is in resx (see #3504).


nijel:
Reworked the PR to fix the tests as well.

unho:
Landed myself to master since we wanted this change in and we are proceeding to release soon. Thanks for your contribution!

PR #3509, opened 2016-10-14
[requires.io] dependency update on stable/1.14.0 branch

dwaynebailey:


unho:
We are not going to merge this. This is an unsupported branch.

PR #3511, opened 2016-10-14
[requires.io] dependency update on stable/1.12.0 branch

dwaynebailey:


unho:
We are not going to merge this. The version is too old.

PR #3512, opened 2016-10-20
Change YAML files look

nijel:
- In most cases frameworks do not wrap lines at all.
- The values are usually enclosed in double quotes.

### DIFF
--- a/translate/storage/test_yaml.py
+++ b/translate/storage/test_yaml.py
@@ -17,22 +17,22 @@ class TestYAMLResourceStore(test_monolingual.TestMonolingualStore):
 
     def test_serialize(self):
         store = yaml.YAMLFile()
-        store.parse('key: value')
+        store.parse('key: "value"')
         out = BytesIO()
         store.serialize(out)
 
-        assert out.getvalue() == b'key: value\n'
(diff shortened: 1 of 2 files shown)


unho:
@nijel AFAICT this is a deviation from the standard. Can you please add this as a dialect for the format.

nijel:
Looking at real life files, they almost always use this:

* https://github.com/rails/rails/blob/master/activemodel/lib/active_model/locale/en.yml
* https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml

But indeed some of them are mixed:

* https://github.com/noosfero/noosfero/blob/master/config/locales/en-US.yml

unho:
The usage of mixed format really complicates things. @nijel would it be possible the store class can parse any of the formats and also mixed?

nijel:
There is no problem in parsing, the problem is to keep style on saving. What might help is using https://pypi.python.org/pypi/ruamel.yaml instead of PyYAML, but I have no experience with that so far.

nijel:
Closing as we switched to different library in https://github.com/translate/translate/pull/3863 and this no longer applies.

PR #3514, opened 2016-10-28
YAML converters

unho:


rimas-kudelis:
What's been holding this one off for so long?

unho:
Rebased on top of latest master and removed the commits were other converters were being refactored, but this still needs more tests before being ready.

unho:
Also need to adjust to support Ruby dialect.

rimas-kudelis:
@unho may I suggest to move Ruby dialect support into its a separate issue/PR?

unho:
Not sure how I messed up merging the PR, but the commits in this PR have been landed.

Issue #2162, opened 2012-05-02
Use colour for pocount ouput to increase readabilty


dwaynebailey:
_Version: trunk_

Use terminal colour syntax to output pocount using colour.

nijel:
Implemented in https://github.com/translate/translate/pull/3517

PR #3517, opened 2016-11-04
Add color support for pocount output

adhikasp:
Hello, i see that issue #2162 is still uncompleted for a long time.

I am adding color support with ANSI color code that enabled by default, but can be toggled off with `pocount --no-color`.

Here are some screenshot

Edit: I also change the output text a bit, so I'm altering the test data too.
Edit 2: I'm beginner with testing and travis thing, sorry for the failing build. Will fix it soon.

example 1

example 2

example 3

example 4

### DIFF
--- a/translate/tools/pocount.py
+++ b/translate/tools/pocount.py
@@ -46,6 +46,38 @@ style_full, style_csv, style_short_strings, style_short_words = range(4)
 default_style = style_full
 
 
+class ConsoleColor:
+    """Class to implement color mode.
+    """
+    # print using color? Default to true
+    color_mode = True
+    COLOR_PURPLE = "\033[95m"
+    COLOR_GREEN = "\033[92m"
+    COLOR_YELLOW = "\033[93m"
+    COLOR_RED = "\033[91m"
+    COLOR_DEFAULT = "\033[0m"
+
+    @classmethod
+    def HEADER(cls):
+        return cls.COLOR_PURPLE if ConsoleColor.color_mode else ""
+
+    @classmethod
+    def OKGREEN(cls):
+        return cls.COLOR_GREEN if ConsoleColor.color_mode else ""
+
+    @classmethod
+    def WARNING(cls):
+        return cls.COLOR_YELLOW if ConsoleColor.color_mode else ""
+
+    @classmethod
+    def FAIL(cls):
+        return cls.COLOR_RED if ConsoleColor.color_mode else ""
+
+    @classmethod
+    def ENDC(cls):
+        return cls.COLOR_DEFAULT if ConsoleColor.color_mode else ""
+
+
 def calcstats_old(filename):
     """This is the previous implementation of calcstats() and is left for
     comparison and debuging purposes.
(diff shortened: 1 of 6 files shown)


adhikasp:
I already update my change and the Travis build is successfull.

translate/tools/pocount.py
```diff
@@ -46,6 +46,34 @@
 default_style = style_full
 
 
+class bcolors:
+    """Class to implement color mode
```
jayvdb: Add a `.` at the end of the sentence per PEP257

translate/tools/pocount.py
```diff
@@ -46,6 +46,34 @@
 default_style = style_full
 
 
+class bcolors:
```
jayvdb: What does "b" refer to?  Expand the name?

Also the class name should start with a Capital letter per PEP 8.

translate/tools/pocount.py
```diff
@@ -46,6 +46,34 @@
...
+    """Class to implement color mode
+    """
+    @staticmethod
+    def HEADER():
+        return "\033[95m" if bcolors.color_mode else ""
```
jayvdb: The color strings should be CONSTANTS as class attributes.

translate/tools/pocount.py
```diff
@@ -46,6 +46,34 @@
...
+    def FAIL():
+        return "\033[91m" if bcolors.color_mode else ""
+
+    @staticmethod
+    def ENDC():
+        return "\033[0m" if bcolors.color_mode else ""
+
+
+# print using color? Default to true
+bcolors.color_mode = True
```
jayvdb: set this attribute inside the class, which helps static analysers know that `color_mode` is an attribute while lexing the static methods which use it.

adhikasp:
Thanks for the review @jayvdb, I already update as per your suggestion  👍 ✨ 
Please correct me if something still missing in my code.

translate/tools/pocount.py
```diff
@@ -46,6 +46,38 @@
...
+    color_mode = True
+    COLOR_PURPLE = "\033[95m"
+    COLOR_GREEN = "\033[92m"
+    COLOR_YELLOW = "\033[93m"
+    COLOR_RED = "\033[91m"
+    COLOR_DEFAULT = "\033[0m"
+
+    @staticmethod
+    def HEADER():
+        return COLOR_PURPLE if ConsoleColor.color_mode else ""
```
unho: I tried running this PR locally, but it is not able to find `COLOR_PURPLE` (`__main__: ERROR: global name 'COLOR_PURPLE' is not defined`). This can be fixed if this method is changed like:

```python
    @classmethod
    def HEADER(cls):
        return cls.COLOR_PURPLE if ConsoleColor.color_mode else ""
```

Note the usage of `@classmethod` and `cls` to reference the class.

The same happens for all the other methods in this class, so please adjust them the same way.

unho:
This is a fantastic contribution. Please address the raised concern so we can land.

adhikasp:
Thanks @unho, I already update the PR

unho:
Merged. @adhikasp thanks a lot for your contribution!

PR #3522, opened 2016-11-14
Add option to podebug to avoid rewriting placeholders

r-o-b-b-i-e:
This change adds the optional `--preserveplaceholders` flag to podebug, which when enabled ensures that "unicode" and "flipped" rewrites are not applied to sub-strings identified as placeholders. In this way, placeholders remain unchanged in the output so that systems working with the rewritten strings can still inject data at run-time as normal.

At the moment it can identify and preserve placeholders in four 'styles' that are in use at Skyscanner:
- `{{doublebraces}}`
- `{singlebraces}`
- `@@doubleatsymbols@@`
- `<htmltags>`

All of these are held in a single regex that should be easy to extend if other styles are required in future.

### DIFF
--- a/translate/storage/placeables/general.py
+++ b/translate/storage/placeables/general.py
@@ -171,6 +171,24 @@ class FormattingPlaceable(Ph):
     parse = classmethod(regex_parse)
 
 
+class DoubleAtPlaceable(Ph):
+    istranslatable = False
+    iseditable = False
+    # Matches placeholders that use two at symbols @@placeable@@.
+    regex = re.compile(r"@@.*?@@", re.VERBOSE)
+    parse = classmethod(regex_parse)
+
+
+class BracePlaceable(Ph):
+    istranslatable = False
+    iseditable = False
+    # Matches placeholders that use two braces {{placeable}} or one brace {placeable}.
+    # The negative character groups with closing brace [^}] stop the regex from counting several placeholders
+    # as one (e.g. '{open}something{closed}' should produce two distinct BracePlaceables '{open}' and '{closed}')
+    regex = re.compile(r"{{[^}]*}}|{[^}]*}", re.VERBOSE)
+    parse = classmethod(regex_parse)
+
+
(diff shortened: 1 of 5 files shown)


r-o-b-b-i-e:
Hi, I'm aware that this pull request has been sitting for a while -- if there is anything further you need from me before it can get merged, please let me know!

dwaynebailey:
Whoops @r-o-b-b-i-e sorry for missing this.  I think I'd like to see us use placeholders from either the storage/placeholders code or actually use the placeholder code from checks.  If we can reuse those it would make it much easier to extend and reuse in one place.

r-o-b-b-i-e:
@dwaynebailey interesting, I didn't know there was any code available to provide this sort of thing already. I'll track it down and see what I can do, thanks!

r-o-b-b-i-e:
Okay @dwaynebailey, I have reworked this to use the storage/placeables code as you suggested -- this feels a lot better than what I had before so thanks for the tip! I had a look at the checks code as well, but it didn't seem to be such a good fit for this scenario.

The new version now parses the incoming string into StringElems and uses the `is_translatable` flag on each element to determine whether that part should be transformed or not; placeholders have that as False whereas generic chunks of string have it True and this distinction feels much cleaner than juggling string indices and regexes directly.

I needed to add two new placeable classes to support the additional placeholder styles we use in Skyscanner (single/double-brace, double-at); I chose not to add these to the complete parser list as that is dependent on order and I'm not sure where they would go in there. Some of the other parsers didn't seem appropriate for this use case (e.g. punctuation and newlines) either, so I created a smaller list specifically for this placeholder preservation logic in podebug. Happy to adjust this if you think it would be better to use the full parser list and add our new placeables somewhere in there!

Phew, I think that's everything. Let me know what you think of the new version!

unho:
Merged in f681ff51. @r-o-b-b-i-e thanks for your contribution, indeed newer approach looks better.

PR #3524, opened 2016-11-16
Use print_function in all places print is used with multiple parameters

nijel:
Otherwise output is different on Python 2 (which prints tuple) and
Python 3.

I think the example says it all:
```
$ python2 -c 'print(1, 2)'
(1, 2)
$ python3 -c 'print(1, 2)'
1 2
```

### DIFF
--- a/tests/odf_xliff/test_odf_xliff.py
+++ b/tests/odf_xliff/test_odf_xliff.py
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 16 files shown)


claudep:
In test files, does the difference has any impact?
In non test files, I'd rather change the `print(a, b)` to placeholder syntax (%s).

nijel:
The output is used only in case of test failure, that's how I noticed it's producing strange output on Python 2. Changing to placeholder syntax is clearly an option as well, I've chosen the less intrusive way.

unho:
Landed in 02caa8d6. Thanks for your contribution.

PR #3525, opened 2016-11-16
Improve Android resource tests

nijel:
- use valid unit ID for testing
- add more tests for Android escaping


unho:
Landed as f61668d0. Thanks for your contribution.

PR #3526, opened 2016-11-16
Set correct pointer types for use of libgettextpo

llimeht:
For 64 bit Python, the use of `c_int` to mean a pointer to structure is not safe. Recent updates to the Python interpreter have meant that this undefined behaviour causes segmentation faults.

These changes

 * create pointer types for each of the structs involved
 * sets the correct types for all the functions in libgettextpo that are used
 * fix up the usage of `po_message_msgstr_plural`/`po_message_set_msgstr_plural` in line with the argtypes

For more details, see

  http://bugs.python.org/issue28631

Thanks to Eryk Sun and Matthias Klose for their assistance.

### DIFF
--- a/translate/storage/cpo.py
+++ b/translate/storage/cpo.py
@@ -53,14 +53,40 @@ STRING = c_char_p
 
 # Structures
 class po_message(Structure):
-    _fields_ = []
+    pass
+
+
+class po_file(Structure):
+    pass
+
+
+class po_filepos(Structure):
+    pass
+
+
+class po_iterator(Structure):
+    pass
+
+
+po_message_t = POINTER(po_message)
+"""A po_message_t represents a message in a PO file."""
+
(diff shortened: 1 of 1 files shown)


unho:
Landed in 543b6a5b despite there are still plenty of tests failing (they are allowed failures). @llimeht it would be great if you can contribute fixes in that front. Thanks for your contribution.

llimeht:
This change was to allow `test_cpo.py` to pass (and not segfault the interpreter!) with `USECPO=0`.

The use of  `USECPO=1` or `USECPO=2` is a different problem. `USECPO=1` exposes a lot of places where translate-toolkit (mis)uses the libgettext-po API such that lots of tests fail with libgettext-po aborting and crashing the interpreter. With `USECPO=2`, the current state is about 10 failing tests. I'm not sure how easy it will be to solve them but I'll look.

PR #3528, opened 2016-11-23
ts: Write quotes as entities

nijel:
This is improvement on top of #3470. It removes not necessary encoding/decoding of unicode strings gaining about 20% of performance.


unho:
Merged in be690d3b. Thanks @dg0yt and @nijel.

PR #3532, opened 2016-12-05
[requires.io] dependency update on master branch

dwaynebailey:

PR #3534, opened 2016-12-05
Checks: store categories in checker instances

### LINKED ISSUE #3533: Check categories are mixed up after subsequent runs
The check categories are assigned to the *class* upon respective method execution via the decorators. This is problematic if some code wants to read the categories for a specific checker after another checker has run, because the class attribute is shared across instances.

This breaks code/tests in Pootle, e.g. [`get_qualitychecks`](https://github.com/translate/pootle/blob/fcc0a3a9467b856284a661197ce0862fc07a806a/pootle/apps/pootle_misc/checks.py#L1100-L1125).

The weird behavior can be demonstrated with this test, which is expected to fail:
```python
def test_category():
    """Tests checker categories aren't mixed up."""
    from translate.storage import base
    unit = base.TranslationUnit(u'foo')
    unit.target = u'bar'

    standard_checker = checks.StandardChecker()
    assert standard_checker.categories == {}
    standard_checker.run_filters(unit)
    assert standard_checker.categories != {}
    assert 'validxml' not in standard_checker.categories.keys()
    standard_categories_count = len(standard_checker.categories.values())

    libo_checker = checks.LibreOfficeChecker()
    assert libo_checker.categories == {}
    libo_checker.run_filters(unit)
    assert libo_checker.categories != {}
    assert 'validxml' in libo_checker.categories.keys()

    standard_checker = checks.StandardChecker()
    assert standard_checker.categories == {}
    standard_checker.run_filters(unit)
    assert standard_checker.categories != {}
    assert len(standard_checker.categories.values()) == standard_categories_count
    assert 'validxml' not in standard_checker.categories.keys()
```

Leaving aside possible improvements to how tests are categorized and executed, a less surprising behavior would be to store the categories on instances. I will PR a patch in this direction.

julen:
Having the categories in classes is problematic as it is shared across instances
and subsequent check runs end up modifying the list of categories, which is
misleading for 3rd party code that wants to access a list of categories for a
particular checker.

Fixes #3533.

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -307,13 +307,13 @@ class UnitChecker(object):
 
     preconditions = {}
 
-    #: Categories where each checking function falls into
-    #: Function names are used as keys, categories are the values
-    categories = {}
-
     def __init__(self, checkerconfig=None, excludefilters=None,
                  limitfilters=None, errorhandler=None):
         self.errorhandler = errorhandler
+
+        #: Categories where each checking function falls into
+        #: Function names are used as keys, categories are the values
+        self.categories = {}
 
         if checkerconfig is None:
             self.setconfig(CheckerConfig())
--- a/translate/filters/decorators.py
+++ b/translate/filters/decorators.py
@@ -35,8 +35,8 @@ def critical(f):
 
     @wraps(f)
     def critical_f(self, *args, **kwargs):
-        if f.__name__ not in self.__class__.categories:
-            self.__class__.categories[f.__name__] = Category.CRITICAL
+        if f.__name__ not in self.categories:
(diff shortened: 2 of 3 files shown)


julen:
The Travis failure happens because of a newer Sphinx version which someone will need to pin (the same thing happens in #3532 for instance).

julen:
> The Travis failure happens because of a newer Sphinx version which someone will need to pin (the same thing happens in #3532 for instance).

This is now fixed and rebased.

unho:
Landed in https://github.com/translate/translate/commit/22061c8574942fc11e23e21ece9f2821a69fa723 Thanks a lot for your contribution.

PR #3544, opened 2016-12-28
Romanian checks: niciun, nicio and pythonic refactoring

Jobava:
Added more checks for Romanian for "nici un" and "nici o" (a common error).

On top of that, code simplification since reduce logic was flawed for substrings.

### DIFF
--- a/translate/lang/ro.py
+++ b/translate/lang/ro.py
@@ -24,8 +24,6 @@
 
 from __future__ import unicode_literals
 
-from functools import reduce
-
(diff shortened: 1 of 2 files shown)


Jobava:
As for pythonic vs. non-pythonic ways to do matchings, I put together some ipython snippets: https://gist.github.com/Jobava/8831cda7b95439df905fd85d441fd512

unho:
Reviewed and tested. Thanks a lot!

PR #3546, opened 2017-01-09
Do not bundle selector

### LINKED ISSUE #3453: Replace Selector with requirement
Code in TTK is from 2006. It requires updating.

Keeping it there would require us maintaining it, for a non-core feature of TTK. We can probably remove the code and instead include it as a requirement: https://github.com/lukearno/selector

unho:
This is to be merged after final 2.0.0 release.

Recommended requirements should pull required ones. Otherwise building docs on
Travis fails.

Fixes #3453.


nijel:
I guess this has been superseded by https://github.com/translate/translate/pull/3454, right?

unho:
I was keeping this one open in the hope it would be eventually landed when issues are fixed. But I just checked and it doesn't look like the best approach for this piece of code since the package in PyPI hasn't been updated to support Python 3 well enough. So better keep our bundled and tweaked copy of `selector`. Therefore closing this PR.

PR #3547, opened 2017-01-10
Fix RomanianChecker and add tests for Romanian checks in pofilter

unho:
Just that, more tests.

### DIFF
--- a/translate/lang/ro.py
+++ b/translate/lang/ro.py
@@ -24,7 +24,7 @@
 
 from __future__ import unicode_literals
 
-from translate.filters.checks import TranslationChecker
+from translate.filters.checks import FilterFailure, TranslationChecker
 from translate.filters.decorators import cosmetic
 from translate.lang import common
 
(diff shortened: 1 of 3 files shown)


translate/lang/ro.py
```diff
@@ -47,17 +47,22 @@ def contains_illegal(self, illegal_sequence, target_str):
...
-               :param str2: the target (translated) string
-               :return: True if str2 contains a cedilla character
+        """Check if the translation contains an illegal cedilla character
+
+        Cedillas are obsoleted diacritics for Romanian:
+
+          - U+0162 Latin capital letter T with cedilla
+          - U+0163 Latin small letter T with cedilla
+          - U+015E Latin capital letter S with cedilla
+          - U+015F Latin small letter S with cedilla
```
dwaynebailey: Haha, that's a hack of note :) Now why didn't I think of that.  I'm good with this.

translate/lang/ro.py
```diff
@@ -32,19 +32,6 @@
 class RomanianChecker(TranslationChecker):
     """A Checker class for Romanian"""
 
-    def contains_illegal(self, illegal_sequence, target_str):
```
dwaynebailey: Keep this but just move it to a private function outside of the class.

unho: Done.

PR #3548, opened 2017-01-12
Enable custom pofilter checks

### LINKED ISSUE #3129: I would like to be able to run my own tests with pofilter
_Version: trunk_

_Originally posted by simohe:_

My suggestion for the call is:
pofilter -t path/to/my_check.py ...

The file my_check.py should have a defined hooks. (a method named pfcheck?)

Running unknown code is a bit risky, but the user has to explicitly tell the path, so he should be aware of what he is doing.

tropikhajma:
Fixes #3129

### DIFF
--- a/translate/filters/pofilter.py
+++ b/translate/filters/pofilter.py
@@ -31,6 +31,8 @@ for full descriptions of all tests.
 
 import os
 import six
+import sys
+from importlib import import_module
 
 from translate.filters import autocorrect, checks
 from translate.misc import optrecurse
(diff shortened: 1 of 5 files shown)


translate/filters/pofilter.py
```diff
@@ -31,19 +31,27 @@
...
+from importlib import import_module
 
 from translate.filters import autocorrect, checks
 from translate.misc import optrecurse
 from translate.storage import factory
 from translate.storage.poheader import poheader
 
 
-def build_checkerconfig(options):
+def build_checkerconfig(self, options):
```
unho: Why is `self` needed? This is a function, not a method.

unho: Oh, I see.

unho: Actually it doesn't seem like this would work in the case of the test since `BaseTestFilter` does not have `error` method.

unho: Also better use `parser` instead of `self`.

translate/filters/pofilter.py
```diff
@@ -286,6 +297,9 @@ def cmdlineparser():
         "", "--wx", dest="filterclass",
         action="store_const", default=None, const=checks.KdeChecker,
         help="use the standard checks for wxWidgets translations")
+    parser.add_option("", "--custom", dest="customchecks",
+        default=None, type="string", metavar="DIR",
+        help="use custom checks from DIR/custom.py")
```
unho: To get rid of the Travis failure:

```python
    parser.add_option(
        "", "--custom", dest="customchecks",
        default=None, type="string", metavar="DIR",
        help="use custom checks from DIR/custom.py")
```

unho:
Only did a shallow review as I plan not to land this before 2.0.0 final release (merging this might introduce instability when we are just trying to stabilize to release final). Will do a deeper review after the raised concerns are addressed.

PR #3549, opened 2017-01-19
More remove untranslated

unho:
This replaces the `json2po` part of #3266.


dwaynebailey:
lgtm, could you expand the functional test to cover some of the other states but also so that we get a translated file with parts of it removed as expected.

PR #3558, opened 2017-01-25
.lang: parse MAX_LENGTH into comments

dwaynebailey:
Currently we skip MAX_LENGTH

### DIFF
--- a/translate/storage/mozilla_lang.py
+++ b/translate/storage/mozilla_lang.py
@@ -108,7 +108,11 @@ class LangStore(txt.TxtFile):
                 self.is_active = True
                 continue
 
-            if line.startswith("## ") and not line.startswith('## TAG'):
+            header_meta_data = (
+                line.startswith("## ")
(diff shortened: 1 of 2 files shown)


translate/storage/mozilla_lang.py
```diff
@@ -108,7 +108,9 @@ def parse(self, lines):
                 self.is_active = True
                 continue
 
-            if line.startswith("## ") and not line.startswith('## TAG'):
+            if (line.startswith("## ")
+                and not line.startswith('## TAG')
+                and not line.startswith('## MAX_LENGTH')):
```
phlax: can we break up the multiliine conditions as its easy to miss that they are conditions or where the clause is etc

phlax: `any` might make it more readable

```
>>> line = "xx foo"
>>> any(line.startswith(x) for x in ["xx", "yy", "zz"])
```

phlax:
other than nit lgtm

Issue #3499, opened 2016-09-16
translate-toolkit: 2.0.0b5 broken po2txt


elchevive:
Hi,

I received a bug in openSUSE for translate-toolkit 2.0.0b5:

https://bugzilla.opensuse.org/show_bug.cgi?id=999286

I'll past the bug report below:

translate-toolkit 2.0.0b5 breaks po2txt as used by skelcd-openSUSE:

```pytb
$ po2txt --progress=none --threshold=100 license/license.es.po license/license.es.txt
Traceback (most recent call last):
  File "/usr/bin/po2txt", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2927, in <module>
    @_call_aside
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2913, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2940, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 635, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 943, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 829, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'diff-match-patch==20121119' distribution was not found and is required by translate-toolkit
```

If I remove the diff-match-patch line in /usr/lib/python2.7/site-packages/translate_toolkit-2.0.0b5-py2.7.egg-info/requires.txt it still doesn't work:

```console
$ po2txt --progress=none --threshold=100 license/license.es.po license/license.es.txt
po2txt: WARNING: Error processing: input license/license.es.po, output license/license.es.txt, template None: an integer is required
```

Any help debugging the problem? (I'm just a packager with very limited coding capabilities)

unho:
@elchevive Thanks for reporting.

Regarding the first issue we expect diff-match-patch being picked from https://pypi.python.org/pypi/diff-match-patch/ by PyPI.

Regarding the second one I am able to replicate. It seems to work if you don't specify the `--threshold`. We will look further into it.

elchevive:
Thanks.

[...]

PR #3559, opened 2017-01-27
Fix --threshold

dwaynebailey:
unit objects are not hashable, so use getunitid() instead.

Adds functional tests.

Fixes a py3 bug


phlax:
lgtm - i think it fixes #3499

unho:
lgtm

PR #3562, opened 2017-01-28
Fixes for pofilter to work with python 3.

Jobava:
This fixes pofilter to work with python3

How to reproduce the bug:

clone the translate repo somewhere and cd into it

```
virtualenv -p python3 venv_p3
source venv_p3/bin/activate

# then create a test file like below:

echo 'msgid "bla"
msgstr "bla ş"

msgid "cow"
msgstr "bla nici un"

msgid "identical"
msgstr "identical"

msgid "cow2"
msgstr "bla nici un bla ş"' > ro.po

./setup.py build && ./setup.py install

pofilter -i ro.po --errorlevel=traceback
```

=================
OUTPUT
=================

```
pofilter -i ro.po --errorlevel=traceback
processing 1 files...
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE VERSION\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2017-01-28 20:22+0300\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"X-Generator: Translate Toolkit 2.0.0\n"
pofilter: WARNING: Error processing: input ro.po, output None, template None: Traceback (most recent call last):

  File "/home/jobava/Documente/pootle/translate/venv_python3/lib/python3.4/site-packages/translate_toolkit-2.0.0-py3.4.egg/translate/misc/optrecurse.py", line 520, in recursiveprocess
    fulltemplatepath)

  File "/home/jobava/Documente/pootle/translate/venv_python3/lib/python3.4/site-packages/translate_toolkit-2.0.0-py3.4.egg/translate/misc/optrecurse.py", line 576, in processfile
    **passthroughoptions):

  File "/home/jobava/Documente/pootle/translate/venv_python3/lib/python3.4/site-packages/translate_toolkit-2.0.0-py3.4.egg/translate/filters/pofilter.py", line 219, in runfilter
    tofile.serialize(outputfile)

  File "/home/jobava/Documente/pootle/translate/venv_python3/lib/python3.4/site-packages/translate_toolkit-2.0.0-py3.4.egg/translate/storage/pypo.py", line 811, in serialize
    out.write(b'\n')

[...]

PR #3563, opened 2017-01-28
Fixed #3562 -- Set stdout output as binary on Python 3

claudep:


translate/misc/test_optrecurse.py
```diff
@@ -16,3 +17,14 @@ def test_splitext(self):
...
+
+    def test_outputfile_receives_bytes(self, capsys):
+        parser = optrecurse.RecursiveOptionParser({"txt": ("po", None)})
+
+        temp_file = TemporaryFile()
+        with TemporaryFile() as tempfile:
+            out = parser.openoutputfile(None, temp_file.name)
+            out.write(b'binary suff')
+
+        out = parser.openoutputfile(None, None)  # To sys.stdout
```
claudep: Note: To test the failure of the previous code with py.test, you have to pass the `-s` option because py.test replaces sys.stdout with its own capturing object.

unho:
@claudep It seems Travis fails here.

claudep:
Looks like overriding `sys.stdout` is a popular sport! Which makes testing it hard.
I'll try another approach.

translate/misc/test_optrecurse.py
unho: @claudep does this mean that it is necessary to change how to run tests on Travis?

unho:
Merged in https://github.com/translate/translate/commit/536b9344f9c6c2c1f1fee1bf6ce03ea205c140fb Thanks a lot for your contribution!

translate/misc/test_optrecurse.py
claudep: Not at all. But I told that because traditionally, you should demonstrate that the test fails before the fix and that it succeeds after the fix. If you don't run py.test with the `-s`option, the test would still pass before the fix.

unho: Thanks for the clarification. Now I am reading your initial message and I understand it.

PR #3569, opened 2017-02-03
Appveyor for Windows CI build

nschonni:
Noticed that 2.0 says there is Windows suport, so I thought I'd add in a Windows CI setup.
- The commits haven't been cleaned up/squashed till I saw if there was interested
- Currently has 52 failures https://ci.appveyor.com/project/nschonni/translate/build/1.0.8

This could eventually be used to produce the InnoSetup package for tagged builds and uploaded as an artifact for release attachment.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,26 +1,11 @@
 environment:
-
   matrix:
-
-    # For Python versions available on Appveyor, see
-    # http://www.appveyor.com/docs/installed-software#python
-    # The list here is complete (excluding Python 2.6, which
-    # isn't covered by this document) at the time of writing.
-
     - PYTHON: "C:\\Python27"
-    # - PYTHON: "C:\\Python33"
-    # - PYTHON: "C:\\Python34"
-    # - PYTHON: "C:\\Python35"
-    # - PYTHON: "C:\\Python27-x64"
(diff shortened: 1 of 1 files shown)


dwaynebailey:
@nschonni brilliant :)

Been wanting this for a long time.  Looks like the tests are failing in some Windows specific file access issues and in the code to check cli output.

Not sure how to solve the former.  But the later is probably better served by using pytest-catchlog which can capture stderr and stdout.

Artifacts would be good and would ensure we're actually able to package for Windows users.

nschonni:
Circling back to this to see if I can get a reproduction locally.
- Checked out current master
- Installed everything again (bumped lxml to 3.7.3)
- Installed pytest `html` and `catchlog`
- Got the following errors https://gist.github.com/nschonni/685617d1c2e63ad73c183dbe10fc5a56

I did a weird hack because lxml was giving an error around parsing the xliff XSD, but I put that down to proxy issues since it went away after just copying the xml.xsd locally and updating the import

nijel:
I did rebase this on current master and fixed failing tests in https://github.com/translate/translate/pull/3777

nschonni:
Thanks @nijel, I'll close this one and let you run with yours

PR #3575, opened 2017-02-17
Don't output duplicate header

friedelwolff:
I just saw this bug. I think this is the first time I use several of the new developer infrastructure, so please say if there are better ways of doing this.


dwaynebailey:
@friedelwolff lgtm (looks good to merge - as in I checked but didn't run tests myself).  I think the second commit can be rebased into the first IIUC.

Feel free to merge, ping on the channel if you don't have commit rights.

unho:
@friedelwolff Merged. Thanks a lot for your contribution!

PR #3576, opened 2017-02-21
Add setcontext for ts format

phlax:
this allows you to get/set context for a ts unit only when the unit
is not part of a tsfile

it allows pootle to use ttk to derive the correct `id`, but it doesnt
add the required implementation for ts to work when changing a units
context, or inserting a unit that has had its context set

### DIFF
--- a/translate/storage/ts2.py
+++ b/translate/storage/ts2.py
@@ -94,6 +94,7 @@ class tsunit(lisa.LISAunit):
     }
 
     statemap_r = dict((i[1], i[0]) for i in six.iteritems(statemap))
+    _context = None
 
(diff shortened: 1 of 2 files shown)


translate/storage/ts2.py
```diff
@@ -271,12 +272,25 @@ def istranslatable(self):
...
         if context is None:
             return None
         return context.text
 
+    def setcontext(self, value):
+        if value == self.getcontextname():
+            return
+        parent = self.xmlelement.getparent()
+        if parent is None:
+            self._context = context
```
phlax: dammit - this is gonna fail 8/

unho:
lgtm once Travis passes

PR #3577, opened 2017-02-21
Add support for Joomla dialect

nijel:
See https://docs.joomla.org/Specification_of_language_files for specification.


unho:
Landed. Thanks for your contribution!

PR #3582, opened 2017-02-28
Avoid potential Xapian database corruption

### LINKED ISSUE #3581: Don't corrupt the Xapian database
[This code](https://github.com/translate/translate/blob/master/translate/search/indexing/XapianIndexer.py#L377) is a bad idea:

```python
    def _delete_stale_lock(self):
        if not self._writer_is_open():
            lockfile = os.path.join(self.location, 'flintlock')
            if (os.path.exists(lockfile) and
                (time.time() - os.path.getmtime(lockfile)) / 60 > 15):
                logging.warning("Stale lock found in %s, removing.",
                                self.location)
                os.remove(lockfile)
```

Xapian's `flintlock` is not a lock file whose presence indicates that the lock is held, but rather a file which is locked using `fcntl()` locking.  Such locks are automatically released by the OS when the process holding them terminates, so there's no such thing as a "stale lock", and the `flintlock` file should always exist.

If you delete the lockfile and there's an existing writer holding the lock, it'll still believe it holds the lock but now a second writer can now also get the lock - if both writers try to update the database, the result will be database corruption.

I'd strongly recommend removing this machinery entirely.

ojwb:
Xapian's flintlock is not a lock file whose presence indicates that the
lock is held, but rather a file which is locked using fcntl() locking.
Such locks are automatically released by the OS when the process holding
them terminates, so there's no such thing as a "stale lock", and the
flintlock file should always exist.

If you delete the lockfile and there's an existing writer holding the
lock, it'll still believe it holds the lock but now a second writer can
now also get the lock - if both writers try to update the database, the
result will be database corruption.

Fixes #3581.

### DIFF
--- a/translate/search/indexing/XapianIndexer.py
+++ b/translate/search/indexing/XapianIndexer.py
@@ -374,23 +374,23 @@ class XapianDatabase(CommonIndexer.CommonDatabase):
                                (result, fieldnames))
         return result
 
-    def _delete_stale_lock(self):
-        if not self._writer_is_open():
-            lockfile = os.path.join(self.location, 'flintlock')
-            if (os.path.exists(lockfile) and
-                (time.time() - os.path.getmtime(lockfile)) / 60 > 15):
-                logging.warning("Stale lock found in %s, removing.",
-                                self.location)
-                os.remove(lockfile)
-
     def _writer_open(self):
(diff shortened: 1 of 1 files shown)


unho:
Landed. Thanks a lot for your contribution.

PR #3583, opened 2017-03-01
[WIP] Add support iOS XLIFF

unho:
Add support iOS XLIFF


nijel:
I've extracted the state bits from this into https://github.com/translate/translate/pull/6454

PR #3585, opened 2017-03-13
Handle converting back and forth properties with duplicate keys

unho:


translate/convert/test_prop2po.py
```diff
@@ -275,6 +277,18 @@ def test_successive_gaia_plurals(self):
         assert not zero_unit.hasplural()
         assert zero_unit.source == u"Edit"
 
+    @mark.xfail(reason="Still not fixed")
+    def test_duplicate_keys(self):
```
dwaynebailey: Not sure if there are other tests, but I'd add all possible variants to document what we expect to happen:

```properties
key=value
key=another value
```

```properties
key1=value
key2=value
```

unho: k

unho:
Closing in favor of https://github.com/translate/translate/pull/3587

PR #3586, opened 2017-03-13
Duplicate keys

phlax:
handle duplicate keys

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -23,6 +23,7 @@ files (pofile).
 """
 
 import copy
+import logging
 import re
 import six
 import textwrap
@@ -33,6 +34,9 @@ from translate.misc import quote
 from translate.misc.deprecation import deprecated
 from translate.misc.multistring import multistring
 from translate.storage import pocommon, poparser
+
+
+logger = logging.getLogger(__name__)
 
 
 lsep = "\n#: "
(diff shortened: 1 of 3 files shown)


translate/storage/pypo.py
```diff
@@ -791,7 +795,13 @@ def addcomment(thepo):
                         origpo.msgctxt.append('"%s"' % escapeforpo(" ".join(origpo.getlocations())))
                         markedpos.append(thepo)
                     thepo.msgctxt.append('"%s"' % escapeforpo(" ".join(thepo.getlocations())))
-                    uniqueunits.append(thepo)
+                    if not thepo.msgctxt == id_dict[id].msgctxt:
```
phlax: this condition could be improved

translate/storage/pypo.py
```diff
@@ -791,7 +795,13 @@ def addcomment(thepo):
...
+                        uniqueunits.append(thepo)
+                    else:
+                        logger.warn(
+                            "Duplicate unit found with msgctx of '%s' and source '%s'",
```
phlax: this should be msgctxt

dwaynebailey: How about `"Duplicate unit found with msgctxt of '%s' and msgid '%s'"`

unho:
What about cpo implementation?

phlax:
> What about cpo implementation?

added

unho:
Also needs to be added to fpo implementation.

phlax:
closed in favour of #3587

PR #3587, opened 2017-03-13
Po dupes

unho:


translate/storage/html.py
```diff
@@ -326,6 +328,7 @@ def endblock(self):
         self.currentcomment = ""
         self.currenttag = None
         self.currentpos = -1
+        self.currentoffset = 0
```
dwaynebailey: I'm guessing that you might actually want this to be -1 i.e. not defined.

translate/storage/html.py
```diff
@@ -313,6 +314,7 @@ def startblock(self, tag, attrs=None):
         self.currentcomment = ""
         self.currenttag = tag
         self.currentpos = self.getpos()[0]
+        self.currentoffset = self.getpos()[1]
```
dwaynebailey: `self.currentpos, self.currentoffset = self.getpose()[:1]` might be neater, up to you.

translate/storage/html.py
```diff
@@ -176,6 +176,7 @@ def __init__(self, includeuntaggeddata=None, inputfile=None,
         self.currentcomment = u""
         self.currenttag = None
         self.currentpos = -1
+        self.currentoffset = 0
```
dwaynebailey: Ditto here `-1` as in not defined yet.

dwaynebailey:
"Adjustments to match the implementation" ? Needs a better commit message.

dwaynebailey:
I don't see any tests adjusted for html or txt? If the tests are missing I'd recommend adding them now.

unho:
Addressed all raised concerns.

PR #3589, opened 2017-03-15
Keep first unit in store's locationindex if several units have the same location

unho:


translate/storage/base.py
```diff
@@ -675,11 +675,8 @@ def insert_unit(source):
             insert_unit(unit.source)
 
         for location in unit.getlocations():
-            if location in self.locationindex:
-                # if sources aren't unique, don't use them
-                #FIXME: maybe better store a list of units like sourceindex
```
dwaynebailey: There is a reason we set this to None.  The expected results are indeterminate.  But this now changes it to best effort but the commit forgets that this impacts ALL formats and communicates that its po2moz specific.

I'd actually comment with the reason.  Otherwise in a perfect world, you @unho, are going to be made to debug all future brokeness because you failed to caveat this commit.

Please don't remove useful info if you actually aren't fixing things.

I'm wondering if you thought though the option listed in the FIXME? That would possibly allow us to handle the dups in po2prop.  But could have oher bigger fallout.

translate/convert/test_po2moz.py
```diff
@@ -1,8 +1,70 @@
...
+        inputfile = wStringIO.StringIO(posource)
+        templatefile = wStringIO.StringIO(propsource)
+        outputfile = wStringIO.StringIO("")
+        po2prop.convertmozillaprop(inputfile, outputfile, templatefile)
+        print(outputfile.getvalue())
+        return outputfile.getvalue()
+
+    def test_properties_duplicates(self):
+        """Test back conversion of properties with duplicate units."""
+        proptemplate = '''
```
dwaynebailey: This looks like a complete copy over of po2prop.  po2moz is just a wrapper.  Better to fix this properly in po2prop.  It also isn't addressing the test that passed that should have failed.

translate/convert/test_po2prop.py
```diff
@@ -363,6 +363,82 @@ def test_gaia_plurals(self):
...
+'''
+        propfile = self.merge2prop(proptemplate, posource, personality="mozilla")
+        assert propfile == propexpected
+
+        # Test entries with same key and different value.
+        proptemplate = '''
+key1=value
+key2=value
+'''
+        posource = r'''#: key1
```
dwaynebailey: This is wrong, it wouldn't be a valid PO file.  prop2po would have added msgctxt.

translate/storage/base.py
unho: Kept the FIXME and added detailed explanation on commit message.

translate/convert/test_po2prop.py
unho: Changed.

PR #3591, opened 2017-03-16
Add codecov.io support

nijel:
IMHO it is way more usable than coveralls you're using right now. One feature that I like is adding coverage as a status to pull requests, both for patch and overall. I see you're using it for Pootle already, so it might be good idea to have it on translate-toolkit as well.

It needs to be enabled on the codecov.io side as well (and granted access to the GitHub project).

I've kept the coveralls integration in place for now, it can be removed later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,7 @@ matrix:
 #  - sudo apt-get install python-aeidon
 install:
   - pip install coveralls
+  - pip install codecov
   - if [[ $MINVERSION ]]; then make requirements/min-versions.txt; cat requirements/min-versions.txt;  CFLAGS="-O0" pip install -r requirements/min-versions.txt; fi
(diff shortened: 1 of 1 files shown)


unho:
lgtm once tests pass.

Thanks, this is a much welcome contribution. I think we would prefer to keep coveralls as well.

phlax:
@nijel - i use both pretty regularly - originally because they had different killer features - but now days because one or other sometimes blocks/goes down

nijel:
I was using both as well, but I've recently removed coveralls from most of projects as it seemed to be down quite a lot...

dwaynebailey:
I've enabled codecov for the repo. Can't figure out how to turn of comment spam though :(

nijel:
It used to be there (I used to have enabled just the commit status thing), but I can't find it right now as well :-(.

Issue #3578, opened 2017-02-22
Compare TTK plurals against CLDR


unho:
TTK sets plurals data in https://github.com/translate/translate/blob/master/translate/lang/data.py#L29-L202 but some of them might not align with CLDR: http://mlocati.github.io/cldr-to-gettext-plural-rules/

nijel:
The differences can be seen here: https://github.com/WeblateOrg/language-data/blob/master/PLURALS_DIFF.md (but it's comparing more sources and more languages)

Here is reduced table listing only the differences, it also includes default Gettext rules (if they exist) and L10n guide ones:

Code | Name | CLDR plurals | Gettext plurals | L10n guide | Translate toolkit
---- | ---- | ------------ | --------------- | ---------- | -----------------
as | Assamese | nplurals=2; plural=n > 1; |  | nplurals=2; plural=(n != 1); | nplurals=2; plural=(n != 1);
bn | Bengali | nplurals=2; plural=n > 1; |  | nplurals=2; plural=(n != 1); | nplurals=2; plural=(n != 1);
br | Breton | nplurals=5; plural=(n % 10 == 1 && n % 100 != 11 && n % 100 != 71 && n % 100 != 91) ? 0 : ((n % 10 == 2 && n % 100 != 12 && n % 100 != 72 && n % 100 != 92) ? 1 : ((((n % 10 == 3 \|\| n % 10 == 4) \|\| n % 10 == 9) && (n % 100 < 10 \|\| n % 100 > 19) && (n % 100 < 70 \|\| n % 100 > 79) && (n % 100 < 90 \|\| n % 100 > 99)) ? 2 : ((n != 0 && n % 1000000 == 0) ? 3 : 4))); |  | nplurals=2; plural=(n > 1); | nplurals=2; plural=n > 1;
cgg | Chiga | nplurals=2; plural=n != 1; |  | nplurals=1; plural=0; | nplurals=1; plural=0;
cy | Welsh | nplurals=6; plural=(n==0) ? 0 : (n==1) ? 1 : (n==2) ? 2 : (n==3) ? 3 :(n==6) ? 4 : 5; |  | nplurals=4; plural=(n==1) ? 0 : (n==2) ? 1 : (n != 8 && n != 11) ? 2 : 3; | nplurals=2; plural=(n==2) ? 1 : 0;
ff | Fulah | nplurals=2; plural=n > 1; |  | nplurals=2; plural=(n != 1); | nplurals=2; plural=(n != 1);
[...]

PR #3592, opened 2017-03-16
Fix plural forms for Persian, Kazakh and Kirghiz

nijel:
These are the ones which we're fixing up in Weblate, so I think it makes
sense to push them to translate-toolkit. They now also match CLRD, see #3578.


unho:
You might want to also adjust `zu`. Probably there are some other that needs adjusting as well. lgtm anyway.

Thanks for your contribution.

nijel:
There are probably more, right now I'm pushing things we were patching in Weblate.

unho:
lgtm. Any other change can be landed in a future PR.

PR #3593, opened 2017-03-16
Adjusted Slovenian plural equation

nijel:
There has been change in past to make the equation easier to use - the
new one honors ordering. See launchpad for more disussion on the topic:
https://answers.launchpad.net/launchpad/+question/18324

The new formula also matches CLRD, see #3578.

### DIFF
--- a/translate/lang/data.py
+++ b/translate/lang/data.py
@@ -161,7 +161,7 @@ languages = {
     'se': (u'Northern Sami', 2, '(n != 1)'),
     'si': (u'Sinhala; Sinhalese', 2, '(n != 1)'),
     'sk': (u'Slovak', 3, '(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2'),
-    'sl': (u'Slovenian', 4, '(n%100==1 ? 1 : n%100==2 ? 2 : n%100==3 || n%100==4 ? 3 : 0)'),
+    'sl': (u'Slovenian', 4, '(n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 || n%100==4 ? 2 : 3)'),
     'so': (u'Somali', 2, '(n != 1)'),
     'son': (u'Songhai languages', 1, '0'),
     'sq': (u'Albanian', 2, '(n != 1)'),


translate/lang/data.py
```diff
@@ -161,7 +161,7 @@
     'se': (u'Northern Sami', 2, '(n != 1)'),
     'si': (u'Sinhala; Sinhalese', 2, '(n != 1)'),
     'sk': (u'Slovak', 3, '(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2'),
-    'sl': (u'Slovenian', 4, '(n%100==1 ? 1 : n%100==2 ? 2 : n%100==3 || n%100==4 ? 3 : 0)'),
+    'sl': (u'Slovenian', 4, 'n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 || n%100==4 ? 2 : 3'),
```
unho: Please keep the parentheses around the plural formula.

unho:
lgtm, once the commit message typos are fixed and the nit about parentheses addressed.

Just to clearly state, the new proposed formula aligns with CLDR and also makes sense, but what we previously had is what was suggested in the Launchpad issue.

translate/lang/data.py
nijel: Done, however is there some policy on this? The parenthesis are certainly not needed and there doesn't seem to be any consistency in existing equations...

unho: Sadly there is no consistency. Personally I prefer to use them.

But the true reason I raised it in this case is so the PR diff would just reflect the actual changes. I prefer to have cleanups and actual changes on separate commits.

beernarrd:
Please undo this fix. You are running in circles. See:
https://github.com/translate/l10n-guide/issues/22
https://savannah.gnu.org/bugs/?45591

I don't agree that formula in this commit "matches CLRD". CLRD has this order: one, two, few, other whereas gettext has msgid indexes 0, 1, 2, 3. We don't want to put "two" into index 1 and "few" into index 2, especialy because the number 0 happens to be under "other".

We want to have this:
msgid[**0**] = CLRD other {**0**, 5..100, 105..200, ...}
msgid[**1**] = CLRD **one** {**1**, 101, 201, ...}
msgid[**2**] = CLRD **two** {**2**, 102, 202, ...}
msgid[**3**] = CLRD few {**3**,4; 103,104...}

nijel:
Most of the languages do use this ordering even in Gettext - even if the plural index is 0, it is usually used for one. See the discussion at https://answers.launchpad.net/launchpad/+question/18324.

Anyway that's why the plural equation is embedded in the po file and the one here or in gettext should be only used as default one.

I'll followup in more details on https://github.com/translate/l10n-guide/issues/22 as it makes more sense to discuss this there than under merged pull request.

PR #3594, opened 2017-03-16
Remove unused code from CSV

nijel:
This is not being used since cb74f6805ac6cd7e6046938b3d5b2088519b2680, so there is probably no sense to keep it around.


translate/storage/csvl10n.py
```diff
@@ -25,72 +25,9 @@
 import six
 
 from translate.misc import csv_utils
-from translate.misc import sparse
```
unho: Just verified that `translate/misc/sparse.py` can be also dropped. Please include its removal in this PR.

nijel: Good point, I've looked only at SimpleParser class which is used in selector, but it's also defined there...

unho:
lgtm once tests pass.

translate/storage/csvl10n.py
unho: Yes, that is a different `SimpleParser` class.

PR #3599, opened 2017-03-24
Add check for untranslated plural targets.

### LINKED ISSUE #3598: Add critical check for incomplete plural translations
There is a critical check if number of plural forms is correct here
https://github.com/translate/translate/blob/5ab88f78849df8a8eb54f18baa03163189596666/translate/filters/checks.py#L2461-L2477
Though there is no check if all plural forms are filled which can be really useful.

```
#: plurals
msgid "%d pen"
msgid_plural "%d pens"
msgstr[0] "%d ручка"
msgstr[1] ""
msgstr[2] ""
```

unho:
This only reports failure if either:

- There is any plural target translated while there is any other untranslated,
- The unit's plural target count differs from the language's nplural.

The second reason is because because otherwise no error would be reported if
for example a unit has more plural targets than its language's nplurals, but
only has translated exactly the same number of plural targets as the
language's nplurals.

Fixes #3598.


translate/filters/checks.py
```diff
@@ -2476,6 +2476,36 @@ def nplurals(self, unit):
 
         return True
 
+    @critical
+    def untranslated_plurals(self, unit):
+        """Check if any of the plural targets is translated while other isn't.
```
dwaynebailey: Check that all plural forms are provided, OR Check that all plural forms are translated

tests/cli/data/test_pofilter_listfilters/stdout.txt
```diff
@@ -48,6 +48,7 @@ tabs	Checks whether tabs are consistent between the two strings.
 unchanged	Checks whether a translation is basically identical to the original
         string.
 untranslated	Checks whether a string has been translated at all.
+untranslated_plurals	Check if any of the plural targets is translated while other isn't.
```
dwaynebailey: Can we not just extend `untranslated` i.e. do we need a special test at all for this?

translate/filters/test_checks.py
```diff
@@ -1272,6 +1272,155 @@ def test_nplurals():
...
+
+def test_untranslated_plurals_unit_with_plurals():
+    """Test that we detect untranslated plurals.
+
+    Note that this test uses a UnitChecker, not a translation checker.
+    """
+    unit = po.pounit("")
+    unit.source = ["%d file", "%d files"]
+
+    # No nplurals set.
```
dwaynebailey: I think the comment should state what you expect to happen.  IIUC there should be no failures.

translate/filters/checks.py
```diff
@@ -2476,6 +2476,36 @@ def nplurals(self, unit):
...
+            # Report as error if unit has a number of targets different from
+            # its language nplurals.
+            return False
+
+        translated_count = 0
+        for target in unit.target.strings:
+            if len(target) > 0:
+                translated_count += 1
+
+        return translated_count == 0 or translated_count == nplurals
```
dwaynebailey: I'd rather raise SeriousFilterFailure and allow things to be commented such as Nth plural form is missing.

It might be simpler for readers if this is broken up into the two different things that are being checked.  First you're checking that its untranslated in all forms.  Then you're checking that the translated forms are all translated.  Just break that apart and humans will be able to follow it.

translate/filters/checks.py
```diff
@@ -2476,6 +2476,36 @@ def nplurals(self, unit):
...
+
+        nplurals = self.config.lang.nplurals
+
+        if not nplurals > 0:
+            return True
+
+        if len(unit.target.strings) != nplurals:
+            # Report as error if unit has a number of targets different from
+            # its language nplurals.
+            return False
```
dwaynebailey: nplural test already does this IIUC, so no need to duplicate.

dwaynebailey:
Please review comments, otherwise lgtm

PR #3603, opened 2017-03-26
Fix po2ts encoding issues

dwaynebailey:
This fixes an issues with po2ts in which we don't encode the output to UTF-8 when required for writing to file.  This was either fallout from Python 3 changes or lack of Unicode testing in po2ts.

@Artoria2e5 the changes to fix your reported issue are just those in f50aace4e62e37bdc9d122f8b59da4d7b7b5bc68 if you want to patch your local installation of Translate Toolkit


unho:
Reviewed and tested. gtm

PR #3604, opened 2017-03-31
Fix Unicode special chars

dwaynebailey:
* Fixes unmarked Unicode strings using `unicode_literals`
* Bring in `unicode_literals` across `translate/lang`


unho:
lgtm

PR #3606, opened 2017-04-03
Improve newlines handling in Android

nijel:
The original report #3262 was invalid, the format does not work this
way, so this reverts commit a5ffe4c02d0d1b64930baca6512e29665d0d7e42.

Changes:

- \n is the only thing which is parsed as newline
- newline is stored as \n, together with real newline for better readability

See https://github.com/WeblateOrg/weblate/issues/1418 for discussion, most importantly https://github.com/WeblateOrg/weblate/issues/1418#issuecomment-290725588 is statement from original reporter of #3262.

### DIFF
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -34,8 +34,8 @@ from translate.misc.multistring import multistring
 from babel.core import Locale, UnknownLocaleError
 
 EOF = None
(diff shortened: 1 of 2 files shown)


unho:
lgtm. Nice work, can you please land it yourself?

nijel:
Thanks for review, merged.

PR #3607, opened 2017-04-03
Improved behavior on corrupted properties files

nijel:
When the properties file contains only key and lacks value, the
find_delimiter returns None as delimiter and it can be later serialized
as None string into the file. By setting delimiter to empty string, we
keep the file as is.

### DIFF
--- a/translate/convert/test_po2prop.py
+++ b/translate/convert/test_po2prop.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from pytest import mark
+
 from translate.convert import po2prop, test_convert
 from translate.misc import wStringIO
 from translate.storage import po
@@ -58,6 +60,25 @@ class TestPO2Prop(object):
         posource = '''#: prop\nmsgid "value"\nmsgstr "waarde"\n'''
         proptemplate = '''prop  =  value\n'''
         propexpected = '''prop  =  waarde\n'''
+        propfile = self.merge2prop(proptemplate, posource)
+        print(propfile)
+        assert propfile == propexpected
+
+    def test_no_value(self):
+        """check that we can handle keys without value"""
+        posource = '''#: KEY\nmsgctxt "KEY"\nmsgid ""\nmsgstr ""\n'''
+        proptemplate = '''KEY = \n'''
+        propexpected = '''KEY = \n'''
+        propfile = self.merge2prop(proptemplate, posource)
+        print(propfile)
+        assert propfile == propexpected
+
+    @mark.xfail(reason="This doesn't work as expected right now")
+    def test_no_separator(self):
+        """check that we can handle keys without separator"""
+        posource = '''#: KEY\nmsgctxt "KEY"\nmsgid ""\nmsgstr ""\n'''
+        proptemplate = '''KEY\n'''
+        propexpected = '''KEY\n'''
         propfile = self.merge2prop(proptemplate, posource)
         print(propfile)
         assert propfile == propexpected
--- a/translate/convert/test_prop2po.py
+++ b/translate/convert/test_prop2po.py
@@ -60,6 +60,24 @@ class TestProp2PO(object):
         assert pounit.source == "Save file"
         assert pounit.target == ""
 
+    def test_no_value_entry(self):
+        """checks that a properties entry without value is converted"""
+        propsource = 'KEY = \n'
+        pofile = self.prop2po(propsource)
+        pounit = self.singleelement(pofile)
+        assert pounit.getcontext() == "KEY"
+        assert pounit.source == ""
+        assert pounit.target == ""
+
+    def test_no_separator_entry(self):
+        """checks that a properties entry without separator is converted"""
+        propsource = 'KEY\n'
+        pofile = self.prop2po(propsource)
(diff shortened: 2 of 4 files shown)


translate/storage/test_properties.py
```diff
@@ -407,3 +407,12 @@ def test_joomla(self):
         assert propunit.name == u'VALUE'
         assert propunit.source == u'I am a "value"'
         assert bytes(propfile) == propsource
+
+    def test_serialize_missing_value(self):
+        propsource = 'key\n'.encode('utf-8')
```
unho: When I read the PR description I was expecting the "no value" to mean that it was like `key=` but not like `key`. Please add a test for that case too.

unho:
@nijel Besides the missing scenario raised in the added test, this is also lacking roundtrip testing with `prop2po` and `po2prop`.

translate/storage/test_properties.py
nijel: Maybe I should have described it as no delimiter rather :-).

nijel:
I've added the test with missing value as well, but I don't see how this should be tested with prop2po - it does not store the delimiter, so there is no chance this would survive the roundtrip unchanged...

unho:
> I've added the test with missing value as well, but I don't see how this should be tested with prop2po - it does not store the delimiter, so there is no chance this would survive the roundtrip unchanged...

That is exactly what we want to test, a test to check that PO has no-key and no-value and that final properties does have those. I guess that Weblate consumes properties directly and that is the reason why you need this. With po2prop this no-value and no-key should be added back if you provide a template properties with those no-key or no-value.

@nijel please note that **we need a working roundtrip**, and if it doesn't we have no other choice but hold this particular change until this can be done without breaking roundtrip. It is not that we want to impose you a certain way of doing things, but I want to stress the fact that there is people relying on roundtrip to work. I hope you will understand.

nijel:
The roundtrip is certainly not working differently with the patch than before as all this changes is attribute which are not used in the conversion at all. I really don't see what you want to test there.

Anyway I've added two tests which sort of cover this, but still I don't see what you're expecting from them.

translate/convert/test_prop2po.py
```diff
@@ -60,6 +60,15 @@ def test_convertprop(self):
         assert pounit.source == "Save file"
         assert pounit.target == ""
 
+    def test_novalue_entry(self):
```
unho: This is `nokey`. Please add a test for `novalue` too.

unho: I wanted to say this is `nodelimiter`.

translate/convert/test_po2prop.py
```diff
@@ -62,6 +62,15 @@ def test_space_preservation(self):
         print(propfile)
         assert propfile == propexpected
 
+    def test_no_value(self):
```
unho: This is `nodelimiter`. Please add a test for `novalue` too.

translate/convert/test_po2prop.py
```diff
@@ -62,6 +62,15 @@ def test_space_preservation(self):
...
+        """check that we can handle keys without value"""
+        posource = '''#: KEY\nmsgctxt "KEY"\nmsgid ""\nmsgstr ""\n'''
+        proptemplate = '''KEY\n'''
+        propexpected = '''KEY = \n'''
```
unho: `propexpected` is definitely different of `proptemplate`. This is the kind of inconsistency I wanted to add tests for, and it exposes that a change `po2prop` is needed to handle this.

unho: Please check that output is the same as the template, and make the test xfail. I will be landing the needed fix in `po2prop` as it seems unfair to load you with the burden of fixing roundtrip.

unho:
> The roundtrip is certainly not working differently with the patch than before as all this changes is attribute which are not used in the conversion at all. I really don't see what you want to test there.
>
>Anyway I've added two tests which sort of cover this, but still I don't see what you're expecting from them.

Increasing coverage is always a good thing, as this will prevent future messups. Roundtrip is working the same as always, but by introducing this new scenario it exposed a problem with roundtrip not handling correctly this scenario.

Please add the requested changes (add more tests, xfail them, adjust test names) and I will provide a lgtm once done (after another review), and once this PR is landed I will carry on with the require `po2prop` changes.

Thanks for your hard work on this, and sorry for being so picky.

nijel:
Anything more needs to be fixed?

translate/convert/test_po2prop.py
```diff
@@ -62,6 +64,25 @@ def test_space_preservation(self):
...
+        posource = '''#: KEY\nmsgctxt "KEY"\nmsgid ""\nmsgstr ""\n'''
+        proptemplate = '''KEY = \n'''
+        propexpected = '''KEY = \n'''
+        propfile = self.merge2prop(proptemplate, posource)
+        print(propfile)
+        assert propfile == propexpected
+
+    @mark.xfail(reason="This doesn't work as expected right now")
+    def test_no_separator(self):
+        """check that we can handle keys without value"""
```
unho: Typo: *value*

unho:
@nijel lgtm

PR #3611, opened 2017-04-15
Make ResX indent closer to what .NET Resource files look like after editing with Visual Studio

BhaaLseN:
There are a few things that currently happen when roundtripping ResX files that are otherwise modified by Visual Studio:
- The XML Declaration switches from double quotes to single quotes and from lower-case to upper-case encoding
- The inline XML Schema is slightly reformatted (mostly spacing, such as before the `/>` of an empty tag)
- The `<data>` element is incorrectly indented (4 spaces instead of 2)

All those make it rather noisy in the VCS, especially when an automated tool such as [Weblate](https://github.com/WeblateOrg/weblate/) does the committing.

This PR attempts to fix the 3rd bullet by updating the `element.tail` depending on whether a `<comment>` tag is present or not.
However, the majority of the changes ended up in getting the tests to work again, because they currently do a string comparison of two XML structures, which is a pain to use if formatting changes like this are made. It would be a good idea to rework those to use XML/XPath comparisons instead of using a fully formatted and potentially pretty-printed string for both sides; but I simply couldn't get this to happen with my limited python skills.

### DIFF
--- a/translate/convert/test_po2resx.py
+++ b/translate/convert/test_po2resx.py
@@ -98,11 +98,11 @@ class TestPO2RESX(object):
 msgid "Source Text"
 msgstr "Some translated text"'''
         resx_template = self.XMLskeleton % '''<data name="key" xml:space="preserve">
-        <value></value>
-        </data>'''
+    <value></value>
+  </data>'''
         expected_output = self.XMLskeleton % '''<data name="key" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


nijel:
Looks good to me.

unho:
@nijel Thanks for landing this one.

PR #3613, opened 2017-04-18
Add phpfile tests for strings within array()

ta2-1:
This PR adds:
- some tests to `TestPhp2PO`;
- allow to use `'app'->'string1'` location instead of `return+array->'app'->'string1'` for the following example:
```php
<?php
return array(
    'app' => array(
		'string1' => 'Source',
    ),
);
```
It fixes inconsistence how we define locations with phpfile parser and po2php converter.
It's worth to add test to cover this inconsistence specifically.

### DIFF
--- a/translate/convert/test_po2php.py
+++ b/translate/convert/test_po2php.py
@@ -141,6 +141,63 @@ msgstr "stringetjie"
         print(phpfile)
         assert "".join(phpfile) == phpexpected
 
+    def test_named_nested_array(self):
+        """check that we can handle nested arrays"""
+        posource = '''#: $lang->'codes'->'name'\nmsgid "value"\nmsgstr "waarde"\n'''
+        phptemplate = '''$lang = array(\n    'codes' => array(\n        'name' => 'value',\n),\n);\n'''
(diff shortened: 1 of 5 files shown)


translate/storage/php.py
```diff
@@ -267,7 +267,9 @@ def parse(self, phpsrc):
                     equaldel = "=>"
                     enddel = ","
                     inarray = True
-                    prename = line[:line.find('=')].strip() + "->"
+                    eqpos = line.find('=')
+                    if eqpos >= 0:
```
dwaynebailey: `!= -1` may be a better was of indicating that the `eqpos` was found

translate/convert/test_po2php.py
```diff
@@ -141,6 +141,64 @@ def test_arrays(self):
...
+            'list1' => array(
+                'l1' => 'target_l1_1',
+                'list2' => array(
+                    'l1' => 'target_l1_l2_l1',
+                ),
+                'l3' => 'target_l1_3',
+            ),
+            'name2' => 'target2',
+        );\n'''
+        import pdb; pdb.set_trace()
```
unho: This needs to be removed.

unho:
lgtm

PR #3617, opened 2017-04-26
Improve resx skeleton

nijel:
It seems that resheader.reader and resheader.writer are necessary for
Visual Studio to properly load the file.


unho:
lgtm

nijel:
Thanks!

unho:
Thanks for your contribution.

PR #3621, opened 2017-05-10
Fix Travis failures (install Gettext and fix pytest docs URL)

nijel:
Getext does not seem to be installed automatically on Travis now, but
it's used by tests.


unho:
@nijel Great contribution, thanks a lot. Please merge it.

PR #3622, opened 2017-05-11
Subtitles: catch missing subtitle support

dwaynebailey:
Fail and provide a more useful report of missing imports.  This also
breaks out Python 2 vs Python 3 failures as aeidon is only available on
Python 3.


unho:
Merged in e57e738caef483ea448366585d27e46c5563707b

PR #3624, opened 2017-05-19
Handle gracefully empty location tag in QT ts

nijel:
It's not desirable to return None as a location in case the tag is
completely empty.


unho:
@nijel Nice work, lgtm.

PR #3632, opened 2017-06-02
Avoid resolving external entities while parsing XML

nijel:
The lxml allows defining external entities by default and that can lead
to including content of external files using them. This can be really
problematic for services dealing with user uploaded files such as Pootle
or Weblate.

This is least intrusive fix, but maybe it would be better to use [defusedxml](https://pypi.python.org/pypi/defusedxml) instead of lxml directly.

You can find more information about this kind of vulnerability here: https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing

### DIFF
--- a/translate/storage/test_xliff.py
+++ b/translate/storage/test_xliff.py
@@ -416,3 +416,26 @@ class TestXLIFFfile(test_base.TestTranslationStore):
                  </trans-unit>'''
         xlifffile = xliff.xlifffile.parsestring(xlfsource)
         assert xlifffile.units[0].istranslatable()
+
+    def test_entities(self):
+        xlfsource = '''<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE foo [ <!ELEMENT foo ANY > <!ENTITY xxe SYSTEM "file:///etc/passwd" >]>
+<xliff version="1.1" xmlns="urn:oasis:names:tc:xliff:document:1.1">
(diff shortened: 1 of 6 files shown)


nijel:
Full report of the vulnerability against Weblate is [#232614 on HackerOne](https://hackerone.com/reports/232614), it works exactly same in Pootle as well (any authenticated user can upload XLIFF file with such vulnerability).

translate/storage/test_xliff.py
```diff
@@ -416,3 +416,20 @@ def test_parsing(self):
...
         assert xlifffile.units[0].istranslatable()
+
+    def test_entities(self):
+        xlfsource = '''<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE foo [ <!ELEMENT foo ANY > <!ENTITY xxe SYSTEM "file:///etc/passwd" >]>
+<xliff version="1.1" xmlns="urn:oasis:names:tc:xliff:document:1.1">
+        <file original="doc.txt" source-language="en-US">
+                <body>
+                    <trans-unit id="1" xml:space="preserve" translate="yes">
+                        <source>&xxe;</source>
```
unho: Can you expand this test to include a valid entity too?

nijel: Good point, added.

unho:
lgtm. @nijel please land.

PR #3633, opened 2017-06-07
Add Kabyle language

belkacem77:
Adding kabyle entry in the languages dictionary.
Please note that Kabyle hase two codes on CLDR: kab and kab_DZ
We are working on kab code in the CLDR project version 32 (now on submission).
The two codes are also considered on Mozilla Firefox I localized. The priority is for kab.
We used kab as is suggested on iso 639-2 for language codes.
______
This update is for wablate. We are going to localize OpenSuse, so we need this change.

### DIFF
--- a/translate/lang/data.py
+++ b/translate/lang/data.py
@@ -102,6 +102,7 @@ languages = {
     'jbo': ('Lojban', 1, '0'),
     'jv': ('Javanese', 2, '(n != 1)'),
     'ka': ('Georgian', 1, '0'),
+    'kab': ('Kabyle', 1, '(n > 1)'),
     'kk': ('Kazakh', 2, 'n != 1'),
     'kl': ('Greenlandic', 2, '(n != 1)'),
     'km': ('Central Khmer', 1, '0'),


translate/lang/data.py
```diff
@@ -102,6 +102,7 @@
     'jbo': ('Lojban', 1, '0'),
     'jv': ('Javanese', 2, '(n != 1)'),
     'ka': ('Georgian', 1, '0'),
+    'kab': ('Kabyle', 1, '(n > 1)'),
```
unho: This is wrong. Either the number of plurals is greater than `1` or `(n > 1)` is replaced by `0`.

unho: According to what we have in some Pootle deployments this should be:

```python
    'kab': ('Kabyle', 2, '(n != 1)'),
```

but I will defer to you to tell.

belkacem77: Yeah that's right. Sorry. due to copy past.
The right form is as you wrote: 'kab': ('Kabyle', 2, '(n != 1)'),
Shoud I resend an other pull?

claudep:
BTW, you might add some pressure for https://www.sourceware.org/bugzilla/show_bug.cgi?id=18812
It's awfully lengthy to add a locale in glibc :-(

translate/lang/data.py
belkacem77: More information
Name: Kabyle
Code: kab
Plural form: 'kab': ('Kabyle', 2, '(n != 1)'),.
Direction: ltr

dwaynebailey:
Landed in 3d8219d76c07725ee9cb541e636a5f30aee1e97e, many thanks @belkacem77

belkacem77:
thanks for you @dwaynebailey

PR #3634, opened 2017-06-12
Add localedir arg to gettext_lang function

phlax:
as the localedir is gotten from `sys.prefix` this was not working inside virtualenvs

### DIFF
--- a/translate/lang/data.py
+++ b/translate/lang/data.py
@@ -420,7 +420,7 @@ def _fix_language_name(name):
     return name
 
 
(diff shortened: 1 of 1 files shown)


unho:
lgtm

unho:
Alternative approach https://github.com/translate/translate/pull/3636

phlax:
k, but afaict these functions still dont work inside a vm, so they should either be ~fixed or removed i think

phlax:
resolved in other prs

PR #3637, opened 2017-06-15
Strip leading and trailing whitespace from Android resources

nijel:
The leading and trailing whitespace is stripped by Android.

See http://tekeye.uk/android/examples/android-string-resources-gotchas


unho:
@nijel Thanks. Landed myself in b84afb26a06f52a87e7c0d4c6e74b546c67b8e59 since we are going to release today.

PR #3640, opened 2017-06-16
Use language name instead of country name for Montenegrin

nijel:
I think language name is more appropriate here than country name.


unho:
@nijel Merged in 975410ac, thanks.

PR #3642, opened 2017-06-19
Refactor functions for resolving iso639/3166 translation

phlax:
no longer uses module dict for caching as resolving these functions seems to be reasonably fast


unho:
lgtm

PR #3645, opened 2017-06-19
Added a minimal Quality checker.

rimas-kudelis:
This checker only checks for untranslated, unchanged and blank translations, and for repeated words.

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -2312,6 +2312,28 @@ class CCLicenseChecker(StandardChecker):
         StandardChecker.__init__(self, **kwargs)
 
 
+minimalconfig = CheckerConfig()
+
+
+class MinimalChecker(StandardChecker):
+
+    def __init__(self, **kwargs):
+        checkerconfig = kwargs.get("checkerconfig", None)
+
+        if checkerconfig is None:
+            checkerconfig = CheckerConfig()
+            kwargs["checkerconfig"] = checkerconfig
+
+        limitfilters = kwargs.get("limitfilters", None)
+
+        if limitfilters is None:
+            limitfilters = ["untranslated", "unchanged", "blank"]
+            kwargs["limitfilters"] = limitfilters
+
+        checkerconfig.update(minimalconfig)
+        StandardChecker.__init__(self, **kwargs)
+
+
 termconfig = CheckerConfig()
 
 
@@ -2412,6 +2434,7 @@ class IOSChecker(StandardChecker):
 
 
 projectcheckers = {
+    "minimal": MinimalChecker,
     "standard": StandardChecker,
     "openoffice": OpenOfficeChecker,
     "libreoffice": LibreOfficeChecker,
--- a/translate/filters/test_checks.py
+++ b/translate/filters/test_checks.py
@@ -980,6 +980,34 @@ def test_validchars():
     assert passes(stdchecker.validchars, "This sentence contains valid chars", u"\u004c\u032d")
 
 
+def test_minimalchecker():
+    """tests the Minimal quality checker"""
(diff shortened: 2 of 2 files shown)


translate/filters/checks.py
```diff
@@ -2312,6 +2312,28 @@ def __init__(self, **kwargs):
...
+        checkerconfig = kwargs.get("checkerconfig", None)
+
+        if checkerconfig is None:
+            checkerconfig = CheckerConfig()
+            kwargs["checkerconfig"] = checkerconfig
+
+        limitfilters = kwargs.get("limitfilters", None)
+
+        if limitfilters is None:
+            limitfilters = ["untranslated", "unchanged", "blank", "doublewords"]
```
dwaynebailey: I'd leave doublewords out of this one as that's more stylistic and the other ones really are teasing out missing translations.

unho:
Looks OK, other than some coding nits that I prefer not to raise, and the concern raised by @dwaynebailey. It would be great to add some tests for this too.

rimas-kudelis:
I included `doublewords` because I think it's markup-agnostic, and it might be useful even in cases where the source string is vastly different from the target string (like in my case, where source is a slug, and target is the actual string). I would probably even add `doublespace` to the list.

In other words, I just don't think that these tests are prone to too many false positives under any reasonable circumstances, which is why I think that having them wouldn't hurt this checker.

However, I agree that these aren't really absolute baseline, so I'm OK with removing them, if you really think this is for the better.

Regarding tests, would something like this do? This is assuming `doublewords` goes away, but I can easily change it to assert both `doublewords` and `doublespace` are included. I'm not sure if it's worthy to assert the lack of too many `StandardChecker()` tests in `MinimalChecker()`, I guess just a couple would be enough?
```python
def test_minimalchecker():
    """tests the Minimal quality checker"""
    # The minimal checker only checks for untranslated, unchanged and blank strings.
    # All other quality checks should be ignored.
    minimalchecker = checks.MinimalChecker()
    assert fails(minimalchecker.untranslated, "I am untranslated", "")
    assert passes(minimalchecker.untranslated, "I am translated", "Ek is vertaal")
    assert fails(minimalchecker.unchanged, "Unchanged", "Unchanged")
    assert passes(minimalchecker.unchanged, "Unchanged", "Changed")
    assert fails(minimalchecker.blank, "Blank string", " ")

    src, tgt, __ = strprep(u"Save the rhino", u"Save the the rhino")
    unit = base.TranslationUnit(src)
    unit.target = tgt

    # Doublewords check is disabled.
    assert 'doublewords' not in minimalchecker.run_filters(unit).keys()

    src, tgt, __ = strprep(u"Non-matching printf variables", u"Ek is %s")
    unit = base.TranslationUnit(src)
    unit.target = tgt

    # Printf check is disabled.
    assert 'printf' not in minimalchecker.run_filters(unit).keys()
```

dwaynebailey:
@rimas-kudelis double words are valid in some target languages.  I think the goal of a minimal checker should be close to zero false positives.  So I'd say drop.

I think the tests look fine.

rimas-kudelis:
@dwaynebailey But you do have `doublewords` disabled specifically for these target languages, so it wouldn't kick in, isn't that right?

dwaynebailey:
@rimas-kudelis mostly we do, but that doesn't cover all languages.  While the other tests are language agnostic.

unho:
@rimas-kudelis Test looks good, please include in PR. If you do it quick this might make into the next TTK release.

rimas-kudelis:
Okay, I've removed `doublewords` and added the test.

unho:
Merged in 869e1253. @rimas-kudelis Thanks.

PR #3646, opened 2017-06-20
Improve country name handling in lang translations

phlax:
this prevents languages with dialects from showing as `langname (macrolanguage)`

there are other cases where the part of the name in brackets doesnt refer to a country which we might also want to exclude in this way


unho:
lgtm, but would be great to have some macrolanguage test added.

PR #3648, opened 2017-06-20
Reduced tests

rimas-kudelis:
This adds one more checker on top of #3645.

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -2334,6 +2334,28 @@ class MinimalChecker(StandardChecker):
         StandardChecker.__init__(self, **kwargs)
 
 
+reducedconfig = CheckerConfig()
+
+
+class ReducedChecker(StandardChecker):
+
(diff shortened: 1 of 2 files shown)


translate/filters/checks.py
```diff
@@ -2334,6 +2334,28 @@ def __init__(self, **kwargs):
...
+        checkerconfig = kwargs.get("checkerconfig", None)
+
+        if checkerconfig is None:
+            checkerconfig = CheckerConfig()
+            kwargs["checkerconfig"] = checkerconfig
+
+        limitfilters = kwargs.get("limitfilters", None)
+
+        if limitfilters is None:
+            limitfilters = ["untranslated", "unchanged", "blank", "doublespacing", "doublewords", "spellcheck"]
```
rimas-kudelis: I'm pretty sure that startcaps and endcaps woud generate a lot of false positives in my case.

PR #3650, opened 2017-06-20
Checks: remove bidi markers before puncspacing test

### LINKED ISSUE #3635: Directional non printable characters
Since most of my daily translation work is somehow related to Translate Toolkit in some way this bug really bothers me!

Let's pretend I have this source string:
As a customer of (some) Corporation, Product will work for you.

When I come to translate it to Hebrew I need to use RLM characters in order for the text to appear in the correct order:
<div dir="rtl" align="right">
כלקוח של החברה (כלשהי) Corporation,‏ Product יעבוד למענך.</div>
In case you were wondering there's an RLM right after the comma and the word Corporation so the words in English will appear correctly (since comma is agnostic to the directionality of the paragraph it inherits the directionality from the surroundings in this case).

This is the same sentence without directionality characters:
<div dir="rtl" align="right">
כלקוח של החברה (כלשהי) Corporation, Product יעבוד למענך.</div>

In that preticular case I'm seeing an error about changing the order of the punctuation, now that's completely untrue since RLM has nothing to do with punctuation and the order remains.

We're seeing this behaviour mostly in Mozilla Pontoon but there are other cases as well.

dwaynebailey:
Strip out directionality markers before performing the puncspacing test.

Fixes #3635

### DIFF
--- a/translate/filters/checks.py
+++ b/translate/filters/checks.py
@@ -862,7 +862,27 @@ class StandardChecker(TranslationChecker):
             return True
 
         str2 = self.filteraccelerators(self.filtervariables(str2))
+        # Substitute: nbsp
         str2 = str2.replace(u"\u00a0", u" ")
+        # Strip: Bidi markers and ZW* chars
+        str2 = str2.translate(
+            {ord(c): None for c in (
+                # Bidi markers
+                u"\u200e",  # LRM
+                u"\u200f",  # RLM
+                u"\u202b",  # RLE
+                u"\u202a",  # LRE
+                u"\u202e",  # RLO
+                u"\u202d",  # LRO
+                u"\u202c",  # PDF
+                u"\u2069",  # PDI
+                u"\u2068",  # FSI
+                u"\u2067",  # RLI
+                u"\u2066",  # LRI
+                # ZW*
+                u"\u200d",  # ZWJ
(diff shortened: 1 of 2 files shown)


translate/filters/checks.py
```diff
@@ -862,7 +862,13 @@ def puncspacing(self, str1, str2):
             return True
 
         str2 = self.filteraccelerators(self.filtervariables(str2))
+        # Substitute: nbsp
         str2 = str2.replace(u"\u00a0", u" ")
+        # Strip: Bidi markers and ZW* chars
+        str2 = str2.translate(
+            {ord(c): None for c in
+             u"\u200e\u200f\u202b\u202a\u202e\u202d\u202c\u2069\u2068\u2067\u2066" +
+             u"\u200d\u200c"})
```
unho: Nit: Instead of a string I would have used a tuple of characters, one character per line with a comment besides with the name for the character, like in https://github.com/translate/translate/blob/master/translate/lang/ar.py#L51-L62

unho:
Other than raised nit, lgtm.

translate/filters/checks.py
unho: Note that I don't mean having a tuple of tuples (replacement, source) like in the linked example, but just a plain tuple of strings, since all are replaced by `None`.

PR #3653, opened 2017-06-22
Added reduced checker to the projectcheckers list

rimas-kudelis:
This should have been part of #3648, but I forgot.


unho:
Merged. Thanks @rimas-kudelis, we will try to get a bugfix release as soon as possible.

PR #3655, opened 2017-06-27
Skip obsolete units in po2txt

lnussel:


lnussel:
not sure about your test suite here. A better fix might be to make istranslated() not include obsolete strings. I needed a quick fix for an openSUSE release though. I cannot follow up here.

dwaynebailey:
`istranslatable` should exclude obsolete and header.

dwaynebailey:
@lnussel thanks for the fix, I'll land this with some slight changes and a test as a different PR.

dwaynebailey:
Closing, prefer #3657 fix

PR #3656, opened 2017-06-28
Use common_name for lang/country if available

dwaynebailey:
Ensure we get 'Taiwan' not 'Taiwan, Province of China'.  Which impacts which translations we'll pick up.


unho:
lgtm

PR #3657, opened 2017-06-28
Don't use obsolete messages in po2txt

dwaynebailey:
This is a slightly different fix from that proposed in #3655 and includes a test.

### DIFF
--- a/translate/convert/test_po2txt.py
+++ b/translate/convert/test_po2txt.py
@@ -61,6 +61,22 @@ msgstr "Lyfteks"
         assert self.po2txt(posource) == "Opskrif\n\nLyfteks"
         assert self.po2txt(posource, txttemplate) == "Opskrif\n\nLyfteks"
 
+    def test_header_ignore(self):
+        """check that we ignore headers"""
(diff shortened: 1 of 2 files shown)


translate/convert/test_po2txt.py
```diff
@@ -45,6 +45,22 @@ def test_fuzzy_handling(self):
         assert self.po2txt(posource) == "Heading\n\nLyfteks"
         assert self.po2txt(posource, txttemplate) == "Heading\n\nLyfteks"
 
+    def test_obsolete_ignore(self):
+        """check that we handle fuzzy message correctly"""
+        txttemplate = "Heading\n\nBody text"
+        posource = '''
+msgid "Heading"
```
unho: Since the `isheader` check was dropped, I would explicitly add here a header.

unho:
lgtm after raised concern.

translate/convert/test_po2txt.py
dwaynebailey: Added its own test didn't want to overload a test that explicitly looking at obsolete units.

PR #3663, opened 2017-07-03
Add function to retrieve all languages

unho:
This PR is to address https://github.com/translate/pootle/pull/6542/files#r125042772

### DIFF
--- a/translate/lang/factory.py
+++ b/translate/lang/factory.py
@@ -21,6 +21,8 @@
 
 from __future__ import unicode_literals
 
+import pkgutil
+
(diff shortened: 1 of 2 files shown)


translate/lang/factory.py
```diff
@@ -54,3 +54,24 @@ def getlanguage(code):
             return relatedlanguage
         else:
             return common.Common(code)
+
+
+def get_all_languages():
+    """Return all language classes."""
+    import pkgutil
```
phlax: import at module level

PR #3665, opened 2017-07-19
Add basic support for WebExtension JSON

nijel:
See following URLs for format docs:

* https://developer.chrome.com/extensions/i18n
* https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -246,3 +246,60 @@ class JsonNestedFile(JsonFile):
         out.write(json.dumps(units, separators=(',', ': '),
                              indent=4, ensure_ascii=False).encode(self.encoding))
         out.write(b'\n')
+
+
+class WebExtensionJsonUnit(base.TranslationUnit):
+    def __init__(self, source=None, ref=None, item=None):
+        if ref:
+            self._node = ref
+            self.notes = ref['description']
+            self._target = ref['message']
+        else:
+            self._node = OrderedDict((('message', ''), ('description', '')))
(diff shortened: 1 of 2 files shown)


translate/storage/jsonl10n.py
```diff
@@ -246,3 +246,56 @@ def nested_set(target, path, value):
...
+        self._id = value
+
+    def getid(self):
+        return self._id
+
+    def getlocations(self):
+        return [self.getid()]
+
+    def settarget(self, target):
+        super(WebExtensionJsonUnit, self).settarget(target)
```
unho: Is it actually necessary to call super here? I mean, defining `gettarget` instead looks more explicit to me.

translate/storage/jsonl10n.py
```diff
@@ -246,3 +246,56 @@ def nested_set(target, path, value):
...
+        self._node['message'] = target
+
+    def addnote(self, text, origin=None, position="append"):
+        super(WebExtensionJsonUnit, self).addnote(text, origin, position)
```
unho: Instead of calling super I would define `removenotes` to refer to `self._node['description']` directly.

nijel: I didn't want to copy much of the code from the base class, that's why I call super. Another option is to reimplement all logic behind target/notes handling and directly manipulate with inside dict instead of using properties defined by TranslationUnit...

unho:
@nijel lgtm, please land it.

nijel:
Thanks for review!

PR #3667, opened 2017-07-31
Releasing 2.2.4

unho:


docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
...
+
+*Released on 31 July 2017*
+
+This release contains improvements and bug fixes.
+
+
+Major changes
+=============
+
+- Full support for `.xliff` extension.
```
dwaynebailey: Support for .xliff extension

Full support sounds like we're refering the XLIFF itself.

unho: Will change.

docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
...
+- Added support for nested and WebExtension JSON dialects.
+- `po2txt` no longer converts non-translatable strings.
```
dwaynebailey: I think what that fixes actually does is use en if translation does not exist.

docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
+Translate Toolkit 2.2.4
+***********************
+
+*Released on 31 July 2017*
+
+This release contains improvements and bug fixes.
+
+
+Major changes
```
dwaynebailey: I think previous notes dropped Major.  These are not major changes.  They're just 'Summary of Changes'

docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
...
+    extension.
+
+- JSON
+
+  - Added support for nested JSON.
+  - Added support for WebExtension JSON dialect.
+
+- txt
+
+  - `po2txt` skips obsolete and non-translatable strings.
```
dwaynebailey: Again if asked to skip them it uses en instead fo blank.  At least that's what I think my change did :)

docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
...
+------------------
+
+- The `puncspace` check now strips Bidi markers chars before processing.
+- Added `ReducedChecker` checker to list of checkers.
+
+
+Tools
+-----
+
+- Added support for `.xliff` extension in all tools that support `.xlf` one.
```
dwaynebailey: This has appeared now 3 times, put it in one place only in the details.

docs/releases/2.2.4.rst
```diff
@@ -0,0 +1,66 @@
...
+Tools
+-----
+
+- Added support for `.xliff` extension in all tools that support `.xlf` one. 
+
+
+API changes
+-----------
+
+- Language and country default to `common_name` if available.
```
dwaynebailey: I'm not sure this is an API change, maybe you disagree

dwaynebailey:
Quite a few nits, otherwise gtm

docs/releases/2.2.4.rst
unho: I see no better place to put this change. I will leave it here.

docs/releases/2.2.4.rst
unho: No, my description is correct according to the commit messages, the code changes and the tests you landed. In other words, `po2txt` now correctly skips header and obsolete units from PO files when converting back to txt format.

PR #3669, opened 2017-08-07
xliff: add some settarget adjustmnts to handle corner case state changes

dwaynebailey:
Addresses 3 corner case state change scenarios:

* Untranslated -> Fuzzy but translated
* Translated -> Untranslated
* Fuzzy -> Untranslated


translate/storage/xliff.py
```diff
@@ -124,6 +124,18 @@ class xliffunit(lisa.LISAunit):
         S_SIGNED_OFF: (state.FINAL, state.MAX),
     }
 
+    def settarget(self, text, lang='xx', append=False):
+        if not self.istranslated and text:
```
phlax: `self.istranslated()`

translate/storage/xliff.py
```diff
@@ -124,6 +124,18 @@ class xliffunit(lisa.LISAunit):
         S_SIGNED_OFF: (state.FINAL, state.MAX),
     }
 
+    def settarget(self, text, lang='xx', append=False):
+        if not self.istranslated() and text:
+            self.markfuzzy(True)
```
nijel: Why setting target should mark the unit fuzzy? IMHO the fuzzy state is quite confusing in xliffunit and this makes it just worse. Maybe `self.markapproved(False)` would be more appropriate in this case.

Actually it might make more sense to always do `self.markapproved(False)` on settarget...

Issue #3424, opened 2016-04-08
Fix indent in XLIFF output


mathjazz:
XLIFF (and maybe other LISA files) aren't indented properly on output.

They look like this:

``` xml
<trans-unit id="ID" approved="yes">
  <source>Source Text</source>
  <target state="translated">Translation</target> <note>Note</note>
</trans-unit>
```

Instead of this:

``` xml
<trans-unit id="ID" approved="yes">
  <source>Source Text</source>
  <target state="translated">Translation</target>
  <note>Note</note>
</trans-unit>
```

Fix should be pretty simple, but with my level of translate-toolkit experience, I'm afraid of submitting a pull request: https://github.com/mathjazz/translate/commit/29afa0d4fea13cb0c310ce251c91fb3140184d32.

Got the idea for the fix from @flodolo:
https://github.com/mozilla/firefox-ios-build-tools/blob/master/scripts/update-xliff.py#L26

claudep:
Is that from the latest beta version?
I am not able to reproduce. By applying this diff:
```
diff --git a/translate/storage/test_xliff.py b/translate/storage/test_xliff.py
index b2d7150..9d4364d 100644
--- a/translate/storage/test_xliff.py
+++ b/translate/storage/test_xliff.py
@@ -198,6 +198,7 @@ class TestXLIFFfile(test_base.TestTranslationStore):
     def test_notes(self):
         xlifffile = xliff.xlifffile()
         unit = xlifffile.addsourceunit("Concept")
+        unit.target = "tpecnoC"
         # We don't want to add unnecessary notes
         assert "note" not in bytes(xlifffile).decode('utf-8')
         unit.addnote(None)
@@ -209,6 +210,7 @@ class TestXLIFFfile(test_base.TestTranslationStore):
         assert unit.getnotes() == "Please buy bread"
         notenodes = unit.xmlelement.findall(".//%s" % unit.namespaced("note"))
         assert len(notenodes) == 1
+        assert bytes(xlifffile).decode('utf-8') == '?'
 
         unit.addnote("Please buy milk", origin="Mom")
         notenodes = unit.xmlelement.findall(".//%s" % unit.namespaced("note"))
```
I'm getting the following while running tests:
```
>       assert bytes(xlifffile).decode('utf-8') == '?'
E       assert '<?xml versio...le>\n</xliff>\n' == '?'
E         + ?
E         - <?xml version='1.0' encoding='UTF-8'?>
[...]

PR #3670, opened 2017-08-17
Fix issue 3424 - Fix indent in XLIFF output

mastizada:


mastizada:
Tested in pontoon project, in {file} item `<trans-unit id="Add to Bookmarks">`: `note` element was inlined with `target` element, after adding these changes indent solved.
`file`: https://raw.githubusercontent.com/mastizada/firefoxios-l10n/master/az/firefox-ios.xliff

translate/storage/lisa.py
```diff
@@ -302,8 +302,24 @@ def addunit(self, unit, new=True):
         if new:
             self.body.append(unit.xmlelement)
 
+    def cleanindent(self, elem, level=0):
+        i = "\n" + level*"  "
+        if len(elem):
+            if not elem.text or not elem.text.strip():
+                elem.text = i + "  "
```
unho: Altering the text in the elements might have unintended consequences.

unho:
@mastizada I have several concerns about this PR:

- Lack of tests
- It changes not only XLIFF but other XML formats
- It changes the indent for all the tags while it looks like the idea is to just fix indent for `note` tag
- The indent amount is hardcoded

Can you please address these?

BhaaLseN:
Thats basically what I was trying to achieve with #3611, but it still fails in some cases (at least I randomly get misaligned text in our VCS after Weblate touches the files).

However, I _do_ want to point out that this could be useful for all XML formats (not just XLIFF). ResX might benefit from this as well (especially if we can get rid of the terrible inline `tail` tracking/modification), plus a generic flat XML format I'm currently working on (where I'm basically stuffing a method similar to this one in before running `serialize`, albeit mine is called `reindent` instead of `cleanindent`)

A good place for such a method could be `translate.misc.xml_helpers` to decouple it from the actual XML formats.

nijel:
I've tried to write something more generic based on this in https://github.com/translate/translate/pull/3871

unho:
Closing this because issue https://github.com/translate/translate/issues/3424 was fixed by PR https://github.com/translate/translate/pull/3871

@mastizada Thanks for your effort anyway!

mastizada:
Thanks @unho , any ideas when it will be published? (current version is 2.3.1)

unho:
@mastizada I would like some other PRs to be merged first. After that we might do a release. I would personally like a release to be done in like a month, but no promises.

Issue #3672, opened 2017-08-18
po2resx does not work on input with unescaped characters


andreistefan87:
When url encoding locations that are added to "sourcecomments" in a po unit, only the space character is considered. Nevertheless, when building the locations index, url decoding always happens.
Due to this lookups in the location index done for resx units that have unescaped characters in their ids fail, with the consequence of losing existing translations, due to target strings being replaced by their source counterparts.
e.g "DLG_WORKLOW+101" is stored with the + sign unescaped in the po file and when building the locations index we get "DLG_WORKLOW 101" which will get compared with the resx unit id "DLG_WORKLOW+101"

alemangui:
Pr #3673 addresses this issue

unho:
@alemangui @andreistefan87 It would be great if you can contribute a test to ensure that the proposed fixes do work. Or at least share a RESX snippet that someone can use to code that test.

unho:
The fix for this has been landed. But will keep this open because the test is necessary.

PR #3673, opened 2017-08-18
Fix for #3672: Always URL-encode keys

alemangui:
Index lookups always decode them
Spaces are not the only characters requiring escaping (at least '+' needs it)


unho:
The change looks ok. But it makes several tests to fail. Can you please adjust those tests too?

nijel:
It seems that this breaks html2po conversion (the getlocations returns encoded string there), can you please look into that?

Also the urlencoding is way too aggressive here, certainly at least some chars do not have to be urlencoded (looking at the tests at least ```()/:```), I've tried to address this in https://github.com/translate/translate/pull/3676.

alemangui:
@andreistefan87 @raphael-wb thoughts?

unho:
@nijel I don't think #3676 fully fixes the issue outlined in this PR description.

nijel:
I don't think as well, it addresses just the too much of escaping this PR produces for most of the cases. Still this is IMHO valid patch (once the tests pass).

alemangui:
Here are a couple commits to update the PR to latest master and fix the remaining tests. 

FYI @andreistefan87 @raphael-wb

unho:
Landed the changes in this PR. Thanks for the contribution.

PR #3674, opened 2017-08-24
Correctly handle webextension without description

nijel:
Apparently the description is optional here, so let's not fail if it is not present.

### DIFF
--- a/translate/storage/test_jsonl10n.py
+++ b/translate/storage/test_jsonl10n.py
@@ -70,6 +70,14 @@ class TestWebExtensionStore(test_monolingual.TestMonolingualUnit):
 
         assert out.getvalue() == b'{\n    "key": {\n        "message": "value",\n        "description": "note"\n    }\n}\n'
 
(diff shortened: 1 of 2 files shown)


translate/storage/test_jsonl10n.py
```diff
@@ -70,6 +70,14 @@ def test_serialize(self):
 
         assert out.getvalue() == b'{\n    "key": {\n        "message": "value",\n        "description": "note"\n    }\n}\n'
 
+    def test_serialize_no_description(self):
+        store = self.StoreClass()
+        store.parse('{"key": {"message": "value"}}')
+        out = BytesIO()
+        src = store.serialize(out)
+
+        assert out.getvalue() == b'{\n    "key": {\n        "message": "value"\n    }\n}\n'
```
unho: Nit: I would avoid having `\n` and instead use string in triple quotes to make it more readable.

unho:
lgtm once Travis passes. The nit is not a blocker.

PR #3676, opened 2017-08-24
Avoid escaping some chars in po files

nijel:
These really do not have to be escaped in the po file, so keep them as
they are.

The list probably could be extended, but this is what I've spotted while reviewing https://github.com/translate/translate/pull/3673


unho:
Please add testing for this change.

nijel:
Good point, added.

unho:
It is still missing testing for `()` characters. Once added, lgtm.

PR #3678, opened 2017-09-01
Add support for i18next format with plurals

nijel:
This is basically JSON with additional semantics to handle plurals.

I've done some refactoring in JSON format to avoid too much code duplication.

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -81,12 +81,8 @@ class JsonUnit(base.TranslationUnit):
 
     def __init__(self, source=None, ref=None, item=None, **kwargs):
         self._id = None
-        self._item = str(os.urandom(30))
-        if item is not None:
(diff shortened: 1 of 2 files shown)


translate/storage/jsonl10n.py
```diff
@@ -81,12 +81,14 @@ class JsonUnit(base.TranslationUnit):
...
-        self._item = str(os.urandom(30))
         if item is not None:
             self._item = item
-        self._ref = {}
+        else:
+            self._item = str(os.urandom(30))
         if ref is not None:
             self._ref = ref
+        else:
+            self._ref = {}
```
unho: I would make this shorter:

```python
        self._item = str(os.urandom(30)) if item is None else item
        self._ref = {} if ref is None else ref
```

unho:
lgtm provided Travis passes.

PR #3683, opened 2017-09-01
Add setcontext to mo file

### LINKED ISSUE #3441: setcontext is not working correctly for mounit
Calling setcontext on mounit does currently nothing as it inherits code from base class:

``` python
    def setcontext(self, context):
        """Set the message context"""
        pass
```

I'd expect it to properly update context as it does for other storages.

nijel:
To make the API consistent.

Fixes #3441

### DIFF
--- a/translate/storage/test_mo.py
+++ b/translate/storage/test_mo.py
@@ -9,6 +9,11 @@ from translate.storage import factory, mo, test_base
 class TestMOUnit(test_base.TestTranslationUnit):
     UnitClass = mo.mounit
 
+    def test_context(self):
+        unit = self.UnitClass("Message")
+        unit.setcontext('context')
+        assert unit.getcontext() == 'context'
+
 
 posources = [
     r'''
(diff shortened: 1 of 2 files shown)


translate/storage/test_mo.py
```diff
@@ -124,6 +128,14 @@ def test_language(self):
         store.updateheader(add=True, Language="zu")
         assert store.gettargetlanguage() == "zu"
 
+    def test_context(self):
+        store = self.StoreClass()
+        unit = mo.mounit('source')
```
unho: Nit: Use `store.UnitClass` instead of `mo.mounit`.

unho:
Other than raised nit and styling issue picked up by Travis, lgtm.

translate/storage/test_mo.py
nijel: `AttributeError: 'TestMOFile' object has no attribute 'UnitClass'`

Should I add it or keep using mo.mounit?

nijel: In the end I've chosen to use `self.StoreClass.UnitClass`...

PR #3685, opened 2017-09-03
Simplify handling of JSON files

nijel:
- share most of the code among different variants
- avoid storing internal representation, what did lead to storing whole
  JSON dictionary for every unit on flat files
- the `__str__` for unit now returns something reasonable

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -80,15 +80,21 @@ from translate.storage import base
 class JsonUnit(base.TranslationUnit):
     """A JSON entry"""
 
-    def __init__(self, source=None, ref=None, item=None, **kwargs):
+    def __init__(self, source=None, item=None, notes=None, **kwargs):
         identifier = str(uuid.uuid4())
+        # Global identifier across file
         self._id = '.' + identifier
+        # Identifier at this level
         self._item = identifier if item is None else item
-        self._ref = OrderedDict() if ref is None else ref
-        if ref is None and item is None:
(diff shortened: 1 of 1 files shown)


translate/storage/jsonl10n.py
```diff
@@ -322,43 +255,22 @@ class WebExtensionJsonFile(JsonFile):
...
-        units = OrderedDict()
-        for unit in self.unit_iter():
-            units[unit.getid()] = unit._node
-        return units
+            value = data[item]
+            yield (
+                item,
+                value['message'] if 'message' in value else '',
+                item,
+                value['description'] if 'description' in value else '',
```
unho: Nit: Use instead (do similar above):

```python
value.get('description', '')
```

translate/storage/jsonl10n.py
```diff
@@ -322,43 +255,22 @@ class WebExtensionJsonFile(JsonFile):
...
                                name_last_node=None, last_node=None):
         for item in data:
-            yield (item, item, data[item], item)
-
-    def serialize_units(self):
-        units = OrderedDict()
-        for unit in self.unit_iter():
-            units[unit.getid()] = unit._node
-        return units
+            value = data[item]
```
unho: Is it not possible to do:

```python
        for item, value in six.iteritems(data):
```

instead of:

```python
        for item in data:
            value = data[item]
```

unho:
lgtm after nits.

nijel:
Thanks for suggestions, merging.

PR #3696, opened 2017-09-27
PHP testsuite fixes and improvements

nijel:
While working with PHP parser, I've noticed some minor issues with it, mostly invalid PHP code used in tests or generated by tests.

### DIFF
--- a/translate/convert/php2po.py
+++ b/translate/convert/php2po.py
@@ -71,14 +71,10 @@ class php2po(object):
                                                         "translator")
             else:
                 translatedoutputunit = None
-            # If we have a valid po unit, get the translation and add it.
-            if outputunit is not None:
-                if translatedoutputunit is not None and not blankmsgstr:
-                    outputunit.target = translatedoutputunit.source
-                outputstore.addunit(outputunit)
-            elif translatedoutputunit is not None:
-                logger("error converting original properties definition %s",
-                       templateunit.name)
+            # We have a valid po unit, get the translation and add it.
+            if translatedoutputunit is not None and not blankmsgstr:
+                outputunit.target = translatedoutputunit.source
+            outputstore.addunit(outputunit)
         outputstore.removeduplicates(duplicatestyle)
         return outputstore
 
--- a/translate/convert/test_php2po.py
+++ b/translate/convert/test_php2po.py
@@ -131,7 +131,7 @@ $lang['prefPanel-smime'] = 'Security';'''
     def test_spaces_in_name(self):
         """checks that if we have spaces in the name we create a good PO with no spaces"""
         phptemplate = '''$lang[ 'credit' ] = 'Something';'''
-        phpsource = '''$lang[ 'credit' ] = ''n Ding';'''
+        phpsource = '''$lang[ 'credit' ] = 'n Ding';'''
         pofile = self.php2po(phpsource, phptemplate)
         pounit = self.singleelement(pofile)
         assert pounit.getlocations() == ["$lang[ 'credit' ]"]
--- a/translate/convert/test_php2po.py
+++ b/translate/convert/test_php2po.py
@@ -20,12 +20,14 @@ class TestPhp2PO(object):
             outputpo = convertor.convertstore(inputphp)
         return outputpo
 
-    def convertphp(self, phpsource):
+    def convertphp(self, phpsource, template=None, expected=1):
         """call the convertphp, return the outputfile"""
         inputfile = wStringIO.StringIO(phpsource)
         outputfile = wStringIO.StringIO()
         templatefile = None
-        assert php2po.convertphp(inputfile, outputfile, templatefile)
+        if template:
+            templatefile = wStringIO.StringIO(template)
+        assert php2po.convertphp(inputfile, outputfile, templatefile) == expected
         return outputfile.getvalue()
 
     def singleelement(self, pofile):
@@ -57,6 +59,34 @@ class TestPhp2PO(object):
         pounit = self.singleelement(pofile)
         assert pounit.source == "entry"
         assert pounit.target == ""
+
+    def test_convertphptemplate(self):
+        """checks that the convertphp function is working with template"""
+        phpsource = """$_LANG['simple'] = 'entry';"""
+        phptemplate = '''$_LANG['simple'] = 'source';'''
+        posource = self.convertphp(phpsource, phptemplate)
+        pofile = po.pofile(wStringIO.StringIO(posource))
+        pounit = self.singleelement(pofile)
+        assert pounit.source == "source"
+        assert pounit.target == "entry"
+
+    def test_convertphpmissing(self):
+        """checks that the convertphp function is working with missing key"""
+        phpsource = """$_LANG['simple'] = 'entry';"""
+        phptemplate = '''$_LANG['missing'] = 'source';'''
+        posource = self.convertphp(phpsource, phptemplate)
+        pofile = po.pofile(wStringIO.StringIO(posource))
+        pounit = self.singleelement(pofile)
+        assert pounit.source == "source"
+        assert pounit.target == ""
+
+    def test_convertphpempty(self):
+        """checks that the convertphp function is working with empty template"""
+        phpsource = ''
+        phptemplate = ''
+        posource = self.convertphp(phpsource, phptemplate, 0)
+        pofile = po.pofile(wStringIO.StringIO(posource))
+        assert len(pofile.units) == 0
 
     def test_unicode(self):
         """checks that unicode entries convert properly"""
--- a/translate/convert/test_po2php.py
+++ b/translate/convert/test_po2php.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from pytest import mark
+from pytest import mark, raises
 
 from translate.convert import po2php, test_convert
 from translate.misc import wStringIO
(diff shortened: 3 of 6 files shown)


translate/storage/test_php.py
```diff
@@ -325,7 +343,7 @@ def test_parsing_define_entries_with_quotes(self):
         """Parse define syntax for entries with quotes"""
         phpsource = """define('_SETTINGS_COOKIEPREFIX', 'Prefixo da "cookie"');
 define('_YOUR_USERNAME', 'O seu nome de usuario: "cookie"');
-define("_REGISTER", "Register <a href=\"register.php\">here</a>");"""
+define("_REGISTER", "Register <a href=\\"register.php\\">here</a>");"""
```
unho: This was correctly escaped before.

translate/storage/php.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, source=""):
         """Construct a blank phpunit."""
         self.escape_type = None
         super(phpunit, self).__init__(source)
-        self.name = ""
+        self.name = "$text"
```
unho: I would use something that clearly states that this is a placeholder, like `$TTK_PLACEHOLDER` or similar.

unho:
I have raised some concerns/comments.

translate/storage/test_php.py
nijel: No - that there is escaping for Python and then for PHP to have quotes inside quotes.

The original code produces:

```php
define("_REGISTER", "Register <a href="register.php">here</a>");
```
What is not valid PHP code.

translate/storage/php.py
nijel: Good point, will change this.

translate/storage/test_php.py
unho: Then it looks like the problem is in the original code, not in the test.

nijel: Yes, the original PHP parser is broken in many ways, in this case it happily parsed invalid PHP code.

translate/convert/test_po2php.py
```diff
@@ -28,6 +28,19 @@ def merge2php(self, phpsource, posource):
...
+        posource = '''#: $lang['name']\nmsgid "value"\nmsgstr "waarde"\n'''
+        phptemplate = '''$lang['name'] = 'value';\n'''
+        phpexpected = b'''$lang['name'] = 'waarde';\n'''
+        inputfile = wStringIO.StringIO(posource)
+        templatefile = wStringIO.StringIO(phptemplate)
+        outputfile = wStringIO.StringIO()
+        assert po2php.convertphp(inputfile, outputfile, templatefile) == 1
+        assert outputfile.getvalue() == phpexpected
+        # Try without template
+        assert raises(ValueError, po2php.convertphp, inputfile, outputfile, None)
```
unho: I find it preferable to use:

```python
        with pytest.raises(ValueError):
            po2php.convertphp(inputfile, outputfile, None)
```

translate/convert/test_po2php.py
```diff
@@ -46,6 +59,38 @@ def test_space_preservation(self):
         print(phpfile)
         assert phpfile == [phpexpected]
 
+    def test_unused_preservation(self):
+        """check that we preserve any unused statements in php files when merging"""
+        posource = '''#: $lang['name']\nmsgid "value"\nmsgstr "waarde"\n'''
```
unho: This is easier to read if you get rid of the `\n`. After all this is a multiline string.

translate/convert/test_php2po.py
```diff
@@ -58,6 +60,34 @@ def test_convertphp(self):
...
+
+    def test_convertphpmissing(self):
+        """checks that the convertphp function is working with missing key"""
+        phpsource = """$_LANG['simple'] = 'entry';"""
+        phptemplate = '''$_LANG['missing'] = 'source';'''
+        posource = self.convertphp(phpsource, phptemplate)
+        pofile = po.pofile(wStringIO.StringIO(posource))
+        pounit = self.singleelement(pofile)
+        assert pounit.source == "source"
+        assert pounit.target == ""
```
unho: I would also assert the number of units in `pofile`.

translate/convert/test_po2php.py
nijel: Well this is how rest of the test_po2php.py looks like, but you're right that without \n it looks better.

translate/convert/test_php2po.py
```diff
@@ -58,6 +60,34 @@ def test_convertphp(self):
...
+
+    def test_convertphpempty(self):
+        """checks that the convertphp function is working with empty template"""
+        phpsource = """"""
+        phptemplate = ''''''
```
unho: Nit: In this case I would just go for the simplest way to specify a blank string: `""`

translate/convert/test_php2po.py
nijel: singleelement already does that

translate/convert/php2po.py
```diff
@@ -71,14 +71,10 @@ def mergestore(self, templatestore, inputstore, blankmsgstr=False,
...
                 translatedoutputunit = None
-            # If we have a valid po unit, get the translation and add it.
-            if outputunit is not None:
-                if translatedoutputunit is not None and not blankmsgstr:
-                    outputunit.target = translatedoutputunit.source
-                outputstore.addunit(outputunit)
-            elif translatedoutputunit is not None:
-                logger("error converting original properties definition %s",
-                       templateunit.name)
+            # We have a valid po unit, get the translation and add it.
```
unho: Nice cleanup!

translate/convert/test_php2po.py
unho: That's right. Sorry for the noise.

unho:
lgtm

translate/storage/test_php.py
unho: I still don't think this fix in the test is the right way to do this.

translate/convert/test_po2php.py
unho: We have a lot of code follows patterns that is better to avoid, so better lets start adding code following nicer patterns even if we don't refactor all legacy code.

translate/storage/test_php.py
nijel: Okay the current situation is:

- The test contains invalid PHP code.
- The current php parser happily parses that.

Apparently I'm not going to fix broken parser to thrown error on this when there is  #3697. I'm open to other ways to fix this but I really don't see how :-).

translate/convert/test_po2php.py
nijel: Once this is done, I'll probably submit some cleanup PRs as well - it cases like this it's pretty easy to start by copying other testcases and that should lead to acceptable PR...

translate/convert/test_po2php.py
```diff
@@ -28,6 +28,47 @@ def merge2php(self, phpsource, posource):
...
+        assert po2php.convertphp(inputfile, outputfile, templatefile) == 1
+        assert outputfile.getvalue() == phpexpected
+
+    def test_convertphp_notemplate(self):
+        """test convertphp helper without template"""
+        posource = '''#: $lang['name']
+msgid "value"
+msgstr "waarde"
+'''
+        phpexpected = b'''$lang['name'] = 'waarde';
```
unho: This variable is unused.

translate/convert/test_po2php.py
```diff
@@ -28,6 +28,47 @@ def merge2php(self, phpsource, posource):
...
+'''
+        inputfile = wStringIO.StringIO(posource)
+        outputfile = wStringIO.StringIO()
+        with raises(ValueError):
+            po2php.convertphp(inputfile, outputfile, None)
+
+    def test_convertphp_empty(self):
+        """test convertphp helper"""
```
unho: Test must have an actual description.

translate/convert/test_po2php.py
```diff
@@ -28,6 +28,47 @@ def merge2php(self, phpsource, posource):
...
+msgstr "waarde"
+'''
+        phpexpected = b'''$lang['name'] = 'waarde';
+'''
+        inputfile = wStringIO.StringIO(posource)
+        outputfile = wStringIO.StringIO()
+        with raises(ValueError):
+            po2php.convertphp(inputfile, outputfile, None)
+
+    def test_convertphp_empty(self):
```
unho: I would rename it to `test_convertphp_empty_template`.

translate/convert/test_po2php.py
```diff
@@ -46,6 +87,44 @@ def test_space_preservation(self):
         print(phpfile)
         assert phpfile == [phpexpected]
 
+    def test_unused_preservation(self):
```
unho: Nit: Rename to `test_preserve_unused_statements`.

translate/convert/test_po2php.py
unho: That's true. Improvements and cleanups are welcome. I have myself a branch refactoring all TTK converters to a common pattern that I want to start merging after your PHP changes.

translate/storage/test_php.py
unho: I don't see how

```php
"Register <a href=\"register.php\">here</a>"
```

is invalid PHP code. If I am not wrong this renders to valid HTML:

```html
Register <a href="register.php">here</a>
```

nijel: But here is not `\"register.php\"`, but only `"register.php"`!

```python
>>> print """define("_REGISTER", "Register <a href=\\"register.php\\">here</a>");"""
define("_REGISTER", "Register <a href=\"register.php\">here</a>");
>>> print """define("_REGISTER", "Register <a href=\"register.php\">here</a>");"""
define("_REGISTER", "Register <a href="register.php">here</a>");
```

nijel: Alternatively (if you don't like double escaping), this could be addressed by using raw string:

```patch
diff --git a/translate/storage/test_php.py b/translate/storage/test_php.py
index fa1b11ca..45e99483 100644
--- a/translate/storage/test_php.py
+++ b/translate/storage/test_php.py
@@ -323,7 +323,7 @@ define( '_CM_POSTED','Enviado');"""
 
     def test_parsing_define_entries_with_quotes(self):
         """Parse define syntax for entries with quotes"""
-        phpsource = """define('_SETTINGS_COOKIEPREFIX', 'Prefixo da "cookie"');
+        phpsource = r"""define('_SETTINGS_COOKIEPREFIX', 'Prefixo da "cookie"');
 define('_YOUR_USERNAME', 'O seu nome de usuario: "cookie"');
 define("_REGISTER", "Register <a href=\"register.php\">here</a>");"""
         phpfile = self.phpparse(phpsource)
```

unho: Works for me.

nijel:
I've changed the escaping bit in the php test, the coverage is as much close to 100% as possible.

translate/convert/test_po2php.py
```diff
@@ -46,6 +85,44 @@ def test_space_preservation(self):
...
+        phptemplate = '''
+error_reporting(E_ALL);
+$lang['name']  =  'value';
+'''
+        phpexpected = '''
+error_reporting(E_ALL);
+$lang['name']  =  'waarde';
+'''
+        phpfile = self.merge2php(phptemplate, posource)
+        print(phpfile)
```
unho: Nit: unnecessary statement.

translate/convert/test_po2php.py
```diff
@@ -46,6 +85,44 @@ def test_space_preservation(self):
...
+$lang['second']  = "
+value";
+'''
+        phpexpected = '''
+$lang['name']  =  'waarde';
+$lang['second']  = "
+value";
+'''
+        phpfile = self.merge2php(phptemplate, posource)
+        print(phpfile)
```
unho: Nit: Also unnecessary.

translate/convert/test_po2php.py
```diff
@@ -86,6 +163,22 @@ def test_inline_comments(self):
...
+msgstr "waarde"
+'''
+        phptemplate = '''/* some comment */
+$lang[ 'name' ]  =  'value';
+'''
+        phpexpected = '''/* some comment */
+$lang[ 'name' ]  =  'waarde';
+'''
+        phpfile = self.merge2php(phptemplate, posource)
+        print(phpfile)
```
unho: Nit: also unnecessary statement.

unho:
Raised some nits, but lgtm even without those. Very nice work.

nijel:
Thanks for review, merged (with nits fixed).

PR #3697, opened 2017-09-27
Use phply for parsing PHP files

### LINKED ISSUE #1685: php2po doensn't handle arrays where the last value does not end with a comma
_Version: trunk_

We can extract this:
    <?php

```
// Language definitions used in userlist.php
$lang_ul = array(

'User find legend'              =>      'Find and sort users',
'User search info'              =>      'Enter a username to search for and/or a user group to filter by. The username field can be left blank. Use the wildcard character * for partial matches. Sort users by name, date registered or number of posts and in ascending/descending order.',
'User group'                    =>      'User group',
'No of posts'                   =>      'No. of posts',
'All users'                             =>      'All'
);
```

Because the last entry does not close with a comma.

### LINKED ISSUE #1898: PHP conversion with space before comma fails
_Version: trunk_

We are not able to extract units from this file from the second unit onwards because we fail with a space before the comma.  Remove the command and we succeed.

``` php
$english = array(
         'item:site'  =>  "Sítios",
         'login'  =>  "Entrar" ,
         'loginok'  =>  "Entrou com sucesso." ,
```

### LINKED ISSUE #1929: Importing PHP arrays is broken-multiple arrays elements combined in one
_Version: 1.9.0_

_Originally posted by Johan:_

When importing a PHP array the parser gets confused by certain context (perhaps the HTML syntax?) and combines many keys/values into one single translation scell. When I imported ~1700 text lines roughly 900 of the strings ended up in the same cell with the PHP array syntax and all. (Unfortunately I cannot seem to add an attachment to show a screenshot which will help clarify what I mean)

It seems that the string key/text that managed to break the parser was (the array index is the key and the value the translation - in this case english->swedish):

$strings['The Bug Genie can use your <a href="http://www.gravatar.com" target="_blank">Gravatar</a> profile picture, if you have one. If you don't have one but still want to use Gravatar for profile pictures, The Bug Genie will use a Gravatar <a href="http://blog.gravatar.com/2008/04/22/identicons-monsterids-and-wavatars-oh-my/" target="_blank">auto-generated image unique for your email address</a>.'] = 'The Bug Genie kan använda din <a href="http://www.gravatar.com" target="_blank">Gravatar</a> profilbild, om du har en. Om du inte har ett men ändå vill använda Gravatar för profilen bilder, Bug Genie kommer användningen en Gravatar <a href="http://blog.gravatar.com/2008/04/22/identicons-monsterids-and-wavatars-oh-my/" target="_blank">automatiskt genererade bilden unika för din e-postadress</a> .'

### LINKED ISSUE #2611: php2po doesn't handle PHP translation files with heredoc syntax
_Version: 1.9.0_

php2po doesn't handle PHP translation files with heredoc syntax:

http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc

### LINKED ISSUE #2612: php2po doesn't handle PHP translation files with nowdoc syntax
_Version: 1.9.0_

php2po doesn't handle PHP translation files with nowdoc syntax:

http://php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc

### LINKED ISSUE #2647: php2po: failed to convert nested arrays declared in the next line
_Version: 1.9.0_

php2po failed to convert nested arrays declared in the next line, that is entries in an array that are arrays but are not declared in the same line as its key. For example the 'newsletter_frequency_dom' array in:

```php
$lang = array(
    'item1' => 'value1',
    'newsletter_frequency_dom' =>
        array(
            'Weekly' => 'Weekly',
        ),
    'item2' => 'value2',
);
```

### LINKED ISSUE #2648: php2po: failed to convert nested arrays with blank entries
_Version: 1.9.0_

php2po failed to convert nested arrays with blank entries. Blank entries have a blank string as key and a blank string as value:

$lang = array(
    'item1' => 'value1',
    'newsletter_frequency_dom' =>
        array(
            '' => '',
            'Weekly' => 'Weekly',
    ),
    'item2' => 'value2',
);

Found in SugarCRM translations.

Bye

### LINKED ISSUE #3608: PHP: Nested array in php store
```php
$t = array(
    'key' => ' text (comment **),**',
);
return $t;
```
If value contains ``**),**`` row is ignored

### LINKED ISSUE #3625: Use a third-party PHP parser for PHP support
I think that instead of having TTK's own minimal PHP parser which relies on regex and such, it might make sense to use a third-party PHP parser, even if it has way more to offer than necessary for TTK.

One such parser might be https://github.com/viraptor/phply. I haven't tested it, but the README looks quite promising.

### LINKED ISSUE #3626: PHP: Add support for short array syntax
Reported by @rimas-kudelis:

Since PHP 5.4 it is possible to have arrays using the [short array syntax](http://php.net/manual/en/language.types.array.php):

```php
<?php
return [
    'key2' => 'Othervalue2',
];
```

The issue was originally spotted with:

```php
<?php
return array(
    'parent' => array(
        'key1' => 'value1',
        'key2' => 'value2',
    ),
    'otherparent' = [
        'key1' => 'Othervalue1',
        'key2' => 'Othervalue2',
    ],
    'thirdparent' = [
        'key1' => 'Thirdvalue1',
        'key2' => 'Thirdvalue2',
    ],
);
```

It might be worth investigating if #3625 is a way forward instead of adding more bits to current parser. But first add tests.

### LINKED ISSUE #3627: PHP: Incorrectly parsing strings with /* on them
Reported by @rimas-kudelis:

Strings that contain a `/*` within them are not correctly parsed:

```php
<?php
return array(
    'parent1' => array(
        'key1' => 'value1',
        'key2' => 'value2 /* value2 continued',
        'key3' => 'value3',
        'key4' => 'value4',
    ),
    'parent2' => array(
        'key5' => 'value5',
        'key6' => 'value6',
    ),
);
```

### LINKED ISSUE #3628: PHP: Can't parse if newline between delimiter and value
Reported by @rimas-kudelis:

When there is a newline after delimiter (`=>` in the example below) and the value within an array the parser doesn't work:

```php
<?php
return array(
    'parent3'=> array(
            'key1' =>
                'value2',
        ),
);
```

### LINKED ISSUE #3629: PHP: Can't parse array defined in a single line
Reported by @rimas-kudelis:

```php
<?php
return array(
    'parent'=>array( 'key1' => 'value1', 'key2' => 'value2', 'key3' => 'value3', ),
);
```

nijel:
Current PHP parser is really problematic, so how about using some real parser for PHP?

I've tried to base on https://github.com/viraptor/phply. ~~There are some tricky parts to solve (naming of units is slightly different and comments extraction does not work), but I'd like to get some feedback before spending more time on this.~~ It tries to be as much compatible as possible, but I've decided to change comments handling, see https://github.com/translate/translate/pull/3697#issuecomment-332743699 for more details.

This bases on https://github.com/translate/translate/pull/3696 to get fixed some errors in tests.

Fixes #1685
Fixes #1898
Fixes #1929
Fixes #2611
Fixes #2612
Fixes #2647
Fixes #2648
Fixes #3608
Fixes #3625
Fixes #3626
Fixes #3627
Fixes #3628
Fixes #3629

### DIFF
--- a/translate/storage/php.py
+++ b/translate/storage/php.py
@@ -54,11 +54,138 @@ implemented as outlined in the PHP documentation for the
 `String type <http://www.php.net/language.types.string>`_.
 """
 
-import logging
+from __future__ import unicode_literals
+
 import re
+
 import six
 
+from phply.phpparse import make_parser
+from phply.phplex import lexer, FilteredLexer, full_lexer, unparsed
+from phply.phpast import (Array, ArrayElement, ArrayOffset, Assignment,
+                          BinaryOp, FunctionCall, InlineHTML, Node, Return,
+                          Variable)
+
 from translate.storage import base
+
+
+def wrap_production(func):
+    """Decorator for production functions to store lexer positions."""
+    def prod(n):
+        func(n)
+        if isinstance(n[0], Node):
+            startpos = min([getattr(i, 'lexpos', 0) for i in n.slice[1:]])
+            endpos = max([getattr(i, 'endlexpos', 0) for i in n.slice[1:]])
+            n[0].lexpositions = startpos, endpos
+        elif isinstance(n[0], list) and n[0] and isinstance(n[0][-1], ArrayElement):
+            startpos = min([getattr(i, 'lexpos', 0) for i in n.slice[1:]])
+            endpos = max([getattr(i, 'endlexpos', 0) for i in n.slice[1:]])
+            n[0][-1].lexpositions = startpos, endpos
+    return prod
+
+
+class PHPLexer(FilteredLexer):
+    def __init__(self):
+        super(PHPLexer, self).__init__(full_lexer.clone())
+        self.tokens = []
+        self.pos = 0
+        self.codepos = 0
+
+    def next_token(self):
+        token = self.lexer.token()
+        if token is not None:
+            self.tokens.append(token)
+        return token
+
+    def token(self):
+        """Copied from FilteredLexer, only difference is that it uses
+        next_token method
+        """
+
+        t = self.next_token()
+
+        # Filter out tokens that the parser is not expecting.
+        while t and t.type in unparsed:
+
+            # Skip over open tags, but keep track of when we see them.
+            if t.type == 'OPEN_TAG':
+                if self.last_token and self.last_token.type == 'SEMI':
+                    # Rewrite ?><?php as a semicolon.
+                    t.type = 'SEMI'
+                    t.value = ';'
+                    break
+                self.last_token = t
+                t = self.next_token()
+                continue
+
+            # Rewrite <?= to yield an "echo" statement.
+            if t.type == 'OPEN_TAG_WITH_ECHO':
+                t.type = 'ECHO'
+                break
+
+            # Insert semicolons in place of close tags where necessary.
+            avoid_tags = ('OPEN_TAG', 'SEMI', 'COLON', 'LBRACE', 'RBRACE')
+            rewrite_as_semicolon = (t.type == 'CLOSE_TAG' and
+                                    (not self.last_token or
+                                     self.last_token.type not in avoid_tags))
+            if rewrite_as_semicolon:
+                # Rewrite close tag as a semicolon.
+                t.type = 'SEMI'
+                break
+
+            t = self.next_token()
+
+        self.last_token = t
+        return t
+
+    def extract_comments(self, end):
+        """Extract comments related to given parser positions.
+
+        Must be called sequentially for consequent statements.
+        """
+        comments = []
+        # Process all tokens to end of statement
+        while self.tokens[self.pos].lexpos < end:
+            if self.tokens[self.pos].type in ('COMMENT', 'DOC_COMMENT'):
+                comments.append(self.tokens[self.pos].value.strip())
+            self.pos += 1
+        # Skip end of statement
+        self.pos += 1
+        # Proceed comments till newline
+        length = len(self.tokens)
+        while self.pos < length and self.tokens[self.pos].type in ('COMMENT', 'WHITESPACE'):
+            token_type = self.tokens[self.pos].type
+            token_value = self.tokens[self.pos].value
+            self.pos += 1
+            if token_type == 'WHITESPACE':
+                if '\n' in token_value:
+                    break
+                continue
+            comments.append(token_value.strip())
+            if '\n' in token_value:
+                break
+        return comments
+
+    def extract_name(self, terminator, start, end):
+        """Extract current value name."""
+        result = ''
+        pos = self.pos
+        while self.tokens[pos].lexpos < start:
+            pos += 1
+        while self.tokens[pos].lexpos < end and self.tokens[pos].type != terminator:
+            result += self.tokens[pos].value
+            pos += 1
+        self.codepos = pos
+        return result.rstrip()
+
+    def extract_quote(self):
+        """Extract quote style."""
+        pos = max(self.pos, self.codepos)
+        while self.tokens[pos].type not in ('QUOTE', 'CONSTANT_ENCAPSED_STRING', 'START_NOWDOC'):
+            pos += 1
+        if self.tokens[pos].type == 'QUOTE':
+            return '"'
+        return '\''
 
 
 def phpencode(text, quotechar="'"):
@@ -129,7 +256,7 @@ class phpunit(base.TranslationUnit):
 
     def __init__(self, source=""):
         """Construct a blank phpunit."""
-        self.escape_type = None
+        self.escape_type = '\''
         super(phpunit, self).__init__(source)
         self.name = "$TTK_PLACEHOLDER"
         self.value = ""
@@ -158,9 +285,21 @@ class phpunit(base.TranslationUnit):
         """Convert to a string."""
         return self.getoutput()
 
-    def getoutput(self):
+    def getoutput(self, indent='', name=None):
         """Convert the unit back into formatted lines for a php file."""
-        return "\n".join(self._comments + ["%s='%s';\n" % (self.name, self.translation or self.value)])
+        if '->' in self.name:
+            fmt = '{0} => {1}{2}{1},\n'
+        elif self.name.startswith('define'):
+            fmt = '{0}, {1}{2}{1});\n'
+        else:
+            fmt = '{0} = {1}{2}{1};\n'
+        out = fmt.format(
+            name if name else self.name,
+            self.escape_type,
+            self.translation or self.value
+        )
+        joiner = '\n' + indent
+        return indent + joiner.join(self._comments + [out])
 
     def addlocation(self, location):
         self.name = location
@@ -194,6 +333,9 @@ class phpunit(base.TranslationUnit):
     def getid(self):
         return self.name
 
+    def setid(self, value):
+        self.name = value
+
 
 class phpfile(base.TranslationStore):
     """This class represents a PHP file, made up of phpunits."""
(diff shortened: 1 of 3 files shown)


nijel:
In the end the remaining bits were not that complex, so the PR is complete now.

It does change handling of comments - now it's consistently including comment markers (previously it did so for `/* ... */` comments but did not for `//...` comments) and multiline block comments are inserted as one note instead of splitting them up. Both could be changed, but I think this way it's better.

nijel:
The Python 2 tests will fail until https://github.com/viraptor/phply/pull/33 is accepted and released.

nijel:
Okay, the above mentioned fix got merged, but new version introduced another regression, so the tests will fail until https://github.com/viraptor/phply/pull/39 is merged...

nijel:
All needed fixes have landed in phply, so all tests now pass :-).

nijel:
BTW: This is running since September 2017 on Hosted Weblate without any problems, so at least for storage part it is pretty well tested in the real life.

translate/storage/php.py
```diff
@@ -54,13 +54,133 @@
 `String type <http://www.php.net/language.types.string>`_.
 """
 
-import logging
 import re
 import six
+from phply.phpparse import make_parser
+from phply.phplex import lexer, FilteredLexer, full_lexer, unparsed
+from phply.phpast import InlineHTML, FunctionCall, Assignment, ArrayOffset, Variable, Array, ArrayElement, Return, BinaryOp, Node
```
unho: These imports don't follow our import guidelines: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#imports

Also the lines shouldn't be that long.

phlax: wierd that it didnt break any linting checks @unho

translate/storage/php.py
```diff
@@ -54,13 +54,133 @@
...
+
+    def next_token(self):
+        token = self.lexer.token()
+        if token is not None:
+            self.tokens.append(token)
+        return token
+
+    def token(self):
+        # Copied from FilteredLexer, only difference is that it uses
+        # next_token method
```
unho: Nit: If the comment applies to the whole method, then make it a docstring.

translate/storage/test_php.py
```diff
@@ -164,7 +165,7 @@ def test_comment_output(self):
  * Comment line 1
  * Comment line 2
  */
-$foo='bar';
+$foo = 'bar';
```
unho: Remove the comment 5 lines up since I think no longer applies.

translate/storage/test_php.py
```diff
@@ -763,3 +755,8 @@ def test_simpledefinition_after_define(self):
         phpunit = phpfile.units[1]
         assert phpunit.name == "$lang['mediaselect']"
         assert phpunit.source == "Bestand selectie"
+
+    def test_quotes(self):
+        phpsource = b'<?php\n$txt[\'DISPLAYEDINFOS\'] = "<a href=\\"__PARAM2__\\">Modificar...</a>";\n'
```
unho: Is it not possible to use triple quoted string to avoid having `\n` in the middle of the string?

translate/storage/test_php.py
```diff
@@ -762,6 +762,17 @@ def test_quotes(self):
         phpfile = self.phpparse(phpsource)
         assert phpfile.__bytes__() == phpsource
 
+    def test_concatenation(self):
+        """Check that a simple definition after define is parsed correctly."""
+        phpsource = """
+$lang['mediaselect'] = "Really \\\"something\\\"" . $variable . "\\\"something else\\\"";"""
+        phpfile = self.phpparse(phpsource)
+        print(len(phpfile.units))
```
unho: Unnecessary `print` call.

unho:
Can you please add a specific test for #1898?

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
+                        name,
+                        item.value,
+                        lexer.extract_quote(),
+                        lexer.extract_comments(item.lexpositions[1]),
+                    )
+
+        def concatenate(item):
+            if isinstance(item, six.string_types):
+                return item
+            assert isinstance(item, BinaryOp)
```
unho: Is this `assert` necessary? If not please remove.

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
+            self.addunit(newunit)
+
+        def format_key(key):
+            if isinstance(key, (int, float)):
+                return '{0}'.format(key)
+            return '\'{0}\''.format(key)
+
+        def handle_array(prefix, nodes, lexer):
+            for item in nodes:
+                assert isinstance(item, ArrayElement)
```
unho: Again please avoid use of `assert` unless necessary.

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
-                # fall back to default dialect (PHP simple variable syntax).
-                if not inarray:
-                    equaldel = "="
-                    enddel = ";"
-
-            # If this is part of a multiline translation, just append it to the
-            # previous translation lines.
-            if invalue:
-                lastvalue = lastvalue + value + "\n"
+        def newunit(name, value, escape_type, comments):
```
unho: I would rename this to `create_and_add_unit` and make it a method.

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
+            newunit = phpunit()
```
unho: Nit: Use `self.UnitClass()` instead of `phpunit()`.

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
+            if isinstance(item, six.string_types):
+                return item
+            assert isinstance(item, BinaryOp)
+            return concatenate(item.left) + concatenate(item.right)
+
+        parser = make_parser()
+        for item in parser.productions:
+            item.callable = wrap_production(item.callable)
+        lexer = PHPLexer()
+        pos = 0
```
unho: Unused variable?

translate/storage/php.py
```diff
@@ -216,171 +339,88 @@ def serialize(self, out):
...
+        tree = parser.parse(phpsrc.decode(self.encoding), lexer=lexer, tracking=True)
+        # Handle text without PHP start
+        if len(tree) == 1 and isinstance(tree[0], InlineHTML):
+            return self.parse(b'<?php\n' + phpsrc)
```
unho: Is it necessary to call `self.parse`? Doesn't the following work?:

```python
            phpsrc = b'<?php\n' + phpsrc
            tree = parser.parse(phpsrc.decode(self.encoding), lexer=lexer, tracking=True)
```

This way we should be able to avoid initializing stuff again.

translate/storage/php.py
```diff
@@ -54,13 +54,133 @@
...
+
+            # Insert semicolons in place of close tags where necessary.
+            if t.type == 'CLOSE_TAG':
+                if self.last_token and self.last_token.type in ('OPEN_TAG', 'SEMI', 'COLON', 'LBRACE', 'RBRACE'):
+                    # Dont insert semicolons after these tokens.
+                    pass
+                else:
+                    # Rewrite close tag as a semicolon.
+                    t.type = 'SEMI'
+                    break
```
unho: Better to shorten this code block like:

```python
            # Insert semicolons in place of close tags where necessary.
            avoid_tags = ('OPEN_TAG', 'SEMI', 'COLON', 'LBRACE', 'RBRACE')
            rewrite_as_semicolon = (t.type == 'CLOSE_TAG' and
                                    (not self.last_token or
                                     self.last_token.type not in avoid_tags))
            if rewrite_as_semicolon:
                # Rewrite close tag as a semicolon.
                t.type = 'SEMI'
                break
```

unho:
@nijel This is really amazing. I've left some comments and nits. I am really **really** sorry about it took me so long to review this great contribution. lgtm once comments are addressed.

unho:
Weird, some of the comments I've made are shown collapsed despite they still apply.

translate/storage/php.py
nijel: The lexer has internal state, so it should be reinitialized. I'm not sure about parser, so I've rather did it the safe way...

translate/storage/php.py
nijel: Good point. At some earlier point it did use some of the namespace, but it's now gone.

translate/storage/test_php.py
nijel: It's used all over the file this way, so it's there after copying existing code. Will remove that in separate commit.

translate/storage/php.py
nijel: It's there merely to catch in case parser produces something wrong. Therefore I've used assert which would discover this while running tests, but it's situation which really should not happen.

nijel:
Thanks for the review, I've addressed most of the comments (though it's possible that I've missed something due to GitHub hiding them as outdated). The only thing which I didn't change are the assert statements, I think they are useful.

translate/storage/php.py
unho: I am ok with leaving it as is.

translate/storage/test_php.py
unho: Thanks.

translate/storage/php.py
```diff
@@ -54,13 +54,140 @@
...
+
 import re
+
 import six
 
+from phply.phpparse import make_parser
+from phply.phplex import lexer, FilteredLexer, full_lexer, unparsed
+from phply.phpast import (InlineHTML, FunctionCall, Assignment, ArrayOffset,
+                          Variable, Array, ArrayElement, Return, BinaryOp,
+                          Node)
```
unho: These imports still need some changes to conform to http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#imports

translate/storage/test_php.py
```diff
@@ -755,11 +733,131 @@ def test_simpledefinition_after_define(self):
...
+        'and' => 'Und',
+        'second' => 'Zweite',
+    ),
+);
+"""
+        phpfile = self.phpparse(phpsource)
+        assert phpfile.__bytes__() == phpsource
+
+    def test_space_before_comma(self):
+        """check that spacing before comma or semicolon doesn't break parser"""
```
unho: Nit: Add reference in docstring to issue #1898

translate/storage/test_php.py
```diff
@@ -755,11 +733,131 @@ def test_simpledefinition_after_define(self):
...
+        phpsource = """
+        $english = array(
+            'item:site'  =>  "Sítios",
+            'login'  =>  "Entrar" ,
+            'loginok'  =>  "Entrou com sucesso." ,
+        );
+        $month_jan = 'Jan';
+        $month_feb = 'Feb'  ;
+        $month_mar = 'Mar';
```
unho: Also include the following test cases:

```php
define("_SEARCH","Search");
define("_LOGIN","Login"  );
define("_POLLS","Polls");
```

translate/storage/test_php.py
```diff
@@ -755,11 +733,131 @@ def test_simpledefinition_after_define(self):
...
+        """
+        phpfile = self.phpparse(phpsource)
+        assert len(phpfile.units) == 6
+
+    def test_equals_in_id(self):
+        """check that equals in id doesn't break parser"""
```
unho: Please add reference to issue #1929 in test's docstring.

translate/storage/test_php.py
```diff
@@ -755,11 +733,131 @@ def test_simpledefinition_after_define(self):
...
+        phpsource = """
+        $strings['key = value'] = 'Message';
+        """
+        phpfile = self.phpparse(phpsource)
+        assert len(phpfile.units) == 1
+        assert phpfile.units[0].source == 'Message'
+        assert phpfile.units[0].name == "$strings['key = value']"
+
+    def test_comma_in_string(self):
+        """check that comma in string doesn't break parser"""
```
unho: Also add reference in this docstring to the related issue #3608.

translate/storage/test_php.py
```diff
@@ -755,11 +733,131 @@ def test_simpledefinition_after_define(self):
...
+        """
+        phpfile = self.phpparse(phpsource)
+        assert len(phpfile.units) == 1
+        assert phpfile.units[0].source == '''Example of string
+spanning multiple lines
+using nowdoc syntax.'''
+        assert phpfile.units[0].name == '$str'
+
+    def test_concatenation(self):
+        """check parsing concatenadted strings"""
```
unho: Typo: concatenadted

unho:
@nijel reviewed the PR again and raised some extra nits, mostly on the new changes. lgtm after those are addressed. Again, thanks a lot!

unho:
Sorry, hit the wrong button.

nijel:
Thanks for review.

rimas-kudelis:
Wow, thanks for giving this a try! I'm pretty sure TTK has become even better now. :)

PR #3700, opened 2017-10-11
Add --no-segmentation flag to txt2po

### LINKED ISSUE #3698: txt2po should support not splitting by paragraph
We are using Pootle and translate to localise metadata for apps in the iOS App Store. Our data source are [deliver](https://github.com/fastlane/fastlane/tree/master/deliver)'s TXT files. Some of them, e.g. for the app description or the release notes contain long text, sometimes split across several paragraphs. When converting from TXT to PO, these end up in separate messages, e.g.

```
> cat test.txt 
foo bar foo bar foo bar

123 456 123 456 123 456

foo bar foo bar foo bar
> txt2po --i test.txt -o test.po
processing 1 files...
[###########################################] 100%
> cat test.po
#. extracted from test.txt
#, fuzzy
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE VERSION\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2017-10-10 15:29+0200\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"X-Generator: Translate Toolkit 2.2.5\n"

#: test.txt:2
msgctxt "test.txt:2"
msgid "foo bar foo bar foo bar"
msgstr ""

#: test.txt:4
msgid "123 456 123 456 123 456"
msgstr ""

#: test.txt:5
msgctxt "test.txt:5"
msgid "foo bar foo bar foo bar"
msgstr ""
```

This behavior is not always desirable for us since in some cases certain locales may not include all paragraphs of the English version, or may have to include additional paragraphs not present in English.

For this use case it would be nice to be able to keep the whole TXT content in one single message when converting to PO, e.g. by adding a `-1` flag.

Johennes:
This adds a flag `--no-segmentation` to txt2po which when specified will cause the whole file to be dumped into a single messages in the PO file.

Fixes  #3698

### DIFF
--- a/translate/convert/txt2po.py
+++ b/translate/convert/txt2po.py
@@ -35,7 +35,8 @@ class txt2po(object):
     TargetUnitClass = po.pounit
 
     def __init__(self, input_file, output_file, template_file=None,
-                 duplicate_style="msgctxt", encoding="utf-8", flavour=None):
+                 duplicate_style="msgctxt", encoding="utf-8", flavour=None,
+                 nosegmentation=False):
         """Initialize the converter."""
         self.duplicate_style = duplicate_style
 
@@ -43,14 +44,16 @@ class txt2po(object):
         self.output_file = output_file
         self.source_store = self.SourceStoreClass(input_file,
                                                   encoding=encoding,
-                                                  flavour=flavour)
+                                                  flavour=flavour,
+                                                  nosegmentation=nosegmentation)
         self.target_store = self.TargetStoreClass()
         self.template_store = None
 
         if template_file is not None:
             self.template_store = self.SourceStoreClass(template_file,
                                                         encoding=encoding,
-                                                        flavour=flavour)
+                                                        flavour=flavour,
+                                                        nosegmentation=nosegmentation)
 
     def convert_store(self):
         """Convert a single source format file to a target format file."""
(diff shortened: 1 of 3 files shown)


Johennes:
@unho any thoughts on this?

unho:
@Johennes Didn't notice this PR and I have refactored the converter a bit. Your approach looks ok, so if you rebase and add some tests I will give it another look. Thanks for the contribution.

translate/convert/txt2po.py
```diff
@@ -74,5 +75,9 @@ def main(argv=None):
                       help="The flavour of text file: plain (default), dokuwiki, mediawiki",
                       metavar="FLAVOUR")
     parser.passthrough.append("flavour")
+    parser.add_option("-1", "--onemessage", dest="onemessage",
```
unho: Nit: I would rather call this `--no-segmentation` or similar.

translate/convert/txt2po.py
```diff
@@ -74,5 +75,9 @@ def main(argv=None):
...
+                      default=False, action="store_true",
+                      help="Treat whole file as one single message")
```
unho: I would rephrase like `Don't segment the file, treat it like a single message`.

Johennes:
@unho I rebased on master and renamed the parameter. Please let me know if you'd like to see anything else fixed. Thanks for looking at this. :)

PR #3701, opened 2017-10-24
Fix local, editable installation instructions

Gallaecio:
The current instructions are wrong, as the referenced file does not seem to exist.

The provided solution is a more modern approach, I believe, and works both inside and outside a virtual environment.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -147,8 +147,7 @@ or, running the tools from the source directory:
 
 .. code-block:: console
 
-   $ ./setuppath # Only needed the first time
-   $ . setpath  # Do this once for a session
+   $ sudo pip3 install -e .
 
 
 .. _installation#verify_installed_version:


docs/installation.rst
```diff
@@ -147,8 +147,7 @@ or, running the tools from the source directory:
 
 .. code-block:: console
 
-   $ ./setuppath # Only needed the first time
-   $ . setpath  # Do this once for a session
+   $ sudo pip3 install -e .
```
nijel: - I'd call just pip instead of pip3.
- There is no need for sudo, it will install in ~/.local or in virtualenv.

Gallaecio: Yes, just «pip» is probably better.

I do believe you need sudo, otherwise I get:

    [Errno 13] Permission denied: '/usr/lib/python3.5/site-packages/test-easy-install-1587.write-test'

However, maybe sudo could be mentioned as a note below, something in the lines of “you might need to run this command as root for a system-wide installation”.

nijel: When I execute pip as non root user it defaults to --user. However I've just found out that this is [patched on Debian and Ubuntu](https://anonscm.debian.org/cgit/python-modules/packages/python-pip.git/tree/debian/patches/set_user_default.patch) and is not yet an upstream behavior, see https://github.com/pypa/pip/issues/1668.

Anyway you certainly don't want sudo if you're in virtualenv :-).

unho: In the previous snippet a few lines above `su` is used, so I understand that this installation method assumes you are installing as `root` and thus `sudo` is not needed. I am wondering if it is better to change all of this to assume a virtualenv is used.

nijel: I've made the local example consistent with normal install few lines above in https://github.com/translate/translate/pull/3758.

nijel:
Superseded by https://github.com/translate/translate/pull/3758

PR #3702, opened 2017-10-24
pypo: Allow to configure output width

nijel:
Setting this to high value can avoid wrapping altogether to mimic --no-wrap from gettext, setting to -1 completely removes line wrapping (including newlines).

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -83,22 +83,26 @@ wrapper.wordsep_re = re.compile(
 wrapper.wordsep_re_uni = re.compile(wrapper.wordsep_re.pattern, re.UNICODE)
 
 
-def quoteforpo(text):
+def quoteforpo(text, wrapper_obj=None):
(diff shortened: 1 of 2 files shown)


unho:
I have two concerns with this PR:

- It only changes one of the implementations of the PO format.
- It passes around a `wrapper` object which only seems to differ in its width, so is it not better to instead pass the width and then set the width in the `wrapper` when calling it in `quoteforpo` function?

nijel:
AFAIK there is no way to achieve this with libgettextpo, but I might be wrong.

I've chosen passing wrapper as this can be constructed once for the file and used then. In case only width would be passed, then quoteforpo would have to construct new wrapper every time or change attribute of global instance on every call, what sounds like bad pattern and could lead to not expected side effects for callers. This way quoteforpo without parameter behave as it used to.

unho:
Ok, given the provided response I am ok with lgtm this. If any problem arises we can revisit it.

nijel:
Thanks!

PR #3708, opened 2017-11-22
Cleanup: Consistenly use pytest.raises as context manager

nijel:
The code is currently pretty inconsistent as pointed out in https://github.com/translate/translate/pull/3696#discussion_r152043518


phlax:
lgtm

unho:
Very nice cleanup.

PR #3709, opened 2017-11-22
YAML: Gracefully handle adding nested keys in place of values

### LINKED ISSUE #3703: YAML serialization raises an exception when a node disappears
https://github.com/translate/translate/blob/e5d4d38fbcc7fb310683e7b12f9ae7deab9d7788/translate/storage/yaml.py#L142

```Exception Value:	'str' object does not support item assignment```

In my case (through weblate), the existing file has:
```
it:
  base:
    path: Italian path
```

And a new node is used in the source translation (`base->path->sublevel`):
```
en:
  base:
    path:
      sublevel: Now I want these nested actually
```

The code in `serialize` will raise an exception on line 154 (and make the file empty).

nijel:
Fixes #3703


unho:
lgtm

nijel:
Thanks for review!

Issue #3681, opened 2017-09-01
RemovedInTTK2Warning seems strange


nijel:
There is ``RemovedInTTK2Warning`` which apparently was meant to flag feature which will be removed in translate-toolkit 2. However it is already out and that did not happen :-).

Either RemovedInTTK2Warning should be renamed as translate-toolkit 2 has already been released, or the deprecation should be applied.

However it seems that quite a lot of the code seems to rely on that behavior.

nijel:
The code update is in https://github.com/translate/translate/pull/3710

PR #3710, opened 2017-11-22
Cleanup: Pass unicode strings to multistring

nijel:
See https://github.com/translate/translate/issues/3681

### DIFF
--- a/translate/storage/poxliff.py
+++ b/translate/storage/poxliff.py
@@ -28,6 +28,7 @@ import six
 
 from lxml import etree
 
+from translate.lang import data
 from translate.misc.multistring import multistring
 from translate.misc.xml_helpers import setXMLspace
 from translate.storage import base, lisa, poheader, xliff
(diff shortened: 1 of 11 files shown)


translate/storage/poxliff.py
```diff
@@ -150,8 +151,9 @@ def settarget(self, text, lang='xx', append=False):
 
     def gettarget(self):
         if self.hasplural():
-            strings = [unit.target for unit in self.units]
+            strings = [data.forceunicode(unit.target) for unit in self.units]
             if strings:
+                print(strings)
```
unho: Nit: unnecessary print.

unho:
lgtm after nit

translate/storage/poxliff.py
nijel: Apparently I forgot some of the debugging code...

nijel:
Thanks, merged!

PR #3711, opened 2017-11-22
Get only name attribute when it exists

nijel:
This avoids raising 'DeprecationWarning: use the name attribute' on some
files.


unho:
lgtm

nijel:
Thanks for review!

PR #3712, opened 2017-11-22
Make DeprecationWarning a failure in tests

nijel:
I think this would be useful to avoid using deprecated interfaces.

Depends on #3711 and #3710


unho:
lgtm provided all checks pass.

nijel:
In the end I had to do two more fixes to make this pass on all python versions...

unho:
lgtm. Nice cleanups!

Issue #3751, opened 2017-12-19
pypo fails to parse Gettext file with unix newlines on Windows


nijel:
The poparser assumes that the file uses platform native newlines, what can easily be false on Windows when working with Gettext as most of the Gettext files use unix newlines. In this case the resulting pypo is empty.

nijel:
Testcase covering this is in https://github.com/translate/translate/pull/3752.

PR #3752, opened 2017-12-19
pypo: Add tests for parsing files with different newlines

nijel:
Note that this currently *is* broken on Windows - if the file has unix newlines, it won't get properly parsed, see https://github.com/translate/translate/issues/3751

### DIFF
--- a/translate/storage/test_pypo.py
+++ b/translate/storage/test_pypo.py
@@ -369,3 +369,17 @@ msgstr[1] "toetse"
         store.units[0].settarget('Hello ' * 100)
         # Should contain additional newlines now
         assert bytes(store) != posource
+
+    def test_unix_newlines(self):
+        """checks that unix newlines are properly parsed"""
+        posource = b'\nmsgid "test me"\nmsgstr ""\n'
+        pofile = self.poparse(posource)
+        assert len(pofile.units) == 1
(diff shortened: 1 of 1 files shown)


translate/storage/test_pypo.py
```diff
@@ -369,3 +369,17 @@ def test_wrap_newlines(self):
...
+
+    def test_unix_newlines(self):
+        """checks that unix newlines are properly parsed"""
+        posource = b'\nmsgid "test me"\nmsgstr ""\n'
+        pofile = self.poparse(posource)
+        assert len(pofile.units) == 1
+        assert pofile.units[0].source == 'test me'
+
+    def test_dos_newlines(self):
+        """checks that unix newlines are properly parsed"""
```
unho: Nit: Docstring should refer to DOS.

unho:
When a single Travis run fails and it doesn't look related to PR it is worth a try restarting that specific run on Travis (provided you have access to do so).

PR lgtm after nit.

nijel:
The build failure is permanent, but looks like a bug in Sphinx or in the docs which I didn't touch at all...

unho:
It looks like I inadvertently introduced this error with my latest commit in master. I will rollback that commit once you land this PR.

nijel:
Great, merged.

Issue #3664, opened 2017-07-05
rc2po does not work on input with windows newlines


learn-more:
When Reading .rc files that use windows newlines, rc2po failts to output anything.
(No error message, no .po file).
By adding:
`rcsrc = rcsrc.replace('\r\n', '\n')`
in `rcfile.__init__` between the close and the parse, files are output successfully.

dwaynebailey:
@learn-more could you PR that proposed with with a test? That would be greatly appreciated and help get that code into TTK.

learn-more:
Well it is more of a workaround, since it does not remember line endings, end thus effectively destroys them.

SantosSi:
I do not understand: replacing the Windows style newlines with *nix style ones make them forgotten? Isn't it possible to make the software work with Windows style newlines instead, e.g. by finding the incompatible code?

PR #3755, opened 2017-12-31
Ignore \r characters in RC files.

learn-more:
PR created for issue #3664

### DIFF
--- a/translate/storage/rc.py
+++ b/translate/storage/rc.py
@@ -125,6 +125,7 @@ class rcfile(base.TranslationStore):
         if inputfile is not None:
             rcsrc = inputfile.read().decode(self.encoding)
             inputfile.close()
+            rcsrc = rcsrc.replace('\r', '')
             self.parse(rcsrc)
 
     def parse(self, rcsrc):
--- a/translate/storage/test_rc.py
+++ b/translate/storage/test_rc.py
@@ -234,3 +234,59 @@ END
         rc_unit = rc_file.units[5]
         assert rc_unit.name == "STRINGTABLE.IDS_ERRORACTIV"
         assert rc_unit.source == "Error doing things"
+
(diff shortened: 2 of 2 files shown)


translate/storage/rc.py
```diff
@@ -125,6 +125,7 @@ def __init__(self, inputfile=None, lang=None, sublang=None, **kwargs):
         if inputfile is not None:
             rcsrc = inputfile.read().decode(self.encoding)
             inputfile.close()
+            rcsrc = rcsrc.replace('\r', '')
```
nijel: Wouldnt' it be better to adjust parser to treat '\r' same as '\n'? It already accepts '\r' in few places....

This way it needs double memory for replace and sole '\r' would be ignored, what looks like a bug to me.

learn-more: This is intentional (note the last testcase).
Visual studio (tested on 2015) would not compile an RC file if it used only '\r'

nijel: But when the compiler does not parse it then translate-toolkit should not parse it as well, or am I missing something?

learn-more: Which is why I removed all \r characters, so that the file would be parsed roughly the same as with the visual studio compiler (all as one line).

nijel: What I think compiler actually does is parse only '\r\n' as newline, can you check that only '\n' works? (I'm not on Windows...)

learn-more: I did, before writing the tests.
MSVC has been fine with just '\n' for years....

unho:
@learn-more Can you please rebase this branch on top of latest `master`? Thanks in advance.

nijel:
While I still think this is not the best approach, it's probably the simplest one. This parser anyway has quite many other problems (see issue tracker), so let's take the simple fix for now.

Issue #3754, opened 2017-12-29
TZ changes will cause tests to fail


llimeht:
`translate/storage/test_poheader.py` tests various timezone offsets as part of testing the header of the PO files. One such test at line 125 is

```
        os.environ['TZ'] = 'Africa/Windhoek'
        time.tzset()
        assert time.timezone == -3600
```
Namibia changes its TZ definitions in 2018 and so this test fails when run after 2018-04-01

```
tzdata
Release 2017c - 2017-10-20 14:49:34 -0700

  Changes to future time stamps

    Namibia will switch from +01 with DST to +02 all year on
    2017-09-03 at 02:00.  This affects UT offsets starting 2018-04-01
    at 02:00.  (Thanks to Steffen Thorsen.)
```
Picking some other timezone that isn't changing would be one solution. Simply dropping these lines from the test is probably also OK since they're not testing anything beyond what remains.

We discovered this test failure because all package builds (which includes running the tests) are routinely run 398 days into the future to spot packages that are including unwanted information like compilation dates in their output. See [build/test log](https://tests.reproducible-builds.org/debian/logs/buster/amd64/translate-toolkit_2.2.5-1.build2.log.gz), [variations used](https://tests.reproducible-builds.org/debian/index_variations.html), [more on reproducible builds](https://reproducible-builds.org/).

PR #3756, opened 2018-01-04
Stop testing with changing Africa/Windhoek TZ

llimeht:
The Africa/Windhoek TZ is changing; Namibia changes its TZ definitions in 2018
and so this test fails when run after 2018-04-01.

From tzdata Release 2017c - 2017-10-20 14:49:34 -0700:

    Namibia will switch from +01 with DST to +02 all year on
    2017-09-03 at 02:00.  This affects UT offsets starting 2018-04-01
    at 02:00.  (Thanks to Steffen Thorsen.)

This particular test isn't hitting code paths not addressed by other tests
so removing it is simplest.

Closes: https://github.com/translate/translate/issues/3754

### DIFF
--- a/translate/storage/test_poheader.py
+++ b/translate/storage/test_poheader.py
@@ -122,13 +122,6 @@ def test_timezones():
         # Typically "+0200"
         assert poheader.tzstring() == time.strftime("%z")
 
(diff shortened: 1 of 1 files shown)


nijel:
Thanks, I think this is reasonable approach for timezone which is changing.

PR #3758, opened 2018-01-05
Use pip for editable installation

nijel:
This is best approach for this.

See #3701


unho:
lgtm

nijel:
Thanks for review.

PR #3759, opened 2018-01-05
Avoid running sphinx in parallel

nijel:
This hides real error in some cases making it harder to figure out real
cause of build failure.

Right now the build is failing because of missing virtaal documentation
which is linked through intersphinx, see
https://github.com/translate/virtaal/issues/3277


nijel:
Merging to avoid cryptic failures of cron scheduled CI jobs on Travis.

PR #3760, opened 2018-01-05
Use splitlines instead of implicit iterating on file

### LINKED ISSUE #3751: pypo fails to parse Gettext file with unix newlines on Windows

nijel:
When the file is open in binary mode (what usually is), the implicit
iterator works only on platform native newlines. For example that means
that parser would fail to parse po files with unix newlines on Windows.

On the other side the splitlines uses universal newlines handling
correctly any newlines on any platform.

Fixes #3751

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -586,7 +586,7 @@ class pounit(pocommon.pounit):
         return len(self.msgid_plural) > 0
 
     def parse(self, src):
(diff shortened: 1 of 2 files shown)


unho:
lgtm, thanks for your contribution.

nijel:
Thanks for review!

PR #3761, opened 2018-01-05
RFC: Remove USECPO=1 tests

nijel:
These are failing and just wasting build resources.

AFAIK nobody is working on making this work, they can be easily added back once somebody does.


nijel:
Thanks for review.

PR #3762, opened 2018-01-09
Android: Avoid using Babel for plural tags

nijel:
These seem to be broken as they are missing the 'other' tag. The problem
is that not all languages use 'other' tag, so adding it back requires
CLDR knowledge. In this case it's better to have full tags defined
inside translate-toolkit than relying on broken library and fixing the
data.

See following issue trackers for more details:

* https://github.com/WeblateOrg/weblate/issues/1550
* https://github.com/python-babel/babel/issues/495


nijel:
This patch was confirmed to behave correctly in https://github.com/WeblateOrg/weblate/issues/1550#issuecomment-360266827

translate/storage/aresource.py
```diff
@@ -31,12 +31,219 @@
...
 
-from babel.core import Locale, UnknownLocaleError
-
 EOF = None
 WHITESPACE = ' \n\t'  # Whitespace that we collapse.
 MULTIWHITESPACE = re.compile('[ \n\t]{2}(?!\\\\n)')
 
+# List of plural tags generated from CLDR 32.0.1 using
+# https://github.com/WeblateOrg/language-data
+PLURAL_TAGS = {
```
unho: Please put this on `translate.lang.data` besides all other language-related data.

unho:
lgtm once the raised concern is addressed. Sorry for taking so long to review.

nijel:
Thanks, moved the data, will merge after tests pass.

PR #3763, opened 2018-01-10
Clean docs generated from storage/factory.py functions

llimeht:
Including the `classes_str` in the function signature of the factory means that sphinx includes the entire dict in each entry which is then hard to read. The inclusion of the `_examine_txt` function also results in a memory address being output by sphinx. The memory address is both ugly and non-reproducible so that the docs change slightly each time they are built.

In the [Reproducible Builds](https://reproducible-builds.org/) project, work is being done to help ensure that software can be reproduced, bit-for-bit as part of a chain of trust between author and user. Sphinx itself has already received numerous patches so that its output is now deterministic and with this change applied, `translate-toolkit` becomes reproducible too.

### DIFF
--- a/translate/storage/factory.py
+++ b/translate/storage/factory.py
@@ -31,7 +31,7 @@ decompressclass = {
 }
 
 
-classes_str = {
+_classes_str = {
     "csv": ("csvl10n", "csvfile"),
     "tab": ("omegat", "OmegaTFileTab"), "utf8": ("omegat", "OmegaTFile"),
     "po": ("po", "pofile"), "pot": ("po", "pofile"),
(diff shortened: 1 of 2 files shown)


translate/storage/factory.py
```diff
@@ -61,6 +61,8 @@
 _ext is a pseudo extension, that is their is no real extension by that name.
 """
 
+_classes_str = classes_str
```
nijel: Wouldn't be better to directly define the dict as _classes_str above instead of defining it as classes_str and then assigning to _classes_str.

That would need some other updates in the code, but is IMHO cleaner approach. Anyway translate/convert/pot2po.py should use None after this change instead of passing factory.classes_str as default.

llimeht: The reason for the double-shuffle assignment is that removing `classes_str` and `hiddenclasses` from factory.py is a non-backwards compatible change. Can we be sure that nothing is using it? (ttkit, pootle, virtaal and lokalize all seem to be OK; if that's enough checking, then let's do that.)

unho:
Thanks for your contribution. This change has been landed in fa4416bf.

PR #3772, opened 2018-03-05
JSON: Add support for customizing json.dumps args

nijel:
This can be used for example to force different indentation or keys
sorting.


unho:
lgtm

nijel:
Thanks for review.

PR #3776, opened 2018-03-11
Flat XML storage format

BhaaLseN:
This introduces a new monolingual storage format for a flat XML file that may look like this:
```xml
<root>
    <str key="hello_world">Hello World!</str>
    <str key="new_format">This is the new XML-based format.</str>
    <str key="flat_structure">It uses a flat structure.
Just a root element, and a number of what is basically key/value pairs.</str>
    <str key="customizable">Names and namespaces can be customized as parameters (so you can use any &lt;root&gt; or &lt;element&gt; name as you like).</str>
</root>
```
Main motivation to add such a format was to use it with Weblate (so CC @nijel in case he wants to chime in on things that might be required for Weblate; such as the mandatory `TranslationUnit.settarget` method), as none of the existing formats were close enough to what we use to provide translation in our XSLT based stylesheets (where XML simply made the most sense).

I'd still consider this WIP, as I could use some feedback/help on the following:
- [x] Choice of attributes: _Assuming this is fine, since base classes also do that. Leaving out comments for now (can always add them later if someone really wants them)_
   - The resource key goes to `msgid` (unless I am the only one that thinks this is fitting), while the value goes to `msgstr`. `msgctxt` isn't used; nor are locations (since those strings aren't really from a particular source file but their own seperate entity; similar to RESX)
   - Comments aren't supported (_yet_), mostly because our own sources are both inconsistent (XML comment before/after the actual element it belongs to) and unspecific (XML comments delimit "blocks" of resources where the developer thought it makes sense to seperate them; rather than documenting one particular resource). I can however add support for those if wanted; and we can agree on where the comments should be to be picked up (my choice would be _before_ the resource element)
- [x] Copyright headers: _Going by the other files; it just goes on the actual sources (not on tests); with creation year and original author_
   - Do they go in all files, with no exceptions; or just on actual workable source files (not on tests)?
   - Is there a copy&paste ready template somewhere, or do I just use one from any other file?
   - Do I have to change anything from there, like the copyright year, names or whatever?
- [x] Documentation files:
   - Are they generated from source (if so: which commands can I use?) or written by hand? Or a combination of both?
   - Should I _assume_ their location and just put an URL in the format documentation comment, or is this done in another way?
- [x] Tests:
   - Need more, want less?
   - Some of the existing test files didn't make sense to me, so I wrote them differently. If that was simply something I didn't see/understand/recognize (which is entirely possible; since Python is far from my main language), please point it out so I can address that.
   - :white_check_mark: ~Find out why Python 3 returns a different encoding. Requires me first fixing my local `virtualenv` to return a working `python3` installation newer than 3.2...~ Looks like the default string encoding was different; added an explicit XML Declaration for that.

And also: Everything else I didn't think about while implementing this :)


translate/storage/flatxml.py
```diff
@@ -0,0 +1,211 @@
...
+        self.element = etree.Element(self.namespaced(self.element_name))
+        super(FlatXMLUnit, self).__init__(source, **kwargs)
+
+    def __eq__(self, other):
+        """Compares two XML units"""
+        if not isinstance(other, FlatXMLUnit):
+            return super(FlatXMLUnit, self).__eq__(other)
+
+        if self.getid() != other.getid():
+            return False
```
nijel: I'm not sure if comparing only by id is good idea. On the other side, I have no clue where this is actually being used...

nijel: It's probably necessary, anyway you might better base your code on LISAunit/LISAfile to get basic xml handling for free...

BhaaLseN: Well, if I remove that comparison, `test_eq` will fail at [`assert unit1 != unit4`](https://github.com/translate/translate/blob/master/translate/storage/test_base.py#L78).

That might be because I don't utilize the passed `source` parameter directly but only hand it to the base class; which writes it to the `source` property which is equal to my id here. I could rewrite it to compare `self.source != other.source` but the effective comparison would be the same.
(same applies to `return self.target == other.target` instead of using `node_text`, because it boils down to the same internally).
And `target` doesn't appear to be set at that location, so the only thing I could compare at this point is `source`/id.

Hence why I added my choice of putting the resource key into `msgid` (which I suppose equals `source`) as discussion point - mostly because I didn't know any better and the resulting PO files looked fine to me.
I'm open for suggestions though if this breaks an assumption anywhere else; or otherwise needs to be handled differently.

BhaaLseN: Just noticed your LISA comment; I deliberately did _not_ base it on LISA because it made a bunch of assumptions and had a few fields that I'd just have to empty out/ignore; so I opted to use the bare metal base class instead.
LISA assumes two levels (`languageNode` and `textNode`) which my format doesn't have (just a single `value_name` element); and I cannot override that from LISA without completely ignoring/rewriting most of the base methods. It is also targetting formats (such as XLIFF) that have base/target languages in them, while my format is monolingual. 
I could rewrite it based on LISA though if you insist (and its entirely possible it does some other stuff I'm not yet aware of; my main use case is to throw it into Weblate).

nijel: Well your format is quite similar to Android one and that already uses LISA, that's why I was thinking it might be way to go.

BhaaLseN: Hm yeah, I noticed that Android had a very similar structure back when I was exploring options before rolling my own; but it had some quirk that wouldn't work for my files (I think it had to do with quoting or something).
It only uses `LISAfile`, but not `LISAunit`; and even for that there seems to be little it actually uses from it. I'll take another look once I got some free time again; probably won't be able to check on it until the weekend.

nijel: Well this is really not something what needs to be fixed, I've just mentioned it as an option to get reasonable compliance with the API without too much effort.

BhaaLseN: I don't think there'd be any benefit if I did.
But since I aliased `source` to `getid` and `target` to `node_text` I can actually remove the `__eq__` implementation in favor of `TranslationUnit.__eq__` (which does exactly that). In fact, I can also remove the `getid`/`setid` overloads completely and just use `source`/`target` for even less code overall.

unho:
I only gave this PR a quick review, but I have some major blockers:

- Format needs documentation (documentation is written by hand)
  - This should include a link to the official format specification
- Converters need documentation (documentation is written by hand)
- Converters need tests in Python, despite having functional tests is great

I must say I am quite pleased to see that the converters are coded the pattern I am trying to introduce for all the converters in TTK, which should allow for easier maintenance and refactoring of common code.

BhaaLseN:
Thanks, I did open this PR early to get this kind of feedback.

I'll look into the documentation soon-ish, and add some more tests for the converters.

BhaaLseN:
Rebased on latest master to get the changes in (especially for the tests that now require the `capsys` parameter for `test_help`).
I noticed a bunch of cleanup commits for the converters, is there anything I should take care of? Style-wise it doesn't seem to be too consistent regarding `inputfile` vs. `input_file` (not even in a po2x/x2po pair).

I also added some documentation, taking some inspiration from other documentation pages. I put the format under "translatable documents" next to HTML, since it seemed most fitting (vs. DTD, which would be be more fitting IMO, but is Mozilla-specific).
Feedback (obviously) appreciated, since I usually don't write end-user focused documentation :S

translate/convert/po2flatxml.py
```diff
@@ -0,0 +1,122 @@
...
+
+        self.value_name = value
+        self.key_name = key
+        self.namespace = ns
+
+        indent_chars = None
+        if indent > 0:
+            indent_chars = (" " * indent)
+
+        self.source_store = factory.getobject(inputfile)
```
unho: Nit: Don't use `factory` since we know that this is a PO file.

translate/convert/po2flatxml.py
```diff
@@ -0,0 +1,122 @@
...
+        self.target_store = flatxml.FlatXMLFile(templatefile,
```
unho: Nit: Don't put the target format's class here, instead define it as a `TargetStoreClass` class attribute and then use the attribute here.

translate/convert/po2flatxml.py
```diff
@@ -0,0 +1,122 @@
...
+                                                root_name=root,
+                                                value_name=value,
+                                                key_name=key,
+                                                namespace=ns,
+                                                indent_chars=indent_chars)
+
+    def convert_unit(self, unit):
+        """Convert a source format unit to a target format unit."""
+        target_unit = flatxml.FlatXMLUnit(source=None,
```
unho: Nit: Don't put the target unit class here, instead specify it as a `TargetUnitClass` class attribute above.

docs/formats/flatxml.rst
```diff
@@ -0,0 +1,96 @@
+.. _flatxml:
+
+Flat XML
+********
+
+The Translate Toolkit is able to process flat XML files using the
+:doc:`flatxml2po </commands/flatxml2po>` converter.
+
+Flat XML (:wp:`eXtensible Markup Language <XML>`) is a simple monolingual
+file format similar to a very basic form of the :doc:`android`.
```
unho: Missing "format" before the period.

unho:
@BhaaLseN I have performed another review (sorry for the delay). This PR has greatly improved but I still found some stuff that needs to be addressed. If you manage to get those nits sorted out this will be a very welcome contribution, thanks!

BhaaLseN:
I'll look into getting them addressed either later today or tomorrow, thanks for getting back to it!

BhaaLseN:
Rebased on latest master and updated the tests to not use the mixed quotes, although #3871 would be a nice addition to use instead of my lazy single-level reindent there.

CI fail seems to be the known (?) segfault, not sure I can do anything about that.

nijel:
I've created separate issue for the test segfaults: https://github.com/translate/translate/issues/new

unho:
@BhaaLseN `master` issues seem to be fixed, can you please rebase this PR again? Thanks.

BhaaLseN:
Rebased. CI seems happy now :)
I also dropped my lazy reindent in favor of `xml_helpers.reindent`.

translate/storage/flatxml.py
```diff
@@ -0,0 +1,201 @@
...
+from translate.storage import base
+from translate.storage.placeables import general
+
+
+class FlatXMLUnit(base.TranslationUnit):
+    """A single term in the XML file."""
+
+    element_name = "str"
+    attribute_name = "key"
+    namespace = None
```
unho: I don't think the three class attributes above are necessary since they are already set per instance in `__init__`.

translate/storage/flatxml.py
```diff
@@ -0,0 +1,201 @@
...
+    rich_parsers = general.parsers
```
unho: Is this unused?

translate/storage/flatxml.py
BhaaLseN: Mostly copy&paste from the beginnings of this PR where I wasn't as familiar with how Python works. I suppose I can just drop them.

translate/storage/flatxml.py
BhaaLseN: As with the other one, I guess this is copy&paste from one of the other formats. I'm not sure if it is used indirectly over the base class, but I'll take a look later.

BhaaLseN: It seems to be used here: https://github.com/translate/translate/blob/4718fabd85a42401f7095569613fa6db9df184e6/translate/storage/base.py#L176-L178
And since ResX (where I got this from) overrides it, I thought this was required. But even with it removed all tests pass just fine.

...which is somewhat unsurprising, not being static there. I'll just remove it.

nijel:

PR #3777, opened 2018-03-12
Windows CI using AppVeyor

nijel:
This is followup to https://github.com/translate/translate/pull/3569 with fixes for the Windows build to make tests pass.

The build now succeeds, see https://ci.appveyor.com/project/nijel/translate/build/1.0.18

I've fixed all failures with exception to oo format, where I was not really sure whether it's test or storage error. Additionally this format doesn't seem to have any upstream docs anymore and is probably dead, so I didn't want to spend much time on that...

To make it usable, registration on AppVeyoyr site is needed and enabling build for this repository. Unfortunately seems to be bound to individual user rather than project there.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,19 @@
+environment:
+  matrix:
+    - PYTHON: "C:\\Python27"
+
+install:
+  # We need wheel installed to build wheels
+  - "%PYTHON%\\python.exe -m pip install wheel"
+  # lxml 3.6.0 has binary release for Windows
+  - "%PYTHON%\\python.exe -m pip install lxml==3.6.0"
+  - "%PYTHON%\\python.exe -m pip install -r requirements\\dev.txt"
+  - "%PYTHON%\\python.exe -m pip freeze"
+
+build: off
+
+test_script:
+  - "%PYTHON%\\python.exe -m compileall -q -f ."
+  - "%PYTHON%\\python.exe setup.py --quiet build"
+  - "%PYTHON%\\python.exe -m pip install ."
+  - "%PYTHON%\\python.exe -m pytest -r EfsxX"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,19 +1,29 @@
 environment:
   matrix:
     - PYTHON: "C:\\Python27"
+    - PYTHON: "C:\\Python34"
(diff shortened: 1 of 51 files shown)


appveyor.yml
```diff
@@ -0,0 +1,31 @@
...
+    - PYTHON: "C:\\Python35"
+    - PYTHON: "C:\\Python36"
+    - PYTHON: "C:\\Python36-x64"
+
+install:
+  - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  # We need wheel installed to build wheels
+  - "python.exe -m pip install wheel"
+  # lxml 4.1.1 has binary release for Windows
+  - "python.exe -m pip install lxml==4.1.1"
```
nschonni: I wasn't sure if there was a better way to do this. Did a quick search and found https://gist.github.com/pombredanne/72130ee6f202e89c13bb but that might be overkill

nijel: Well it's possible to use this directly in requirements.txt:

```
lxml==4.1.1; platform_system == 'Windows'
```

The question is whether it's better approach than specifying it for tests...

nschonni: I have been running Weblate on Windows, and I can confirm there is a few hoops like this required. This particular one might be more for avoiding a manual compile that requires a large toolchain and matching Python VC++ tools install

nijel: Anyway it seems they are consistently doing binary releases now, so it's probably best to remove this completely.

translate/convert/test_po2oo.py
```diff
@@ -49,6 +49,8 @@ def check_roundtrip(self, oosource):
 
     def test_convertoo(self):
         """checks that the convertoo function is working"""
+        if os.name == 'nt':
+            skip("test or storage broken on Windows")
```
unho: Better use `skipif`. Don't worry, I will apply this change myself after me merging this PR.

unho:
Great contribution @nschonni and @nijel. I've merged it to `master`. Since PR was not on top of latest `master` I have to manually close this PR:

unho:
@nijel Regarding OO format we should probably drop it in the future since it is a format that won't come back, it will surely become a burden to maintain it, and people needing to convert it will still be able to use older TTK versions.

PR #3785, opened 2018-03-19
YAML: Consistent handling of int and bool

### LINKED ISSUE #3619: OrderedDictYAMLLoader fails on input yaml with numbers (similar as #3363)
As can be seen in https://github.com/translate/translate/blob/master/translate/storage/yaml.py#L188 , the OrderedDictYAMLLoader raises an exception (that currently isn't catched by Weblate) on integer inputs in a Ruby YAML file.

The integers should be used as-is, or just ignored.

nijel:
We convert both to string on loading and store them as string.

Fixes #3619

### DIFF
--- a/translate/storage/test_yaml.py
+++ b/translate/storage/test_yaml.py
@@ -144,7 +144,6 @@ eggs: spam
 eggs: spam
 '''
 
-    @pytest.mark.xfail(reason="Not Implemented")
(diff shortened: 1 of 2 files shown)


unho:
lgtm @nijel Thanks for the contribution.

nijel:
Thanks for review.

PR #3786, opened 2018-03-19
PHP: Improve handling of unit name

nijel:
Consistently use lexer to extract name as is used in the original file
instead of relying on parser to return the exact way.


unho:
lgtm. Thanks for the contribution.

nijel:
Thanks for review.

PR #3787, opened 2018-03-20
TMX: Avoid mentioning po2tmx in creation tool

nijel:
The file can be created by anything using Translate Toolkit, so
mentioning po2tmx might be confusing and makes people think there is
some po file existing for the data.


nijel:
This is based on feedback I got for TMX file generated by https://github.com/nijel/l10n-slovnik.

unho:
lgtm.

https://github.com/nijel/l10n-slovnik is nice, but in the Galician community we are trying to move the other way TBX->HTML. Anyway this is probably a tool that a lot of communities will use, specially those that have little expertise but know how to work using a wiki.

nijel:
Thanks for review. 

As for the script it really has no ambition of general usage, parsing wiki markup is a bit tricky. It was written few years ago and now just only got to publish it...

PR #3789, opened 2018-03-21
Fix tests for po2sub to match method name change

llimeht:
In commit 877cab0a, `convertstore` was turned into `convert_store` but the test was not updated.

(the po2sub/sub2po tests are not run at all on travis-ci because aeidon is not installed on travis; this probably should be remedied)

### DIFF
--- a/translate/convert/test_po2sub.py
+++ b/translate/convert/test_po2sub.py
@@ -19,7 +19,7 @@ class TestPO2Sub(object):
         inputfile = wStringIO.StringIO(posource)
         inputpo = po.pofile(inputfile)
         convertor = po2sub.po2sub()
-        outputsub = convertor.convertstore(inputpo)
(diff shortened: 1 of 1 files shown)


unho:
@llimeht Thanks for the contribution. I will try to have a look into enabling those tests since it looks like `aeidon` has support for Python 3.

unho:
I have enabled the tests for Python 3, but I had no luck yet in enabling them for Python 2.

llimeht:
Indeed, aeidon is dead for Python 2 so I think they just have to be omitted. (Our Debian packages have just omitted this no-longer-existing module for the Python 2 modules and only built/tested/depended on the Python 3 version for a while now.)

unho:
The code kinda relies on `gaupol` for Python 2 support, but it is not available on PyPI so we can't use it.

PR #3791, opened 2018-03-23
INI: Fix ini for Python 3

nijel:
There is no official support in iniparse upstream, but for example
Debian ships patched iniparse version to work with Python 3.

### DIFF
--- a/translate/convert/ini2po.py
+++ b/translate/convert/ini2po.py
@@ -23,6 +23,8 @@ See: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/command
 for examples and usage instructions.
 """
 
+import sys
+
 from translate.convert import convert
 from translate.storage import ini, po
 
@@ -38,10 +40,8 @@ class ini2po(object):
                  blank_msgstr=False, duplicate_style="msgctxt",
                  dialect="default"):
         """Initialize the converter."""
-        import sys
-        if sys.version_info[0] == 3:
-            print("Translate Toolkit doesn't yet support converting from INI in "
-                  "Python 3.")
+        if ini.INIConfig is None:
+            print("Missing iniparse library!")
             sys.exit()
 
         self.blank_msgstr = blank_msgstr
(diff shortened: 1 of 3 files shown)


nijel:
Note that this actually doesn't test ini support on Python 3 in the CI as there is no official package for iniparse supporting Python 3 in the PyPI...

unho:
The ideal solution would be to get that Debian patched version to be published in PyPI and then actually use it. @llimeht or @nijel do you know where that Debian patched version of `iniparse` is?

Anyway the PR is lgtm.

llimeht:
The iniparse patch started at https://code.google.com/p/iniparse/issues/detail?id=22 and was incorporated into the Debian package at https://anonscm.debian.org/cgit/collab-maint/python-iniparse.git/commit/?id=fef9b4532a1a6a03aab82711f52ebde3f2d131d7 (there's also one additional patch following that one to disable some broken tests).

I found one possible new upstream for a Python 3 fork (but that module structure looks wrong and there doesn't appear to be any activity really so I don't think that's viable https://github.com/zeyosinc/iniparse).

Basically, iniparse is dead upstream no new upstream has appeared and every distro has adopted more or less the same patch as a workaround for that.

unho:
So we just have to create a new github repo to have `iniparse` with that patch (so people can refer to it since upstream looks dead), and then create a package and publish it on PyPI. Once that is done we can update TTK requirements to use this new PyPI package and test whether it works in both Python 2 and 3.

@llimeht But before going for that I would like confirmation that the patch actually provides support for both Python 2 and Python 3.

nijel:
Thanks for review, I've merged the changes...

llimeht:
@unho, the test suite passes for both Python 2 and Python 3 using this patched version of iniparse: https://ci.debian.net/data/autopkgtest/unstable/amd64/t/translate-toolkit/102970/log.gz

You can't hijack that namespace on pypi so the old Python 2-only version will continue to be called iniparse there and you'd have to have a new name for it on pypi. I guess you should also consider whether you really want to become upstream for file format modules...

nijel:
Alternatively there is also [inifile](https://pypi.python.org/pypi/inifile) library, though I haven't checked how does it compare (I've noticed it because of ITP for Debian).

unho:
I've found this other option https://pypi.python.org/pypi/fileconfig When I have time I will investigate which way to go.

PR #3792, opened 2018-03-24
[WIP] Deprecate setsource and getsource methods in favor of source property

unho:


translate/storage/cpo.py
```diff
@@ -377,7 +379,11 @@ def setsource(self, source):
         else:
             gpo.po_message_set_msgid(self._gpo_message, gpo_encode(source))
             gpo.po_message_set_msgid_plural(self._gpo_message, None)
-    source = property(getsource, setsource)
+
+    # Deprecated on 2.3.1
+    @deprecated("Use `source` property instead")
+    def getsource(self):
+        return self.source
```
nijel: Isn't is enough to have this on base.TranslationUnit?

unho: Not all the subclasses of `base.TranslationUnit` have the `getsource` and I didn't want to add a new deprecated `getsource` method to those.

unho:
Note we have to keep around the `setsource` method for some XML-based storage classes because it adds an extra argument for allowing specifying the language (those are multilanguage stores), but method wasn't working for the property anyway (you can't have a property setter that accepts more than one value AFAICT).

unho:
I am merging this. I am starting now a similar work regarding the `target` property.

PR #3797, opened 2018-04-07
Deprecate `gettarget` and `settarget` in favor of `target` property

### LINKED ISSUE #3795: Use target property rather than gettarget and settarget methods
Use `target` property rather than `gettarget` and `settarget` methods which are then used to create the `target` property. This should simplify the API exposed.

unho:
Fixes #3795.

### DIFF
--- a/translate/storage/resx.py
+++ b/translate/storage/resx.py
@@ -58,14 +58,16 @@ class RESXUnit(lisa.LISAunit):
     def getsource(self):
         return self.source
 
-    def gettarget(self):
+    @property
+    def target(self):
         targetnode = self._gettargetnode()
         if targetnode is None:
             etree.SubElement(self.xmlelement, self.namespaced("value"))
             return None
         return data.forceunicode(targetnode.text) or u""
 
-    def settarget(self, target):
+    @target.setter
+    def target(self, target):
         # Firstly deal with reinitialising to None or setting to identical
         # string.
         self._rich_target = None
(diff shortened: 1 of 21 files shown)


translate/storage/test_xliff.py
```diff
@@ -321,7 +321,7 @@ def test_fuzzy(self):
         unit.markfuzzy()
         assert not unit.isfuzzy()  # untranslated
         unit.target = "Konsep"
-        assert unit.isfuzzy()
+        assert not unit.isfuzzy()  # translated
```
nijel: This test is a bit confusing, but the new behavior looks more correct to me. Hopefully it won't break anything else.

unho: Yes, I think previous behavior was broken. It doesn't make that if you set a translation the unit is fuzzy automagically.

translate/storage/resx.py
```diff
@@ -58,14 +58,16 @@ def source(self):
...
+    def target(self):
         targetnode = self._gettargetnode()
         if targetnode is None:
             etree.SubElement(self.xmlelement, self.namespaced("value"))
             return None
         return data.forceunicode(targetnode.text) or u""
 
-    def settarget(self, target):
+    @target.setter
+    def target(self, target):
```
nijel: As spotted in https://github.com/WeblateOrg/weblate/commit/c79304c2745c16661a13d1a900c09c621e87ab44#commitcomment-28783478, this breaks compatibility. Anybody who relied on settarget will not get proper behavior here as settarget from LISA will be called instead of customized behavior here.

nijel: PR for this is here: https://github.com/translate/translate/pull/3806

PR #3801, opened 2018-04-12
Improve handling of return and array in PHP

nijel:
Improve handling of some corner cases in PHP:

* Correctly serialize [] style arrays
* Correctly parse and serialize return statements


unho:
My only objection is that docs are not updated despite a new feature of the format is supported now.

nijel:
Both are already covered by the docs. These were working for parsing before, this PR just fixes load/save round trip.

unho:
lgtm

nijel:
Thanks for review.

PR #3802, opened 2018-04-16
Further PHP fixes

nijel:
- Fix serializing of `return []` statement
- Fixed using invalid escape sequences in tests


unho:
lgtm

PR #3803, opened 2018-04-21
resx: remove unused variable

BhaaLseN:
`strings` is only written to, but never read anywhere. Seemed to be in there since ResX was first committed, but has no effect whatsoever. Unless I missed some pythonic side-effect that isn't obvious.


unho:
@BhaaLseN Really nice one! Merged. Thanks!

PR #3804, opened 2018-04-21
Rich target cleanups

unho:


unho:
@nijel Any review will be appreciated.

translate/storage/xliff.py
```diff
@@ -244,7 +244,14 @@ def get_rich_target(self, lang=None):
...
         return self._rich_target
-    rich_target = property(get_rich_target, set_rich_target)
+
+    @property
+    def rich_target(self):
+        return self.get_rich_target()
+
+    @rich_target.setter
+    def rich_target(self, value):
+        self.set_rich_target(value)
```
nijel: Shouldn't this rather deprecate `get_rich_target`/`set_rich_target` and move the implementation to the properties?

unho: Note that these two methods in `xliff` storage allow an extra `lang` argument. Such argument is not used when you use the property, but can be specified if you call those methods directly, which is the reason why I haven't removed them.

nijel: Good point, I didn't notice this.

unho: @nijel Any objection for this PR to get landed?

PR #3806, opened 2018-04-29
RESX: Add settarget compatibily

nijel:
The settarget is deprecated since 2.3.1, but it might still be used and
should behave properly. Without method here, the one from LISAunit gets
called and it behaves completely differently.


nijel:
I've rebased the code on current master and added testcase.

unho:
lgtm. Thanks for your contribution and sorry for me reviewing it so late.

PR #3807, opened 2018-04-29
resx: Properly indent new entries

BhaaLseN:
Yet another change to make ResX play nicer with VCS.

Before this, adding new units to a file (ie. when a source resx has the entry, but the target resx does not) would put the new `<data>` entry where the closing `</root>` was (not indented at all), and the new `</root>` indented as if it was `<data>`.
```diff
--- a/Resources.de.resx
+++ b/Resources.de.resx
@@ -126,4 +126,7 @@
   <data name="ExistingEntry" xml:space="preserve">
     <value>Bestehender Eintrag</value>
   </data>
-</root>
+<data name="NewEntry" xml:space="preserve">
+    <value>Neuer Eintrag</value>
+  </data>
+  </root>
```
After the change, only the newly added `<data>` is part of the diff (as it should be), leaving the surrounding tags alone.
```diff
--- a/Resources.de.resx
+++ b/Resources.de.resx
@@ -126,4 +126,7 @@
   <data name="ExistingEntry" xml:space="preserve">
     <value>Bestehender Eintrag</value>
   </data>
+  <data name="NewEntry" xml:space="preserve">
+    <value>Neuer Eintrag</value>
+  </data>
 </root>
```
Visual Studio would obviously fix those, but this only leads to unnecessary back and forth bloating diffs and making reviews of actual changes a little harder.

That should take care of the most common issues causing conflicts for us, but I'm pretty sure there are other corner-cases that do not behave corretly (in part due to how lxml handles pretty-printing and the format attempts to fix those issues by setting `tail`/`text` on the elements by hand); but it's a little hard to fix an issue that isn't known to me, so there might be follow-up PRs addressing similar formatting issues.

### DIFF
--- a/translate/convert/test_po2resx.py
+++ b/translate/convert/test_po2resx.py
@@ -446,8 +446,8 @@ msgstr "Bézier-kurwe"
         assert resx_file == expected_output
 
 
-class TestPO2TSCommand(test_convert.TestConvertCommand, TestPO2RESX):
-    """ Tests running actual po2ts commands on files """
+class TestPO2RESXCommand(test_convert.TestConvertCommand, TestPO2RESX):
+    """ Tests running actual po2resx commands on files """
     convertmodule = po2resx
 
     def test_help(self, capsys):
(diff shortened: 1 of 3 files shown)


nijel:

PR #3809, opened 2018-05-06
pypo: Fix parsing of files with mixed newlines

nijel:
In case \r is included in comment text and whole file is having \n
newlines, the univerzal newlines approach breaks lines on wrong
locations, while gettext handles this just fine.

This addresses regression introduced by https://github.com/translate/translate/pull/3760, see https://github.com/WeblateOrg/weblate/issues/1986#issuecomment-386704921

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -48,6 +48,29 @@ po_unescape_map = {"\\r": "\r", "\\t": "\t", '\\"': '"', '\\n': '\n', '\\\\': '\
 po_escape_map = dict([(value, key) for (key, value) in po_unescape_map.items()])
 
 
+def splitlines(text):
+    """Split lines based on first newline char.
+
+    Can not use univerzal newlines as they match any newline like
+    character inside text and that breaks on files with unix newlines
+    and LF chars inside comments.
+    """
+    # Find first newline
+    newline = b'\n'
+    for i, ch in enumerate(text):
+        # Iteration over bytes yields numbers in Python 3
+        if ch in ('\n', 10):
+            break
+        if ch in ('\r', 13):
+            if text[i + 1] in ('\n', 10):
+                newline = b'\r\n'
+            else:
+                newline = b'\r'
+            break
+
+    return [x + newline for x in text.split(newline)]
+
+
 def escapeforpo(line):
     """Escapes a line for po format. assumes no \n occurs in the line.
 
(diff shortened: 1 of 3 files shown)


translate/storage/test_pypo.py
```diff
@@ -387,3 +387,10 @@ def test_mac_newlines(self):
         pofile = self.poparse(posource)
         assert len(pofile.units) == 1
         assert pofile.units[0].source == 'test me'
+
+    def test_mixed_newlines(self):
+        """checks that mixed newlines are properly parsed"""
+        posource = b'#Comment\n\n#: foo.c:124\r bar.c:124\r\nmsgid "test me"\nmsgstr ""\n'
```
unho: Please use triple quoted strings for holding the source in tests. It makes it easier to read, spot errors, fix them, and to copy and paste.

translate/storage/test_pypo.py
```diff
@@ -387,3 +387,10 @@ def test_mac_newlines(self):
...
+        pofile = self.poparse(posource)
+        assert len(pofile.units) == 1
+        assert pofile.units[0].source == 'test me'
```
unho: Is roundtrip tested somehow? If not, please add a roundtrip test.

translate/storage/pypo.py
```diff
@@ -48,6 +48,29 @@
...
+    character inside text and that breaks on files with unix newlines
+    and LF chars inside comments.
+    """
+    # Find first newline
+    newline = b'\n'
+    for i, ch in enumerate(text):
+        if ch in ('\n', 10):
+            newline = b'\n'
+            break
+        elif ch in ('\r', 13):
```
unho: Nit: Use `if` instead of `elif`.

translate/storage/pypo.py
```diff
@@ -48,6 +48,29 @@
...
+
+    Can not use univerzal newlines as they match any newline like
+    character inside text and that breaks on files with unix newlines
+    and LF chars inside comments.
+    """
+    # Find first newline
+    newline = b'\n'
+    for i, ch in enumerate(text):
+        if ch in ('\n', 10):
+            newline = b'\n'
```
unho: This line is unnecessary because this value is already set before the `for` loop.

translate/storage/pypo.py
```diff
@@ -48,6 +48,29 @@
...
+    """Split lines based on first newline char.
+
+    Can not use univerzal newlines as they match any newline like
+    character inside text and that breaks on files with unix newlines
+    and LF chars inside comments.
+    """
+    # Find first newline
+    newline = b'\n'
+    for i, ch in enumerate(text):
+        if ch in ('\n', 10):
```
unho: When can `ch` be `10`?

nijel: It is always number in Python 3 (iteration over bytes yields numbers, not chars).

unho:
lgtm. Thanks for your contribution!

Issue #3815, opened 2018-06-28
Please create a new release tag


sotathan:
Fixes [b92873d4](https://github.com/translate/translate/commit/b92873d4c212d959951e37bba5f02c4bc5021c09) and [6289ea55](https://github.com/translate/translate/commit/6289ea5533b63fb0e7cdc4249d1a601af6b98fee) are essential to our project.

Could you please create a new tag.

Thanks,
Sotiris

unho:
@sotathan Translate Toolkit 2.3.1 has just been released.

sotathan:
Thanks @unho !!!

PR #3816, opened 2018-07-22
YAML: Fix serialization of empty keys

nijel:
See https://github.com/WeblateOrg/weblate/issues/2128

### DIFF
--- a/translate/convert/test_po2yaml.py
+++ b/translate/convert/test_po2yaml.py
@@ -130,6 +130,10 @@ msgstr ""
 msgid "booo"
 msgstr ""
 
+#: foo->
+msgid "bar2"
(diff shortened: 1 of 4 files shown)


unho:
What happens if there are multiple empty keys?

Can you also please add tests for the converters?

nijel:
Multiple same keys are invalid in YAML, so it would fail at parser.

nijel:
I've just added empty keys to the converter tests.

unho:
lgtm @nijel Please land.

nijel:
Thanks for review. Any plans for new release? There are quite a lot fixes accumulated in the master, also see https://github.com/translate/translate/issues/3815

PR #3817, opened 2018-07-22
JSON: Preserve ordering of nested values

nijel:
We need to use OrderedDict to ensure the ordering of values is kept
while saving.


unho:
lgtm. Thanks for you contribution @nijel, please land it.

PR #3822, opened 2018-08-25
Catkeys: merge header and fingerprint

nielx:
The catkeys format now has support for fingerprint calculation.
This means that if the tools change the contents of the file (like
remove or change strings), the header will now be correct and the file
will be parseable on Haiku.

When merging with pot2po, it will use the template header to determine
the language of the output format.

### DIFF
--- a/translate/storage/catkeys.py
+++ b/translate/storage/catkeys.py
@@ -113,7 +113,11 @@ class CatkeysHeader(object):
         if not header:
             self._header_dict = self._create_default_header()
         elif isinstance(header, dict):
-            self._header_dict = header
+            for key in FIELDNAMES_HEADER:
+                value = header.get(key, "")
+                if value is None:
+                    value = ""
+                self._header_dict[key] = value
 
     def _create_default_header(self):
         """Create a default catkeys header"""
@@ -127,6 +131,12 @@ class CatkeysHeader(object):
         #XXX assumption about the current structure of the languages dict in data
         self._header_dict['language'] = data.languages[newlang][0].lower()
     targetlanguage = property(None, settargetlanguage)
+
+    def setchecksum(self, checksum):
+        """Set the checksum for the file"""
+        if not checksum:
+            return
+        self._header_dict['checksum'] = str(checksum)
 
 
 @six.python_2_unicode_compatible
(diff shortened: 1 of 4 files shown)


nijel:
Can you please add some tests for the fingerprint handling?

translate/storage/catkeys.py
```diff
@@ -128,6 +132,13 @@ def settargetlanguage(self, newlang):
         self._header_dict['language'] = data.languages[newlang][0].lower()
     targetlanguage = property(None, settargetlanguage)
 
+    def setchecksum(self, checksum):
+        """Set the checksum for the file"""
+        if not checksum:
+            return
+        self._header_dict['checksum'] = str(checksum)
+    checksum = property(None, setchecksum)
```
unho: I don't see the need for this to be a property since we can just live with the `set_checksum` method.

translate/storage/catkeys.py
```diff
@@ -283,8 +299,47 @@ def parse(self, input):
...
         for unit in self.units:
             writer.writerow(unit.dict)
         out.write(output.getvalue().encode(self.encoding))
+
+    def _compute_fingerprint(self):
+        """
+        Compute the current hash key in the header for the current state of the store.
+
+        Inner function hashfun() based on CatKey::HashFun()
+        https://github.com/haiku/haiku/blob/b65adbdfbc322bb7d86d74049389c688e9962f15/src/kits/locale/HashMapCatalog.cpp#L93
```
unho: Nit: this last part of the docstring should be a docstring in the `hashfun` function below.

unho:
I must concur with @nijel in that tests are needed.

Also please add something about this feature to the catkeys format documentation.

nielx:
I will work on the patch the coming week. Thanks for the feedback.

unho:
@nielx It would be great if you can finish this PR.

nielx:
It has been a long time coming, but I have added a test.

translate/storage/catkeys.py
nielx: Done (to be consistent with underscore usage, I have named it setchecksum)

nijel:
Thanks @nielx, looks good now. Any chance to describe this in the docs as well (docs/formats/catkeys.rst)?

nielx:
I added some additional information in the documentation.

nijel:

Issue #3824, opened 2018-09-04
Read OS X multiline strings bug.


dhontecillas:
There is a bug when reading multiline OS X strings. 

In the file: `storage/properties.py` , the `stringsfile` and `stringsutf8file` classes derive from the `properties` class. 

It looks like **.properties** files can have multiline strings but using a backslashes `\` (or that is what I understood looking at the code), but OS X string files do not require that. Instead, to check for the end of the string the last character, that should be a ';'. 

In order to import those OS X Files I had to make a patch, that worked for me (but would need careful review, and unit tests to make it a valid PR)

```diff
diff --git a/translate/storage/properties.py b/translate/storage/properties.py
index a0d71f57..f6b67b00 100644
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -609,6 +609,9 @@ class propfile(base.TranslationStore):
             self.parse(propsrc)
             self.makeindex()
 
+    def is_line_continuation(self, line):
+        return is_line_continuation(line)
+
     def parse(self, propsrc):
         """Read the source of a properties file in and include them as units.
         """
@@ -631,7 +634,7 @@ class propfile(base.TranslationStore):
             if inmultilinevalue:
                 newunit.value += line.lstrip()
                 # see if there's more
-                inmultilinevalue = is_line_continuation(newunit.value)
+                inmultilinevalue = self.is_line_continuation(newunit.value)
                 # if we're still waiting for more...
                 if inmultilinevalue:
                     # strip the backslash
@@ -666,7 +669,7 @@ class propfile(base.TranslationStore):
                     newunit = propunit("", self.personality.name)
                 else:
                     newunit.name = self.personality.key_strip(line[:delimiter_pos])
-                    if is_line_continuation(line[delimiter_pos+1:].lstrip()):
+                    if self.is_line_continuation(line[delimiter_pos+1:].lstrip()):
                         inmultilinevalue = True
                         newunit.value = line[delimiter_pos+1:].lstrip()[:-1]
                     else:
@@ -712,6 +715,12 @@ class stringsfile(propfile):
         kwargs['personality'] = "strings"
         super(stringsfile, self).__init__(*args, **kwargs)
 
+    def is_line_continuation(self, line):
+        l = line.rstrip()
+        if l and l[-1] == ';':
+            return False
+        return True
+
 
 class stringsutf8file(propfile):
     Name = "OS X Strings (UTF-8)"
[...]

PR #3825, opened 2018-09-09
fix import OS X strings file

dhontecillas:
This fixes: https://github.com/translate/translate/issues/3824

There is a bug when reading multiline OS X strings.

In the file: storage/properties.py , the stringsfile and stringsutf8file classes derive from the properties class.

It looks like .properties files can have multiline strings but using a backslashes \ (or that is what I understood looking at the code), but OS X string files do not require that. Instead, to check for the end of the string the last character, that should be a ';'.

Could not make the tests in the master branch pass, and did not understand how to create a new test to raise the issue (in an initial commit) and then commit the fix. So, I know that this is not a good enough PR, but if you try to import the OS X strings file that I attach, you will find it fails.

[os_x_strings_sample.zip](https://github.com/translate/translate/files/2364438/os_x_strings_sample.zip)

After applying this modification, the OS X strings file is imported as expected.

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -328,6 +328,14 @@ class Dialect(object):
         """Strip unneeded characters from the value"""
         return value.lstrip()
 
+    @classmethod
+    def is_line_continuation(cls, line):
+        return is_line_continuation(line)
+
+    @classmethod
+    def strip_line_continuation(cls, value):
+        return value[:-1]
+
 
 @register_dialect
 class DialectJava(Dialect):
@@ -418,6 +426,17 @@ class DialectStrings(Dialect):
     @classmethod
     def encode(cls, string, encoding=None):
         return string.replace("\n", r"\n").replace("\t", r"\t")
+
+    @classmethod
+    def is_line_continuation(self, line):
+        l = line.rstrip()
+        if l and l[-1] == ';':
+            return False
+        return True
+
+    @classmethod
+    def strip_line_continuation(cls, value):
+        return value
 
 
 @register_dialect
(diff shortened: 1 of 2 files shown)


translate/storage/properties.py
```diff
@@ -722,6 +731,12 @@ def __init__(self, *args, **kwargs):
         kwargs['encoding'] = "utf-8"
         super(stringsutf8file, self).__init__(*args, **kwargs)
 
+    def is_line_continuation(self, line):
+        l = line.rstrip()
+        if l and l[-1] == ';':
+            return False
+        return True
+
```
nijel: I don't like the code duplication here, I think this class should inherit from stringsfile to avoid that.

dhontecillas: Indeed. 
But I will have to change `stringsfile`'s __init__ implementaion because it overrides the given `kwarg['personality']` param.

nijel: True, some changes to __init__ will be needed as well.

dhontecillas: PTAL
Tell me if there is something else that should be changed.

(I did not put the `def is_line_continuation` function inside the `propfile` class, to avoid breaking other code that might depend on this function).

nijel:
I think it's okay this way. Can you please also add tests for this change? They are located in `translate/storage/test_properties.py`.

dhontecillas:
Hi Nijel, 

sorry for the delay.

after adding a test for "multi line" Mac OS X strings in files, I found it was better to move the `is_line_continuation` to the `Dialect` class (and `DialectStrings`).

Is a little bit misleading having two ways of reading the mac os x strings file:
- instantiating a `propfile` class of 'strings' personality (as it is done in the tests).
- instantiating a `stringsfile` class (that only overrides params passed to the `propfile`).

My previous "fix" was only working with the second approach (and since weblate worked ok with that, I guess is how it uses it). 

Also fixed a test, (the multiline comments one), because it was missing the `;` at the end of the string.

So, please take a look at the code, and lets see if it is ready to be merged.

translate/storage/properties.py
```diff
@@ -709,11 +733,12 @@ class stringsfile(propfile):
     Extensions = ['strings']
 
     def __init__(self, *args, **kwargs):
-        kwargs['personality'] = "strings"
+        if not kwargs.get('personality'):
+            kwargs['personality'] = "strings"
```
nijel: This change is probably not needed now when  the code has been moved to the dialect...

nijel:
And yes, the whole dialects thing is a bit confusing, I don't really know original reasoning for that.

translate/storage/properties.py
dhontecillas: Yep! this is a leftover from my previous approach. Will remove and squash.

dhontecillas:
Now I think it is ready :)

nijel:

dhontecillas:
You are welcome :)

PR #3830, opened 2018-09-25
Update URL on #121

kant:


unho:
Merged as https://github.com/translate/translate/commit/c9a7cd5177e46cbfb1dc0d0967333495890aaf53 after adjusting the commit message.

Issue #3819, opened 2018-08-05
Better handle nested keys


nijel:
Currently there are several ways in translate-toolkit to handle nested IDs in translation format.

What I know is that:

* XLIFF uses `u"\04"`
* YAML uses `->`
* JSON uses `.`

All these are problematic as they can as well occur in the actual value and leads to possible problems when serializing. Only XLIFF seems to care about escaping in this case.

It would be great if this would be consistent and error prone, but any change can lead to problems in convertors as they would now generate different IDs.

In the end I think using `u"\04"` with proper escaping should be fine. Another option is to use tuple as ID in this case.

Related code snippets:

https://github.com/translate/translate/blob/master/translate/storage/xliff.py#L498-L509

https://github.com/translate/translate/blob/master/translate/storage/jsonl10n.py#L184-L195

https://github.com/translate/translate/blob/master/translate/storage/yaml.py#L177-L194

unho:
This is a very nice one. Thanks for raising it!

nijel:
The similar issue is with lists - they are encoded as `key[index]`, but the same value can be used as actual key in the files.

nijel:
@unho I was thinking how to address this and I think that using strings to encode this is not really a good idea. What about introducing following objects to track this:

```python
class UnidId(object):
    def __init__(self, *items):
        self.items = items

    def __str__(self):
        return ''.join([str(item) for item in items])

class UnitIdKey(object):
    def __init__(self, value):
        self.value = value

    def __str__(self):
        return '.{}'.format(self.value)

class UnitIdList(object):
    def __init__(self, value):
        self.value = value

    def __str__(self):
        return '[{}]'.format(self.value)
```

It will work as strings (for the converters into formats which do not support this), but will keep the detailed information for the ones which do.

It would be used by something like this:

```python
def getid():
    return UnitId(UnitKey('hello'), UnitIdList(1))
```

What do you think about this approach?

unho:
I am not sure I understand. Can you please share an example JSON (or PHP) and explain how this would work?

I am most concerned about if this would break backwards compatibility, and how.

unho:
Also unrelated to all of this I would really like to get rid of `setid` and `getid` in favor of a `unit_id` property, if possible.

nijel:
[...]

PR #3835, opened 2018-10-17
WIP: JSON: Use structured unit IDs

nijel:
This allows to preserve structure of more complex documents and removes
difference between nested and non nested files.

This is proof of concept implementation for https://github.com/translate/translate/issues/3819, not ready for merge right now.

### DIFF
--- a/translate/convert/po2json.py
+++ b/translate/convert/po2json.py
@@ -40,7 +40,7 @@ class rejson(object):
         self.remove_untranslated = remove_untranslated
         self.inputstore.makeindex()
         for unit in self.templatestore.units:
-            inputunit = self.inputstore.locationindex.get(unit.getid())
+            inputunit = self.inputstore.locationindex.get(str(unit.getid()))
             skip_unit = (self.remove_untranslated and
                          (inputunit is None or
                           inputunit.isfuzzy() or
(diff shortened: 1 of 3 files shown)


nijel:
Caveats:

* Apparently it's not possible without changing convertors
* Anybody expecting unit id to be a string might need to update the code
* Currently it will break on unicode keys on Python 2, but that can be addressed if we decide to go this way

translate/convert/po2json.py
```diff
@@ -40,7 +40,7 @@ def convertstore(self, includefuzzy=False, remove_untranslated=False):
         self.remove_untranslated = remove_untranslated
         self.inputstore.makeindex()
         for unit in self.templatestore.units:
-            inputunit = self.inputstore.locationindex.get(unit.getid())
+            inputunit = self.inputstore.locationindex.get(str(unit.getid()))
```
unho: Can't `unit.getid()` be altered to return a string?

nijel: That is certainly an option, this was mostly posted to get such kind of feedback - where to put the abstraction.

unho: Please note I've left a more elaborated comment on https://github.com/translate/translate/issues/3819#issuecomment-449043811

nijel:
Closing in favor of https://github.com/translate/translate/pull/4091

nijel:
Closing in favor of https://github.com/translate/translate/pull/4091

PR #3836, opened 2018-10-21
JSON: Fix serializing of JSON arrays

nijel:
They are properly parsed, but the structure was not restored on
serializing.

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -160,8 +160,13 @@ class JsonFile(base.TranslationStore):
     def serialize(self, out):
         def merge(d1, d2):
             for k in d2:
-                if k in d1 and isinstance(d1[k], dict) and isinstance(d2[k], dict):
-                    merge(d1[k], d2[k])
+                if k in d1:
+                    if isinstance(d1[k], dict) and isinstance(d2[k], dict):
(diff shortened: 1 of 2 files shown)


nijel:
This is in more general addressed by #3835, but this patch is simple and can be merged faster than complex change introduced there.

translate/storage/jsonl10n.py
```diff
@@ -232,6 +237,9 @@ def getkey(self):
     def getvalue(self):
         ret = self.converttarget()
         for k in reversed(self.getkey()):
+            if '[' in k and k[-1] == ']':
+                k = k.split('[')[0]
```
nschonni: Anything odd going to happen if it's an empty array like `[]`?

nijel: Well the keys are ignored anyway, with assumption that the list is constructed in order by units, so it will work just fine.

Yes, it can result in storing something else than was parsed, but that's how the JSON storage currently works (see https://github.com/translate/translate/issues/3819 and https://github.com/translate/translate/pull/3835).

nijel: IOW: It will certainly break in some cases, but I believe it works better than current code. Some solution like outlined in #3835 is needed to make it work reliably in all corner cases.

PR #3838, opened 2018-11-10
Fix + enhance web2py <-> po translation

### LINKED ISSUE #3254: Couple issues with Web2py converter
I just tried using the Web2py converter and it seems to have a couple of issues:
- At first it wouldn't even run because of an unexpected keyword argument (duplicates)
- After changing that to accept **kwargs, I found that it choked on utf-8 strings (in my case, with Umlauts).

In any case, I have made a couple changes and now it works, at least for me - but since I haven't done any proper testing, I don't feel like creating an actual pull request. On the other hand, maybe somebody else can take advantage of this, so here's a gist with the changes I made to web2py2po.py
https://gist.github.com/pgcd/52338762bdcb4728cd39

vinyldarkscratch:
This commit includes numerous updates and bug fixes that allow for the use of the web2py <-> po converter commands:
- Fixes runtime for the converters to properly function (previously throwing errors)
- Includes Unicode support
- Utilizes the standard function calls to add po unit data
- Alphabetically sorts entries in the new web2py file
- Doesn't write the file if contents will be empty
- Fixes whitespace and other formatting
- Adds test for po2web2py

Fixes #3254.

### DIFF
--- a/translate/convert/po2web2py.py
+++ b/translate/convert/po2web2py.py
@@ -24,10 +24,12 @@ See: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/command
 for examples and usage instructions.
 """
 
-from io import BytesIO
+from io import StringIO
 
 from translate.convert import convert
 from translate.storage import factory
+
+import six
 
 
 class po2pydict(object):
@@ -36,7 +38,7 @@ class po2pydict(object):
         return
 
     def convertstore(self, inputstore, includefuzzy):
-        str_obj = BytesIO()
+        str_obj = StringIO()
 
         mydict = dict()
         for unit in inputstore.units:
@@ -46,13 +48,15 @@ class po2pydict(object):
                 mydict[unit.source] = unit.target
             else:
                 mydict[unit.source] = unit.source
-                # The older convention is to prefix with "*** ":
-                #mydict[unit.source] = '*** ' + unit.source
 
-        str_obj.write('{\n')
-        for source_str in mydict:
-            str_obj.write("%s:%s,\n" % (repr(str(source_str)), repr(str(mydict[source_str]))))
-        str_obj.write('}\n')
+        str_obj.write(u'# -*- coding: utf-8 -*-\n')
+        str_obj.write(u'{\n')
+        for source_str, trans_str in sorted(mydict.items()):
+            if six.PY2:
+                source_str = source_str.encode('utf-8')
+                trans_str = trans_str.encode('utf-8')
+            str_obj.write(u"%s: %s,\n" % (repr(source_str), repr(trans_str)))
+        str_obj.write(u'}\n')
         str_obj.seek(0)
         return str_obj
 
@@ -62,16 +66,23 @@ def convertpy(inputfile, outputfile, templatefile=None, includefuzzy=False,
     inputstore = factory.getobject(inputfile)
 
     if not convert.should_output_store(inputstore, outputthreshold):
-        return False
+        return 0
 
     convertor = po2pydict()
     outputstring = convertor.convertstore(inputstore, includefuzzy)
-    outputfile.write(outputstring.read())
+
+    if six.PY2:
+        outputfile.write(outputstring.read().decode('utf-8'))
+    else:
+        outputfile.write(bytes(outputstring.read(), 'utf-8'))
     return 1
 
 
 def main(argv=None):
-    formats = {("po", "py"): ("py", convertpy), ("po"): ("py", convertpy)}
+    formats = {
+        ("po", "py"): ("py", convertpy),
+        ("po", None): ("py", convertpy)
+    }
     parser = convert.ConvertOptionParser(formats, usetemplates=False, description=__doc__)
     parser.add_threshold_option()
     parser.add_fuzzy_option()
--- a/translate/convert/test_web2py2po.py
+++ b/translate/convert/test_web2py2po.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+import sys
+
+from translate.convert import web2py2po
+from translate.storage import po
+from translate.storage.test_base import first_translatable, headerless_len
+
+import six
+import pytest
+
+
+class TestWEB2PY2PO(object):
+
+    def web2py2po(self, web2py_source):
+        """helper that converts po source to web2py source without requiring files"""
+        input_web2py = eval(web2py_source)
+        new_pofile = po.pofile()
+        convertor = web2py2po.web2py2po(new_pofile)
+        output_po = convertor.convertstore(input_web2py)
+        return output_po
+
+    def singleelement(self, storage):
+        """checks that the pofile contains a single non-header element, and returns it"""
+        assert headerless_len(storage.units) == 1
+        return first_translatable(storage)
+
+    def test_basic(self):
+        """test a basic web2py to po conversion"""
+        input_web2py = '''# -*- coding: utf-8 -*-
+{
+'A simple string': 'Du texte simple',
+}
+'''
+
+        po_out = self.web2py2po(input_web2py)
+        pounit = self.singleelement(po_out)
+        assert pounit.source == "A simple string"
+        assert pounit.target == "Du texte simple"
+
+    @pytest.mark.skipif(six.PY2, reason='Skipping unicode tests on PY2')
+    def test_unicode(self):
+        """test a web2py to po conversion with unicode"""
+        input_web2py = '''# -*- coding: utf-8 -*-
+{
+'Foobar': 'Fúbär',
+}
+'''
+
+        po_out = self.web2py2po(input_web2py)
+        pounit = self.singleelement(po_out)
+        assert pounit.source == "Foobar"
+        assert pounit.target == "Fúbär"
--- a/translate/convert/web2py2po.py
+++ b/translate/convert/web2py2po.py
@@ -26,11 +26,14 @@ for examples and usage instructions.
 
 from translate.storage import po
 
+import six
+
 
 class web2py2po(object):
 
-    def __init__(self, pofile=None):
+    def __init__(self, pofile=None, duplicatestyle="msgctxt"):
         self.mypofile = pofile
+        self.duplicatestyle = duplicatestyle
 
     def convertunit(self, source_str, target_str):
         pounit = po.pounit(encoding="UTF-8")
@@ -40,25 +43,26 @@ class web2py2po(object):
         return pounit
 
     def convertstore(self, mydict):
-
         targetheader = self.mypofile.header()
         targetheader.addnote("extracted from web2py", "developer")
 
         for source_str in mydict.keys():
             target_str = mydict[source_str]
+            if six.PY2:
+                target_str = target_str.decode('utf-8')
+                source_str = source_str.decode('utf-8')
             if target_str == source_str:
                 # a convention with new (untranslated) web2py files
-                target_str = u''
-            elif target_str.startswith(u'*** '):
-                # an older convention
                 target_str = u''
             pounit = self.convertunit(source_str, target_str)
             self.mypofile.addunit(pounit)
 
+        self.mypofile.removeduplicates(self.duplicatestyle)
+
         return self.mypofile
 
 
-def convertpy(inputfile, outputfile, encoding="UTF-8"):
+def convertpy(inputfile, outputfile, encoding="UTF-8", duplicatestyle="msgctxt"):
 
     new_pofile = po.pofile()
     convertor = web2py2po(new_pofile)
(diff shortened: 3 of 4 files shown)


petterreinholdtsen:
I suspect it is a good idea to include non-ASCII characters in the test cases. The current ones seem to only include ASCII characters.

petterreinholdtsen:
[Vinyl Darkscratch]
> Good idea.  Added said test!  (This should also be present in the
> other test files, I believe.)

Yes, espesially since the code seem to have problems handling UTF-8
strings in the translations.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I had a look at the failing codecov/patch check, and I believe the reason it report a failure is because there is no unit test for web2py2po.  Perhaps such test can use non-ascii characters?

petterreinholdtsen:
Perhaps the title of this pull request should be changed to reflect that it is not merely updating the code, it is fixing completely broken code?  Something like 'Get web2py support working again' would better reflect the change.

translate/convert/web2py2po.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
vinyldarkscratch: Fixed!  I think I accidentally left this in when I was using the converters on their own, haha.  😛

translate/convert/web2py2po.py
```diff
@@ -58,7 +62,7 @@ def convertstore(self, mydict):
         return self.mypofile
 
 
-def convertpy(inputfile, outputfile, encoding="UTF-8"):
+def convertpy(inputfile, outputfile, encoding="UTF-8", **kwargs):
```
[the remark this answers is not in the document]
vinyldarkscratch: Seems most other converters to pofile have this duplicates handling code implemented, so I'm assuming the omission of `duplicatestyle` was accidental (as far as I saw, only tiki2po didn't have any implementation, which is probably a fluke).  I went ahead and added implementation into this PR.  :wink:

translate/convert/po2web2py.py
```diff
@@ -36,7 +39,7 @@ def __init__(self):
         return
 
     def convertstore(self, inputstore, includefuzzy):
-        str_obj = BytesIO()
+        str_obj = StringIO()
```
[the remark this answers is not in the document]
vinyldarkscratch: Switching to a StringIO was the only way I was able to get both Python 2 and 3 compatibility with support for unicode characters (at least whilst producing consistent, usable results).  I'm open to suggestions if someone has a better idea on how to tackle this, though?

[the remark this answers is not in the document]
petterreinholdtsen: Is the StringIO approach allowing the code to handle non-ASCII translations?  It is not currently tested, so it is hard to be sure.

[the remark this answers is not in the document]
vinyldarkscratch: On Python 2 it didn't seem to matter, but on Python 3, yes, it was my non-ASCII encoding solution.

petterreinholdtsen:
Is there any chance to have a release with this fix included uploaded in Debian before the freeze?  I realize it would require this pull request to be accepted and a new release made very soon, as the Debian freeze is very close.

llimeht:
@petterreinholdtsen I'm cautious about an invasive patch that diverges from upstream. I'd really like it to be reviewed and merged. (To be clear, I also don't want buster to release without fixing the underlying bug.)

vinyldarkscratch:
@llimeht Are you referring to the exchange from `BytesIO` to `StringIO`?

petterreinholdtsen:
[Stuart Prescott]
> @petterreinholdtsen I'm cautious about an invasive patch that diverges
> from upstream. I'd really like it to be reviewed and merged. (To be
> clear, I also don't want buster to release without fixing the
> underlying bug.)

Sure, I would like upstream to review and merge it too.  But so far it
seem unlikely to happen before the Buster freeze, so the question is if
the Debian maintainers are willing to fix this bug even if upstream is
not?  I guess the lack of response from the upstream developer mean they
are busy elsewhere, and I see no sign of this changing any time soon.

-- 
Happy hacking
Petter Reinholdtsen

vinyldarkscratch:
It would be nice to get this PR merged in as quickly as possible, and I'm not saying it because it's my own PR.  Unfortunately, the current version of the converters in the repository fails at runtime, meaning that the `translate-toolkit` is deploying broken code.  A minor bug is more reasonable to deploy versus a major crash.  Really, the only issue I've had is that in Python 2, Unicode characters are encoded in the final output (and that's been the same way with a `BytesIO` object) -- otherwise, I've had no issues, and I have translations from virtually every language.

translate/convert/test_web2py2po.py
```diff
@@ -11,7 +11,8 @@ class TestWEB2PY2PO(object):
     def web2py2po(self, web2py_source):
         """helper that converts po source to web2py source without requiring files"""
         input_web2py = eval(web2py_source)
-        convertor = web2py2po.web2py2po()
+        new_pofile = po.pofile()
+        convertor = web2py2po.web2py2po(new_pofile)
```
unho: Please squash this commit into previous one.

translate/convert/po2web2py.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
unho: Please squash this commit into the first commit in the PR.

translate/convert/po2web2py.py
```diff
@@ -64,7 +64,7 @@ def convertstore(self, inputstore, includefuzzy):
 
 
 def convertpy(inputfile, outputfile, templatefile=None, includefuzzy=False,
-              outputthreshold=None, **kwargs):
+              outputthreshold=None):
```
unho: The remove `**kwargs` is an unrelated change. Please put it into a separate commit.

unho:
Please get rid of the merge commit.

translate/convert/web2py2po.py
```diff
@@ -36,9 +36,9 @@ def __init__(self, pofile=None):
 
     def convertunit(self, source_str, target_str):
         pounit = po.pounit(encoding="UTF-8")
-        pounit.setsource(source_str)
+        pounit.source = source_str
         if target_str:
-            pounit.settarget(target_str)
+            pounit.target = target_str
```
unho: Please squash this commit into first commit in PR since it fixes an usage for a deprecated API that you are introducing there. This way you will be contributing clean code.

translate/convert/po2web2py.py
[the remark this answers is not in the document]
unho: I didn't check yet, but if we don't have tests for those scenarios it would be great if you can add some.

vinyldarkscratch:
@unho Almost all of your suggested changes are to squash the PR into one commit, perhaps you guys should enable [squash merging](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)?

unho:
Or maybe you can create PRs that don't require to be squashed. It is not that difficult :smile: 

The reason why most my comments where about squashing is because I still didn't have time to do a proper review. But please don't let that stop you from squashing commits :smile:

vinyldarkscratch:
Already squashed them into two commits: one for the converters, and one for adding their respective tests.  Last thing is regarding the conversion to `StringIO` it seems, but `StringIO` is built specifically for Unicode strings -- as is the Python 3 default.  Using `BytesIO` limits data to ASCII characters, which is not the best for a package specializing in localization.  In Python 3, I haven't even been able to get a `BytesIO` to work since I've had to encode the strings -- Python 2, it didn't matter either way.

If someone can give examples on what still needs to be tested, I'm happy to implement them.

vinyldarkscratch:
I just noticed that AppVeyor and Travis CI are now failing after the force push, but I have no idea why.  None of my changes affected the `translate/storage/test_aresource.py` file that seems to be segfaulting during the Python 3 tests.

unho:
`master` tests are also failing so it might not be anything related to your PR. Once `master` tests runs are green again we can tell if your PR has any issue that can be detected by test runs.

translate/convert/web2py2po.py
```diff
@@ -25,12 +25,14 @@
 """
 
 from translate.storage import po
+import sys
```
unho:
```python
import sys

from translate.storage import po
```

translate/convert/web2py2po.py
```diff
@@ -40,12 +42,14 @@ def convertunit(self, source_str, target_str):
         return pounit
 
     def convertstore(self, mydict):
-
         targetheader = self.mypofile.header()
         targetheader.addnote("extracted from web2py", "developer")
 
         for source_str in mydict.keys():
             target_str = mydict[source_str]
+            if sys.version_info[0] == 2:
```
unho: I would rather use `if six.PY2:` since that is what is more frequently used in our codebase. It might also help when stripping Python 2 support now that we are less than one year away from Python 2 not being supported anymore.

translate/convert/po2web2py.py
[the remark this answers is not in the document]
unho: If it is not possible to make this work for Python 2 I am open to make this converter a Python 3 only since we are less than one year away from Python 2 being abandoned. In the past we had converters that worked only with Python 2, and this is not one of the most widely used converters, so IMHO it shouldn't such a big issue. Opinions?

translate/convert/test_web2py2po.py
```diff
@@ -0,0 +1,35 @@
...
+        output_po = convertor.convertstore(input_web2py)
+        return output_po
+
+    def singleelement(self, storage):
+        """checks that the pofile contains a single non-header element, and returns it"""
+        assert headerless_len(storage.units) == 1
+        return first_translatable(storage)
+
+    def test_basic(self):
```
unho: I actually don't know anything about this format, but I would be a bit surprised if this is the only test needed to cover all scenarios involved in dealing with this format conversions. Can you please add some additional tests? Thank you very much in advance.

unho: Same for the other test file above.

translate/convert/web2py2po.py
```diff
@@ -24,9 +24,12 @@
 for examples and usage instructions.
 """
 
-from translate.storage import po
 import sys
 
+import six
```
unho: `sys` import above is no longer used.

Also according to our styleguide the imports should be in this order: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#imports `six` counts as a "Third party library import".

translate/convert/test_web2py2po.py
vinyldarkscratch: Sure thing!  Really, only a test regarding Unicode characters is needed -- both this test and the one I just mentioned covers all the scenarios this format will really present.

nijel:
The test failures on master were fixed by https://github.com/translate/translate/pull/3874

vinyldarkscratch:
Lemme fast forward to latest master so we don’t have the big red ❌ for Travis CI!  (I believe that the test may fail for Python 2 due to Unicode stuff, but I’ll be fixing it right away if that’s the case.)

llimeht:
While adding ` @pytest.mark.skipif(six.PY2, reason='Not implemented for PY2')` to the `test_unicode` functions will certainly work, that doesn't really feel like the right answer to the double-encoding problem.

vinyldarkscratch:
> While adding ` @pytest.mark.skipif(six.PY2, reason='Not implemented for PY2')` to the `test_unicode` functions will certainly work, that doesn't really feel like the right answer to the double-encoding problem.

That's true, it doesn't really feel like the right solution, but I don't see any better way of doing it.  The only difference between Python 2's and Python 3's output is that unicode is backslash-encoded, but most programs should be able to handle that fine anyways.  Now, it's just a matter of fighting with the tests to get the unicode checks working on Python 2.  With less than a year on the clock, it's probably better just to start deprecating Python 2 code.

translate/convert/web2py2po.py
```diff
@@ -40,25 +43,26 @@ def convertunit(self, source_str, target_str):
...
 
         for source_str in mydict.keys():
             target_str = mydict[source_str]
+            if six.PY2:
+                target_str = target_str.decode('utf-8')
+                source_str = source_str.decode('utf-8')
             if target_str == source_str:
                 # a convention with new (untranslated) web2py files
                 target_str = u''
-            elif target_str.startswith(u'*** '):
```
unho: What's the reason to remove this?

unho: I see. Nevermind.

unho:
Merged this into master.

vinyldarkscratch:
Woo, it's the moment we've been waiting for!  🎉

unho:
@vinyldarkscratch Forgot to say, thanks for your contribution!

PR #3841, opened 2018-12-02
Add python 3.7 to testing matrix and fix failing tests with pytest 3.10

llimeht:
Python 3.7 is now out and so it's time to add that to the test matrix. Some jiggling is required to express the test matrix in a sensible fashion to get Python 3.7.

There's also a new version of pytest out, and with either pytest 3.9 or 3.10, the way the `deprecated` decorator was tested broke; a second commit to fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,35 +9,18 @@ python:
   - 3.6
 env:
   matrix:
-    - USECPO=0 MINVERSION=1 META=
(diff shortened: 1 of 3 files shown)


nijel:

unho:
@llimeht Thanks for the contribution. @nijel Thanks for reviewing and landing.

PR #3844, opened 2018-12-14
Fix roundtrip of WebExtension JSON with placeholders

nijel:
The placeholders were removed on saving, see https://github.com/WeblateOrg/weblate/issues/2457.


unho:
@nijel lgtm. Please land. Thanks.

PR #3235, opened 2014-12-03
[Blocked by Virtaal] Cleanup: removed dead code

julen:
The `languages` parameter is never used.

The required changes in Pootle/Virtaal have already been made — they were passing the useless argument.

PR #3845, opened 2018-12-14
Remove dependency on diff-match-patch

### LINKED ISSUE #3843: Make toolkit compatible with diff-match-patch 20181111
As of translate-toolkit 2.3.1, in Arch Linux 64-bit, `po2tmx` and other tools fails with diff-match-patch 20181111 ([site]), because `diff-match-patch==20121119` is required.

```
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 581, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 898, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 789, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (diff-match-patch 20181111 (/usr/lib/python3.7/site-packages), Requirement.parse('diff-match-patch==20121119'), {'translate-toolkit'})

During handling of the above exception, another exception occurred:

  File "/usr/bin/po2tmx", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3126, in <module>
    @_call_aside
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3110, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3139, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 596, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 784, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'diff-match-patch==20121119' distribution was not found and is required by translate-toolkit
```

[site]: https://pypi.org/project/diff-match-patch/

nijel:
That is not used by translate-toolkit and it was kept for old Virtaal
releases. The problem is that there is no new Virtaal release and this
dependency now causes problems.

The Virtaal was already fixed, but it's not yet part of any release:

https://github.com/translate/virtaal/commit/77c43a9f9dfda09055794228d3991068abe42d11

Fixes #3843


nijel:
This basically falls into same category as https://github.com/translate/translate/pull/3235 - it will break all existing Virtaal releases because there was no Virtaal release in last six years. I'm not sure it's still good reason to block such changes.

unho:
@nijel Unless anyone @friedelwolff releases a new Virtaal, we cannot drop this from TTK because otherwise we will break all existing Virtaal deployments while not providing any working fallback.

nijel:
I know it would break Virtaal, that's why I've mentioned it's pretty much same situation as #3235. I would not revisit this code without https://github.com/translate/translate/issues/3843. The alternative is to raise the dependency, but that can break Virtaal as well and there is no way to address this without new Virtaal release.

I'm a bit pessimistic about Virtaal release - the last release is six years old, the fix we need here is four years old, last commit on the repository is more than year old. That really doesn't look like there will be new release anytime soon.

unho:
I don't know about if there is going to be a new release for Virtaal soon enough. Maybe @friedelwolff can tell.

But despite that I don't feel the urge to drop this small bit of code (despite I wanted this to be gone for years) since that will piss off translators using Virtaal because we broke it. Plenty of really old Virtaal releases are still out there working perfectly.

To be honest current state of affairs is much more manageable than it previously was, just check e266d7b1 to see how much code we got rid of years ago, so I am happy to keep it as is for a few more years if it is really necessary. But @friedelwolff we really need a new Virtaal release sooner or later.

unho:
I've looked into more detail into Virtaal code re diff-match-patch. Any new release won't depend on TTK for getting diff-match-patch, so the only problem is old Virtaal releases. But it turns out that when Virtaal's dependency on TTK for diff-match-patch was removed (commit https://github.com/translate/virtaal/commit/77c43a9f) diff-match-patch was being imported within a `try..except` block that catched import errors and that provided a fall back to `difflib` from Python's standard library (see https://github.com/translate/virtaal/commit/77c43a9f#diff-5a0cb47e8c4c29dff6305978b3a5c65aL221).

So in summary, unless @friedelwolff raises any problem I think we can safely drop diff-match-patch from TTK.

unho:
@nijel I think we've waited enough for @friedelwolff. Please merge this PR. If any issue arises packagers should have to handle it until a new Virtaal release is out.

nijel:
Thanks, merged.

PR #3846, opened 2018-12-14
Ruby YAML: Add support for plurals

### LINKED ISSUE #3615: Plurals in Ruby YAML Files
As displayed in the good old sample (https://github.com/noosfero/noosfero/blob/master/config/locales/en-US.yml), Ruby (or rails-i18n)'s YAML localization format has some basic plural handling to deal with. This goes down to these `one` `other` things here:

```YML
      less_than_x_minutes:
        one:   "less than a minute"
        other: "less than %{count} minutes"
```

Subkeys used for each language in Ruby's YAML files come directly from [Unicode CIDR Plural Rules](http://www.unicode.org/cldr/charts/28/supplemental/language_plural_rules.html). A basic heuristic for finding plurals is therefore:

1. Look at a key.
2. Does it point to a dictionary? Otherwise return false.
3. Does any of its sub-key members point to a non-string thing? If so, return false.
4. Are all its sub-keys part of the Unicode CIDR cardinal terminology (zero, one, two, few, many, other)? Otherwise return false.
5. Guess true.

Arguably one can never be sure about which key is a plural key, as a perfectly (ab)normal key can also have subkeys named just like plural string groups. Step four can probably improved by some stricter per-language checks.

The current HEAD version does not seem to have such a feature. Since there is no such thing as ruby-yaml-to-po in translate-toolkit, all my observations should be treated as a guess and verified by someone else.

nijel:
Fixes #3615

### DIFF
--- a/translate/storage/yaml.py
+++ b/translate/storage/yaml.py
@@ -146,8 +146,12 @@ class YAMLFile(base.TranslationStore):
         """Returns root node for serialize"""
         return node
 
+    def serialize_value(self, value):
+        return value
+
     def serialize(self, out):
         def nested_set(target, path, value):
+            value = self.serialize_value(value)
             if len(path) > 1:
                 if len(path) == 2 and path[1] and path[1][0] == '[' and path[1][-1] == ']' and path[1][1:-1].isdigit():
                     if path[0] not in target:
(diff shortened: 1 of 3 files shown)


translate/storage/yaml.py
```diff
@@ -171,18 +177,22 @@ def nested_set(target, path, value):
             default_flow_style=False, encoding='utf-8', allow_unicode=True
         ))
 
+    def _parse_dict(self, data, prev):
```
unho: If I am not wrong getting this block of code out to this new `_parse_dict` method doesn't change the behavior, isn't it? If the answer is yes I would prefer this cleanup to happen in a separate commit.

unho: I see now that the intention is to allow to override it on the subclass. Anyway IMHO it is cleaner if this happens in a separate commit.

translate/storage/yaml.py
```diff
@@ -249,3 +259,30 @@ def get_root_node(self, node):
             result[self.targetlanguage] = node
             return result
         return node
+
+    def _parse_dict(self, data, prev):
+        # Does this look like a plural?
+        if all((x in cldr_plural_categories for x in data.keys())):
+            # Ensure we have correct plurals ordering
```
unho: Nit: Comments should end with proper punctuation.

translate/storage/yaml.py
```diff
@@ -249,3 +259,30 @@ def get_root_node(self, node):
...
+            values = [data[item] for item in cldr_plural_categories if item in data]
+            yield (prev, multistring(values))
+            return
+
+        # Handle normal dict
+        for ret in super(RubyYAMLFile, self)._parse_dict(data, prev):
```
unho: Nit: Use a name that is more explanatory than `ret`.

unho:
Adding this feature would require updating the docs on the YAML format.

unho:
@nijel lgtm once comments are addressed. Thanks.

nijel:
I think I've addressed all of the issues.

unho:
@nijel Thanks. lgtm

PR #3847, opened 2018-12-14
Android: Remove not needed code

nijel:
The plurals are already properly sorted in the data module, there is no
need to sort them again.


unho:
@nijel lgtm. Thanks a lot!

PR #3850, opened 2018-12-18
Remove trailing semicolon in Romanian plural definition

### LINKED ISSUE #3837: Romanian plural trailing semicolo
Hi there, is it me or should the definition of Romanian plural not have a trailing semicolon in https://github.com/translate/translate/blob/master/translate/lang/data.py#L160 ? all others don't.

vinyldarkscratch:
Very small patch that doesn't have any effect on the code, mostly for cleanup purposes.  Fixes #3837.


unho:
@vinyldarkscratch Thanks for your contribution. Merged as 00d1fd56.

PR #3851, opened 2018-12-18
Get web2py converter working and add tests to try to keep it like that.

### LINKED ISSUE #3254: Couple issues with Web2py converter

petterreinholdtsen:
Not quite sure the tests are working yet.

Fixes #3254

### DIFF
--- a/tests/cli/data/test_po2web2py/one.po
+++ b/tests/cli/data/test_po2web2py/one.po
@@ -1,0 +1,17 @@
+#. extracted from web2py
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2018-12-18 08:53+0200\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Translate Toolkit 2.0.0\n"
+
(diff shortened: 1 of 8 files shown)


petterreinholdtsen:
I would very much like some help with the tests by someone understanding the test machinery.

vinyldarkscratch:
Hey there!  I've actually got web2py2po and po2web2py fixes in #3838 already (though I'm lacking any tests; I assumed there were already some in translate-toolkit), which includes fixes to get the converters working, as well as many needed adjustments like Unicode support.  :wink:

petterreinholdtsen:
Aha.  Perhaps you can update it with the tests I added to my pull request, and then we can close this pull request.  Would be nice if your pull request would close #3254 too.

petterreinholdtsen:
I replaced the code changes in this pull request with the changes from #3838.  The tests fail, no idea why.

vinyldarkscratch:
Looks like it's failing because a .pyc file is being generated: `web2py2po: WARNING: Couldn't handle input file one.pyc: don't know what to do with input format .pyc`

I actually went ahead and [wrote a test file entirely in Python](https://github.com/translate/translate/pull/3838/files#diff-d0454e2bd0610cfaf8858a5e28d4b396) that copies from the po2json test.  This eliminates the dependency on Travis for the tests, allowing the end user to run `python setup.py test` themselves.  All tests now succeed (aside from a small style check, which I just fixed).

petterreinholdtsen:
Right.  As your patch seem to be better than mine, I close this and hope yours will get merged soon.

PR #3852, opened 2018-12-18
Add macOS and Python 3.7 classifiers

vinyldarkscratch:
Since we're providing support for Python 3.7 and testing against it, we might as well add it as a PyPi classifier.  (Same for macOS, since Windows and Unix are defined.)


unho:
@vinyldarkscratch Thanks for your contribution. Merged in 89f8ec59.

PR #3849, opened 2018-12-18
Update all dependencies and affected functions

vinyldarkscratch:
This PR updates all outdated dependencies and fixes two functions in `translate/lang/data.py` affected by PyCountry's update.

PR #3853, opened 2018-12-25
[requires.io] dependency update on master branch

dwaynebailey:


vinyldarkscratch:
Hey there!  I've actually already got a PR open (#3849) that does all of the dependency updates, and fixes the scripts so that builds don't fail.  🙂

PR #3860, opened 2019-01-21
Remove embedded CherryPy wsgi server

nijel:
It was embedded to avoid depending on CherryPy, but the server is now
standalone package cheroot, which contains just the server.


unho:
@nijel lgtm provided you run some tests with software that uses this, for example amaGama.

nijel:
I've tested the server itself. AFAIK amaGama is not using this code, it's reimplementation of this code.

Looking at it's imports confirms this:

```
amagama/benchmark.py:from translate.storage import factory
amagama/commands.py:from translate.lang.data import langcode_ire
amagama/commands.py:from translate.storage import factory
amagama/tmdb.py:from translate.lang import data
amagama/tmdb.py:from translate.search.lshtein import LevenshteinComparer
amagama/tmdb.py:        from translate.filters.checks import projectcheckers
amagama/tmdb.py:            from translate.lang import factory
amagama/views/api.py:    from translate.storage import base
amagama/views/api.py:    from translate.storage import base
amagama/views/api.py:    from translate.storage import factory
```

Are you aware of something what actually uses this code?

nijel:
Anyway I've added some basic tests for the tmserver. It turns out that it was broken on Python 3 :-).

nijel:
Apparently, I was grepping wrongly :-(. Anyway this API is not changed by this patch. Virtall seems to just execute tmserver program, both these should work without change...

unho:
lgtm.

nijel:
Thanks for review.

PR #3861, opened 2019-01-21
Remove Python 3.3 from test matrix

nijel:
- current lxml no longer supports it making installation fail
- 3.3 has reached end of life on 2017-09-29


unho:
Nice!

Issue #3695, opened 2017-09-14
Advanced Yaml support


Lozik:
In the docs ( http://docs.translatehouse.org/projects/translate-toolkit/en/latest/formats/yaml.html ), I saw that parts of Yaml features are not yet supported. I just wanted to ask if that is planned and if so, if there is any progress being made. Especially the multiline and quoted strings features would be of great importance.

My use case: I use Symfony as web framework and Weblate as translation tool that is based on the Translate Toolkit. As the tool makes it possible to directly push changes, it would be extremely useful to have compatible files build directly by the tool without having to change the files to make them compatible.

Thank you for your valuable work!

unho:
Contributions are always welcome, and we offer paid development if you are willing to pay. But we don't have any specific plan yet on completing YAML support. Weblate developers often contribute, and in fact YAML format support was contributed by them, so I won't be surprised if they contribute the required improvements.

nijel:
I've written most of the YAML support, but I currently do not have any plans to extend it as it serves well in situations I need. Is there some specific issue you're hitting?

Lozik:
Thank you for your answers. Yes, i'm getting the issue that exporting a longer (multiline) string to yaml is not saved as multiline string (using | or >) but as a quoted string which has backslashes before every newline character. 
However, I'm not enough in the topic to tell if this is an issue/missing feature of the Translate Toolkit or if it is only caused by the Weblate project. Sorry about that.

nijel:
The second issue is actually per YAML specification and there are more strings which get parsed as true/false values what can be pretty confusing, see http://yaml.org/type/bool.html

You can try it yourself, enter "yes: no" into http://yaml-online-parser.appspot.com/ and you will get:

```
{
  "true": false
}
```

Lozik:
Hmm, yes, I see.
 I suppose, for the booleans, the error is on the Symfony side, as the parsing component isn't using a strict verification for translations and accepts a `yes: Yes`.

nijel:
Most of incompatiblities should be addressed by https://github.com/translate/translate/pull/3926

PR #3863, opened 2019-01-23
YAML: Switch to ruamel.yaml

nijel:
- This allows us to remove majority of custom code around preserving
  order of keys
- It supports YAML 1.2 which supports yes/no as keys (see #3695)

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -8,5 +8,5 @@ iniparse==0.4 ;python_version < '3.0'      # INI (ini2po)
 vobject==0.9.6.1       # iCal (ical2po)
 aeidon==1.2.1 ;python_version > '3.0'      # Subtitles (sub2po)
 l20n==4.0.0a1        # l20n (l20n2po)
-PyYAML==3.13         # yaml
+ruamel.yaml==0.15.87 # yaml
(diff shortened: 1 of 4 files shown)


nijel:
I'm not sure what to do with the `test_no_duplicates` test - it starts with an invalid YAML document and assumes the tool somehow handles that, what is IMHO wrong...

unho:
Nice PR. This will allow us to circumvent the warning we are having with PyYaml being insecure.

Regarding the test, if it is there it is probably because in the real world we've found some YAML with duplicate keys. The question is, how does ruamel.yaml handle this?

nijel:
It throws an `DuplicateKeyError` error as the file violates the YAML specification, it can be seen here https://travis-ci.org/translate/translate/jobs/483580011#L990

nijel:
Looking at it again, it can be enabled for now, but I'm not sure this is a good idea, also see https://yaml.readthedocs.io/en/latest/api.html#duplicate-keys

unho:
Ok, we can forbid duplicate keys for now. I we should keep the test because maybe in the future it is necessary to enable back the duplicate keys.

lgtm

nijel:
Thanks for review.

PR #3864, opened 2019-01-26
Gettext: Improve finding out newline format for a file

nijel:
Run detection starting at "msgid" string. This should be better way to
detect actual newlines used to detect the strings.

See https://github.com/WeblateOrg/weblate/issues/2519


unho:
@lgtm

nijel:
Thanks for review.

PR #3868, opened 2019-01-27
Gettext: Strip UTF-8 BOM from input

### LINKED ISSUE #1640: PO file with utf-8 BOM signature at beginning of file will not be recognized as translatable
_Version: trunk_

_Originally posted by langtechie:_

(Tested on Firefox 3.6.8)

Uploading a PO file with utf-8 BOM signature at the beginning of the file, and Pootle will say there are no words to translate, even though the file has actual strings to translate.  

Remove the utf-8 BOM and re-upload, and Pootle will recognized words to translate.

Expected: utf-8 files with BOM signature should be supported in Pootle as it is a common practice.

nijel:
Such file is not conforming with how Gettext tools behave, but
apparently some editors create such files.

Fixes #1640

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -59,6 +59,10 @@ def splitlines(text):
     should safely cover weird newlines used in comments or filenames, while
     properly parsing po files with any newlines.
     """
+    # Strip UTF-8 BOM if present. This file would not be accepted
+    # by gettext, but some editors might create it, so better handle it.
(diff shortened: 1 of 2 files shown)


unho:
Doesn't this apply to the `fpo.py` and `cpo.py` implementations as well?

nijel:
No, only to `pypo`.

unho:
lgtm

nijel:
Thanks for review.

Issue #3451, opened 2016-06-21
'dict' object has no attribute 'has_key'


richmind:
Hi,

I installed the translate_toolkit-2.0.0b2 to python3.5.1, 
and got errors as bellow.
(though it's fine, when using python2.7)

Tmserver:
==================
tmserver -b localhost -p 8080 -d TM_JE.db -f TM_JE.tmx -s ja-ja-jp -t en-us

Request:
==================
http://localhost:8080/tmserver/ja-ja-jp/en-us/unit/"TMテストです"

Result:
==================
```pytb
AttributeError("'dict' object has no attribute 'has_key'",)
Traceback (most recent call last):
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/wsgiserver/wsgiserver3.py", line 1032, in communicate
    req.respond()
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/wsgiserver/wsgiserver3.py", line 828, in respond
    self.server.gateway(self).respond()
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/wsgiserver/wsgiserver3.py", line 1839, in respond
    response = self.req.server.wsgi_app(self.env, self.start_response)
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/selector.py", line 146, in __call__
    self.select(environ['PATH_INFO'], environ['REQUEST_METHOD'])
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/selector.py", line 174, in select
    if method_dict.has_key(method):
AttributeError: 'dict' object has no attribute 'has_key'
```

richmind:
Thanks for your job! 
I just met an other error with same request.

# datails
NameError("name 'apply' is not defined",)
Traceback (most recent call last):
  File "/Users/user1/anaconda/envs/virt_conda/lib/python3.5/site-packages/translate_toolkit-2.0.0b2-py3.5.egg/translate/misc/wsgiserver/wsgiserver3.py", line 1032, in communicate
[...]

PR #3869, opened 2019-01-27
Misc: Remove has_key implementation

nijel:
It was deprecated in dict years ago and translate-toolkit doesn't use it
(see #3451), so there is no good reason to implement it.


unho:
lgtm

nijel:
Thanks for review.

PR #3871, opened 2019-02-11
XML indentation for Android, Resx and Xliff

### LINKED ISSUE #3424: Fix indent in XLIFF output

nijel:
This introduces generic XML formatting code based on #3670 and uses that in Android, Resx (what simplifies it's code) and Xliff (what fixes #3424).

### DIFF
--- a/translate/misc/test_xml_helpers.py
+++ b/translate/misc/test_xml_helpers.py
@@ -1,0 +1,35 @@
+from lxml import etree
+
+from translate.misc.xml_helpers import reindent
+
+
+class TestReindent:
+    def _xmlfromstring(self, xmlstring):
+        return etree.fromstring(xmlstring)
+
+    def _xmltostring(self, xml):
+        return etree.tostring(xml, pretty_print=True, xml_declaration=True, encoding='utf-8')
+
+    def test_indent_four_spaces(self):
+        """Test that using 4 spaces for indent yields a consistent result."""
+        xmlsource = self._xmlfromstring('<root><str key="test">Test</str></root>')
+        reindent(xmlsource, indent="    ")
+        actual = self._xmltostring(xmlsource)
+        expected = b'''<?xml version='1.0' encoding='utf-8'?>
+<root>
+    <str key="test">Test</str>
+</root>
+'''
+        assert actual == expected
+
+    def test_indent_tab(self):
+        """Test that using a tab for indent yields a consistent result."""
+        xmlsource = self._xmlfromstring('<root><str key="test">Test</str></root>')
+        reindent(xmlsource, indent="\t")
+        actual = self._xmltostring(xmlsource)
+        expected = b'''<?xml version='1.0' encoding='utf-8'?>
+<root>
+\t<str key="test">Test</str>
+</root>
+'''
+        assert actual == expected
--- a/translate/misc/xml_helpers.py
+++ b/translate/misc/xml_helpers.py
@@ -133,3 +133,33 @@ def normalize_xml_space(node, xml_space, remove_start=False):
 
     for child in node:
         normalize_xml_space(child, remove_start)
+
+
+def reindent(elem, level=0, indent="  ", max_level=4, skip=None):
+    """Adjust indentation to match specification.
+
+    Each nested tag is identified by indent string, up to
+    max_level depth, possibly skipping tags listed in skip.
+    """
+    i = "\n" + (indent * level)
+    if skip and elem.tag in skip:
+        next_level = level
+        extra_i = i
+    else:
+        next_level = level + 1
+        extra_i = i + indent
+    if len(elem) and level < max_level:
+        if (not elem.text or not elem.text.strip()) and getXMLspace(elem) != 'preserve':
+            elem.text = extra_i
+        if not elem.tail or not elem.tail.strip():
+            elem.tail = i
+        for child in elem:
+            reindent(child, next_level, indent, max_level)
+        if not child.tail or not child.tail.strip():
+            child.tail = i
+    if level:
+        if not elem.tail or not elem.tail.strip():
+            elem.tail = i
+    else:
+        if not elem.tail or not elem.tail.strip():
+            elem.tail = ''
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -29,6 +29,7 @@ from lxml import etree
 
 from translate.lang import data
 from translate.misc.deprecation import deprecated
+from translate.misc.xml_helpers import reindent
 from translate.storage import base, lisa
 from translate.misc.multistring import multistring
 
@@ -57,7 +58,6 @@ class AndroidResourceUnit(base.TranslationUnit):
                 self.xmlelement = etree.Element("plurals")
             else:
                 self.xmlelement = etree.Element("string")
-            self.xmlelement.tail = '\n'
         if source is not None:
             self.setid(source)
         super(AndroidResourceUnit, self).__init__(source)
@@ -337,7 +337,6 @@ class AndroidResourceUnit(base.TranslationUnit):
             if self.xmlelement.tag != "plurals":
                 old_id = self.getid()
                 self.xmlelement = etree.Element("plurals")
-                self.xmlelement.tail = '\n'
                 self.setid(old_id)
 
             locale = self.gettargetlanguage().replace('_', '-').split('-')[0]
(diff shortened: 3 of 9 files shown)


translate/storage/test_xliff.py
```diff
@@ -479,3 +479,34 @@ def test_multiple_filenodes(self):
         assert newxfile.getfilenode("file0") is not None
         assert newxfile.getfilenode("file1") is not None
         assert not newxfile.getfilenode("foo")
+
+    def test_indent(self):
+        xlfsource = b'''<?xml version='1.0' encoding='UTF-8'?>
```
eighthave: `"` double quotes are by far the most common quotes used for this in XML.

eighthave:
The quoting stuff in XML is a nightmare, there is no consistency!  The XLIFF spec shows `'` single quotes most of the time, but seems to show `"` double quotes for the `<xml>` lines. http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#Specs_XMLDecl  They also seem to mostly use `'` throughout the example XML, but sometimes with `"` mixed in.

At least one test case should have `"` double quotes, that's what the Android tools use in all the XML, and seems to be the de facto standard for XML https://github.com/lxml/lxml/pull/277

Reading this code makes me think that it would be better to use an existing pretty-printer like LXML, but then just force the `<xml>` declaration lines to use `"` double quotes rather than `'` single quotes.  Or maybe another Python XML pretty printer uses `"`.

nijel:
This PR is about indentation, not quoting, so let's keep that separate please :-).

As for indentation, the pretty printer in lxml is conservative about changing existing whitespace (see https://lxml.de/FAQ.html#why-doesn-t-the-pretty-print-option-reformat-my-xml-output), so it's not really a good fit in this case on it's own. The approach I've chosen seems to be most reliable see https://stackoverflow.com/q/749796/225718 for related discussions.

BhaaLseN:
I just replaced my own `reindent` method in my flat xml with your `cleanindent`, and it fails a few of my tests - an unindented xml run through it will not fix the indent of the first element. Not sure if that is intended or not. Elements following the first (ie. second and onwards) are fine.

Might be worthwhile adding a few tests for `cleanindent` itself with a constructed `lisa` instance
```py
class TestCleanIndent:
    def _xmlfromstring(self, xmlstring):
        """TODO: return etree document constructed from passed xmlstring"""
        pass

    def _xmltostring(self, xml):
        """TODO: return serialized xml string from passed etree xml"""
        pass

    def _cleanindent(self, elem, indent="  ", maxlevel=4):
        """TODO: call lisa.cleanindent with the passed parameters"""
        pass

    def test_indent_four_spaces(self):
        """Test that using 4 spaces for indent yields a consistent result."""
        xmlsource = self._xmlfromstring('<root><str key="test">Test</str></root>')
        xmltarget = self._cleanindent(xmlsource.getroot(), indent="    ")
        actual = self._xmltostring(xmltarget)
        expected = (
            "<?xml version='1.0' encoding='UTF-8'?>\n"
            '<root>\n'
            '    <str key="test">Test</str>\n'
            '</root>\n'
        )
        assert actual == expected

    def test_indent_tab(self):
        """Test that using a tab for indent yields a consistent result."""
        xmlsource = self._xmlfromstring('<root><str key="test">Test</str></root>')
        xmltarget = self._cleanindent(xmlsource.getroot(), indent="\t")
        actual = self._xmltostring(xmltarget)
        expected = (
            "<?xml version='1.0' encoding='UTF-8'?>\n"
            '<root>\n'
            '\t<str key="test">Test</str>\n'
            '</root>\n'
        )
        assert actual == expected
```
(hope I didn't mess up the indents or other stuff shortened/changed when typing this into the GitHub comment box; but I suppose you'll see what I mean)

nijel:
@BhaaLseN good idea to add separate tests, I've just added those. Also the case you've mentioned is now properly handled.

translate/storage/test_lisa.py
```diff
@@ -0,0 +1,38 @@
+from lxml import etree
+
+from translate.storage.lisa import LISAfile
+
+class TestCleanIndent:
+    def _xmlfromstring(self, xmlstring):
+        """TODO: return etree document constructed from passed xmlstring"""
```
BhaaLseN: You can probably remove that `TODO` over here, was mostly left out from my comment because my methods had all that storage handling from the usual storage tests and I wanted to put the focus on the actual test methods themselves.

translate/storage/test_lisa.py
```diff
@@ -0,0 +1,38 @@
...
+        return etree.fromstring(xmlstring)
+
+    def _xmltostring(self, xml):
+        """TODO: return serialized xml string from passed etree xml"""
```
BhaaLseN: Same down here regarding `TODO`

nijel: Already did that in next push ;-)

translate/misc/test_xml_helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_indent_four_spaces(self):
+        """Test that using 4 spaces for indent yields a consistent result."""
+        xmlsource = self._xmlfromstring('<root><str key="test">Test</str></root>')
+        reindent(xmlsource, indent="    ")
+        actual = self._xmltostring(xmlsource)
+        expected = (
+            "<?xml version='1.0' encoding='utf-8'?>\n"
+            '<root>\n'
+            '    <str key="test">Test</str>\n'
+            '</root>\n'
```
unho: Here you have enclosing double quotes in first line of implicit string continuation, but single quotes in the following three lines.

unho: Same below.

unho: I now see that below we are testing tab indenting, and using triple quoted string might hide the presence of a tab in the expected string.

Anyway I find weird to have different quoting along the string. As long as it works ok, I am fine to let this one go.

translate/misc/xml_helpers.py
```diff
@@ -133,3 +133,23 @@ def normalize_xml_space(node, xml_space, remove_start=False):
 
     for child in node:
         normalize_xml_space(child, remove_start)
+
+
+def reindent(elem, level=0, indent="  ", maxlevel=4):
+    """adjust indentation to match specification"""
```
unho: Nit: Start docstring with Titlecase and add a trailing period.

translate/misc/xml_helpers.py
```diff
@@ -133,3 +133,23 @@ def normalize_xml_space(node, xml_space, remove_start=False):
 
     for child in node:
         normalize_xml_space(child, remove_start)
+
+
+def reindent(elem, level=0, indent="  ", maxlevel=4):
```
unho: Use `max_level` instead.

translate/misc/xml_helpers.py
unho: I would also add some additional description to tell about how the arguments affect the behavior.

translate/storage/aresource.py
```diff
@@ -29,6 +29,7 @@
 
 from translate.lang import data
```
unho: Typo in commit message.

translate/storage/resx.py
```diff
@@ -24,7 +24,7 @@
 
 from translate.lang import data
```
unho: Typo in commit message.

unho:
My main concern is tests not passing. I see that tests are not passing on `master` either. I have also left some minor comments.

nijel:
I've adjusted the changes based on your feedback.

There seems to be some issue in python or some of it's modules on Travis, as occasionally the tests segfault. This happens on master as well and I have no clue how to approach this.

nijel:
Updated the patch set to unify ts2 to new reindent code as well.

unho:
@nijel lgtm

nijel:
Thanks for review.

PR #3875, opened 2019-02-14
CSV: Make default dialect more flexible

nijel:
Changed default quoting to csv.QUOTE_ALL from csv.QUOTE_NONNUMERIC. The
problem with csv.QUOTE_NONNUMERIC is that it tries to convert all non
quoted stringst to float leading to exceptions when parsing the files
without quoted strings:

ValueError("could not convert string to float: 'ID'")

The root cause is that sniffer is not able to detect a delimiter, but
this seems to be fixed in Python 3.7.

### DIFF
--- a/translate/storage/test_csvl10n.py
+++ b/translate/storage/test_csvl10n.py
@@ -44,6 +44,19 @@ class TestCSV(test_base.TestTranslationStore):
         assert store.units[0].target == 'zkouška sirén'
         assert bytes(store) == content
 
+    def test_default(self):
+        content = '''ID,English
+GENERAL@2|Notes,"cable, motor, switch"
+*****END CALL*****|Ask,-'''
+        store = self.parse_store(content)
+        assert len(store.units) == 3
+        assert store.units[0].location == 'ID'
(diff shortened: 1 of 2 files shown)


translate/storage/test_csvl10n.py
```diff
@@ -44,6 +44,13 @@ def test_utf_8_sig(self):
         assert store.units[0].target == 'zkouška sirén'
         assert bytes(store) == content
 
+    def test_default(self):
+        content = '''ID,English
+GENERAL@2|Notes,"cable, motor, switch"
+*****END CALL*****|Ask,-'''
+        store = self.parse_store(content)
+        assert len(store.units) == 3
```
unho: Please extend this test to check the `source`, the `target` and the `context` for each of the parsed units.

If we don't have roundtrip test for this it would be great to add some as well.

nijel: I've added tests, roundtrip will change the file because of mixed quoting is not really supported on saving.

unho:
lgtm provided builds pass

nijel:
Thanks for review.

PR #3878, opened 2019-02-15
Misc: Fix indentation of toplevel tags

nijel:
The toplevel tags were wrongly indented in some cases.


unho:
lgtm

nijel:
Thanks for review.

PR #3880, opened 2019-02-22
CSV: Avoid hardcoding escape char

### LINKED ISSUE #3246: csv2po new line
Hi there, i have CSV with:

```csv
"yellow pencil","żółty\nołówek"
"photo description","opis zdjęcia"
```

And converted with csv2po:

```po
msgid "yellow pencil"
msgstr "żółtynołówek"

msgid "photo description"
msgstr "opis zdjęcia"
```

How to convert with save format "\n", script delete \ from string.

nijel:
Really rely on sniffer here, otherwise we might strip too much (the
escape char in CSV is meant only for quotes, otherwise the parser just
strips the escape char).

Fixes #3246

### DIFF
--- a/translate/convert/test_csv2po.py
+++ b/translate/convert/test_csv2po.py
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
 from translate.convert import csv2po, test_convert
 from translate.misc import wStringIO
 from translate.storage import csvl10n, po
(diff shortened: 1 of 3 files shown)


nijel:
I'm not really sure by original intentions of this code, but I currently don't see where it could bring something good. It's there since introduction of sniffer in cb74f6805ac6cd7e6046938b3d5b2088519b2680.

unho:
lgtm

nijel:
Thanks for review.

PR #3881, opened 2019-02-27
Subtitles: Initialize duration in __init__

nijel:
It is set in SubtitleFile._parse, but the attribute should be already
present.


unho:
lgtm

nijel:
Thanks for review.

PR #3882, opened 2019-02-27
Storage: consistent API for removenotes

nijel:
Similarly like with addnote and getnotes, this now accepts origin on all
storages, even when most of them do ignore it.


unho:
lgtm

nijel:
Thanks for review.

PR #3886, opened 2019-03-27
Change empty file handling

nijel:
I'm not sure with this change, but to me it looks weird that *all* formats accept empty files through `parsefile` or `parsestring` even when those are not syntactically valid (eg. XML or JSON based formats).

The code in question was introduced in https://github.com/translate/translate/commit/9141370e95ac5ced702aa1b503d1cfbb96051c6a where it was ported from some formats.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -774,11 +774,10 @@ class TranslationStore(object):
     def parsestring(cls, storestring):
         """Convert the string representation back to an object."""
         newstore = cls()
-        if storestring:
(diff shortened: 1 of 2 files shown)


nijel:
When digging deeper I *think* that reason for this was to avoid failure on calling `parsefile` on files opened for writing. I've addressed this directly in `parse` instead of guessing based on empty value in `parsefile`.

unho:
@nijel lgtm

nijel:
Thanks for review.

PR #3894, opened 2019-05-29
json files with BOM char, fix and unittest

zodman:


nschonni:
BOM isn't allowed in JSON, so this shouldn't need to be handled https://stackoverflow.com/a/38036753/455535

zodman:
OMG! ok ok ok!

PR #3897, opened 2019-06-07
PO: Allow unicode in po headers

### LINKED ISSUE #3896: Traceback with non-ascii header key
If a PO header contains a non-ASCII key (left of the colon), we have a traceback. An example is file here:
https://gitlab.gnome.org/GNOME/gimp/blob/master/po-python/oc.po#L21

The traceback comes from poheader.gettargetlanguage():

```python-traceback
  File "env/lib/python2.7/site-packages/translate/storage/poheader.py", line 286, in gettargetlanguage
    header = self.parseheader()
  File "env/lib/python2.7/site-packages/translate/storage/poheader.py", line 219, in parseheader
    return parseheaderstring(header.target)
  File "env/lib/python2.7/site-packages/translate/storage/poheader.py", line 52, in parseheaderstring
    key = str(key.strip())
UnicodeEncodeError: 'ascii' codec can't encode character u'\xed' in position 4: ordinal not in range(128)
```

I don't have a usable toolkit environment configured, so anybody is free to fix this if they can/want. I guess translate/storage/poheader.py::parseheaderstring() line 52 the `str(key.strip())` could probably just be without the str(), but we'll have to check the tests. Also, I think in some places header dictionaries might be passed as **kwargs, and non-ascii values might not work in Python 2, so I'm not sure it is a trivial change.

nijel:
Fixes #3896

### DIFF
--- a/translate/storage/test_pypo.py
+++ b/translate/storage/test_pypo.py
@@ -269,6 +269,33 @@ class TestPYPOFile(test_po.TestPOFile):
         print(bytes(oldfile))
         assert len(oldfile.units) == 1
 
+    def test_unicode_header(self):
+        """checks that unicode header is parsed and saved correctly"""
+        posource = r'''msgid ""
+msgstr ""
+"PO-Revision-Date: 2006-02-09 23:33+0200\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8-bit\n"
+"Zkouška: něco\n"
+'''.encode('utf-8')
+        pofile = self.poparse(posource)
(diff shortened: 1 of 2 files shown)


translate/storage/test_pypo.py
```diff
@@ -269,6 +269,33 @@ def test_unassociated_comments(self):
         print(bytes(oldfile))
         assert len(oldfile.units) == 1
 
+    def test_unicode_header(self):
+        """checks that prevmsgid (i.e. #|) is parsed and saved correctly"""
```
friedelwolff: I guess this is a copy+paste mistake...

nijel: good point, fixed

Issue #3901, opened 2019-06-18
Qt Linguist translations since Qt 5 now uses type="vanished" instead of type="obsolete"


Wohlstand:
Qt Linguist's TS parser lacks the support for `type="vanished"` which is the same as `type="obsolete"` since Qt 5. Both are meaning the same.

**Little about Qt 5 and Qt 4 TS differences**: they are similar to each other, however, there are some differences between each other:
* `<message>` tag no more uses `utf8="true"` tag, the codepage is always UTF8 or specified in the header of entire TS file.
* Version of TS is 2.1 on Qt 5 while 2.0 on Qt 4
* Vanished messages are now marking with `type="vanished"` in Qt 5 and `type="obsolete"` in Qt 4. However, Qt 4's "obsolete" is still work on Qt Linguist of Qt 5, but no more coming by default.

An example TS file which contains "vanished" translations: https://github.com/Wohlstand/OPL3BankEditor/blob/master/src/translations/opl3bankeditor_fr_FR.ts

Wohlstand:
I have made a small change that adds the support for "vanished" type to work the same as "obsolete".

Wohlstand:
As this has been implemented and merged, so, this issue has been solved!

ossilator:
the analysis behind this report is incorrect: 'vanished' didn't replace 'obsolete'; rather, one complements 'unfinished' and the other 'finished'/default (i don't remember which way around). i implemented this specifically to mirror .po's 'obsolete fuzzies'.

Adama-WADE:
From what I observed after running a few tests, "obsolete" is assigned when a message with a translation that hadn't been validated yet is removed, whereas "vanished" is assigned when the translation of the deleted message had been validated. Both show up in Qt 5.

nijel:
I still don't see an issue with the implementation in https://github.com/translate/translate/pull/3902. If you think something should be changed, please open a separate issue describing how it should behave.

ossilator:
i didn't check the current state of affairs, but the original PR treats vanished and obsolete identically, which is just wrong and doesn't map to the full capabilities of the po format.

ossilator:
prospectively, you might want to use this information when seeding translation memory with these strings.

anyway, as long as you're not falsifying the state info when writing back modified files, everything is fine.

PR #3902, opened 2019-06-18
TS: Add support for type="vanished" as type="obsolete" for Qt 5

Wohlstand:
Issue #3901

### DIFF
--- a/translate/storage/ts2.py
+++ b/translate/storage/ts2.py
@@ -82,6 +82,7 @@ class tsunit(lisa.LISAunit):
 
     statemap = {
         "obsolete": S_OBSOLETE,
+        "vanished": S_OBSOLETE,
         "unfinished": S_FUZZY,
         "": S_TRANSLATED,
         None: S_TRANSLATED,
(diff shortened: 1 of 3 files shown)


translate/storage/ts2.py
```diff
@@ -349,7 +350,7 @@ def merge(self, otherunit, overwrite=False, comments=True, authoritative=False):
             self.markfuzzy(False)
 
     def isobsolete(self):
-        return self._gettype() == "obsolete"
+        return self._gettype() == "obsolete" or self._gettype() == "vanished"
```
nijel: Can you please add test covering this? Apparently this method is not tested at all.

Wohlstand: Do you mean adding the test case into existing unit test to verify this feature? Ok, I'll add that once I'll reach my PC.

nijel:

Wohlstand:
Thanks! :fox_face: 
The last question for curiosity: is a new release planned in some time moment, or will happen once some internal milestones will be completed?

nijel:
This is more question to @unho  and I'd like to know answer as well ;-)

unho:
@Wohlstand We can certainly look into a new release soon since last one was in october and @nijel has been actually quite a good progress.

unho:
@Wohlstand @nijel Translate Toolkit 2.4.0 is out!

PR #3906, opened 2019-06-27
CSV: Share StringIO for parsing and detection

nijel:
There is no good reason constructing the StringIO twice.

### DIFF
--- a/translate/storage/csvl10n.py
+++ b/translate/storage/csvl10n.py
@@ -248,9 +248,8 @@ def valid_fieldnames(fieldnames):
     return False
 
 
-def detect_header(sample, dialect, fieldnames):
+def detect_header(inputfile, dialect, fieldnames):
     """Test if file has a header or not, also returns number of columns in first row"""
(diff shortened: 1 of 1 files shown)


friedelwolff:
The CSV sniffer exercises some regular expressions over the supplied data. The current code limits the amount of text the sniffer processes, and I guess the sniffer assumes it won't receive a lot of text. With your change, the whole file will be given to the sniffer.

Maybe just check that you are still happy with the performance on big files. I don't expect major problems, and don't consider this terribly performance sensitive.

nijel:
This patch does not affect CSV sniffer, just the header detection. That uses normal CSV reader which consumes input by lines, so the patch makes no difference here (only in case where first line does not fit into 1024 bytes, where it would actually make the reader behave better because it would receive complete line).

The CSV sniffing is problematic as well (see https://github.com/WeblateOrg/weblate/issues/2805), but I didn't yet have time to look deeper into that.

unho:
lgtm

nijel:
Thanks for review.

PR #3908, opened 2019-07-05
FlatXML: Rename element attribute

nijel:
This provides API compatibility with LISA based storages.


unho:
lgtm

nijel:
Thanks for review.

PR #3912, opened 2019-08-02
pypo: raise erorr on invalid file content

nijel:
The pypo parser happily parsed any file as containing no unit - it just
stopped on first line and did not report any error. Now the ValueError
is raised in case it can not parse the whole file.

### DIFF
--- a/translate/storage/poparser.py
+++ b/translate/storage/poparser.py
@@ -53,6 +53,7 @@ class ParseState(object):
         # and for decoding all further strings.
         self._input_iterator = input_iterator
         self.next_line = ''
+        self.lineno = 0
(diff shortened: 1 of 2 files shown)


stanislav-brabec:
I am not sure what caused the crash, but a new commit fixed it. Differences between these commits:
- Fix missing POT-Creation-Date
- The original source contains lines like msgstr "Paket^@Pakete", where ^@ is a NUL character.
- Context comments starting with "#:"
- "Language-Team: header is in a single line instead of two physical lines.

One of these changes fixed the problem. I suspected missing POT-Creation-Date, but NUL msgstr could be the problem source as well.

PR #3915, opened 2019-08-12
Android: Fix XML enttites handling

### LINKED ISSUE #3907: Android xml parsing fails if it contains an Entity
I'm using WeblateOrg and I encountered an issue I reported on their project: https://github.com/WeblateOrg/weblate/issues/2843

From what they said, the problem comes from your library.
I try to use an Entity in a Doctype (see below image)
Screenshot_20190628_161758
and the error I get is 
`app/src/main/res/values/strings.xml | attribute 'text' of 'lxml.etree._Entity' objects is not writable`

You can try on the `app/src/main/res/values/strings.xml` file in the `feature/trying_to_use_entity_in_resources` branch of https://github.com/BelledonneCommunications/linphone-android repository 

Thanks in advance.

nijel:
Properly parse, indent and store XML entities.

Fixes #3907

### DIFF
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -273,7 +273,7 @@ class AndroidResourceUnit(base.TranslationUnit):
             if cloned_target.text is not None:
                 cloned_target.text = self.unescape(cloned_target.text, False)
             for xmlelement in cloned_target.iterdescendants():
-                if xmlelement.text is not None:
+                if xmlelement.text is not None and xmlelement.tag is not etree.Entity:
                     xmlelement.text = self.unescape(xmlelement.text, False)
                 if xmlelement.tail is not None:
                     xmlelement.tail = self.unescape(xmlelement.tail, False)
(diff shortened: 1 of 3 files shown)


translate/storage/aresource.py
```diff
@@ -292,11 +292,23 @@ def get_xml_text_value(self, xmltarget):
...
+                for child in cloned_root.xpath('//string'):
+                    cloned_root.remove(child)
+                template = etree.tostring(cloned_doc, encoding='unicode')
+            else:
+                template = '<resources></resources>'
             try:
-                # Try to handle it as legacy XML
-                parser = etree.XMLParser(strip_cdata=False, resolve_entities=False)
-                newstring = etree.fromstring('<string>%s</string>' % target, parser)
+                print(template.replace('</resources>', '<string>%s</string></resources>' % target))
```
unho: Remove this `print`.

nijel: Good catch, forgotten debugging.

PR #3916, opened 2019-08-12
pypo: Fix serializing long msgidcomments

nijel:
This repeatedly added blank lines to the output, see https://github.com/WeblateOrg/weblate/issues/2954


SlavekB:
I backported it to my Weblate instance, so we'll see what happens :smiley_cat:

PR #3917, opened 2019-08-14
Stop removing blank lines in .ts files

smokris:
Fixes issue where po2ts would remove consecutive linebreaks in source and translated strings.  Added tests.

(I'm not sure why the blank-line removal code was there in the first place; I traced it back as far as 9836716 (12 years ago) and it isn't documented.  Removing it doesn't break any tests.)

### DIFF
--- a/translate/storage/test_ts2.py
+++ b/translate/storage/test_ts2.py
@@ -60,6 +60,12 @@ The other line&apos;s translation may have quotes, too (&quot;&apos;&quot;,
 &apos;&quot;&apos;)</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Line 1
+
+Line 3</source>
+        <translation type="unfinished"></translation>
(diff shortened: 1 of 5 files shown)


translate/storage/test_ts2.py
```diff
@@ -60,6 +60,12 @@
 &apos;&quot;&apos;)</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Line 1
+
+Line 3</source>
+        <translation type="unfinished"></translation>
+    </message>
```
nijel: You're adding test for ts2 storage, but patching ts storage, that looks like a mismatch.

smokris:
> use ts2 instead of ts in the converters

Ah, good point.  In the latest version of this PR, I reworked `po2ts` to use `ts2` (and removed the changes to `ts`).

nijel:

PR #3930, opened 2020-01-09
po <-> web2py: don't copy "@markmin" string

vinyldarkscratch:
web2py includes `@markmin\x01` in the unit's source string when the string uses web2py's markmin formatting.  However, it does not copy over to the translation.  This PR updates the converter to reflect this same behavior, allowing a smoother experience.

### DIFF
--- a/translate/convert/test_po2web2py.py
+++ b/translate/convert/test_po2web2py.py
@@ -71,3 +71,17 @@ msgstr "zab"
 '''
         web2py_out = self.po2web2py(input_po)
         assert web2py_out == expected_web2py
+
+    def test_markmin(self):
+        """test removal of @markmin in po to web2py conversion"""
+        input_po = '''
+msgid "@markmin\x01Hello **world**!"
+msgstr ""
+'''
+        expected_web2py = '''# -*- coding: utf-8 -*-
+{
+'@markmin\\x01Hello **world**!': 'Hello **world**!',
+}
+'''
+        web2py_out = self.po2web2py(input_po)
+        assert web2py_out == expected_web2py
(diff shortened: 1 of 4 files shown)


nijel:
Thanks for the pull requests, can you please also add some tests for this behavior?

vinyldarkscratch:
Sure thing @nijel -- done and pushed!

translate/convert/test_po2web2py.py
```diff
@@ -68,6 +68,20 @@ def test_ordering_serialize(self):
 'baz': 'zab',
 'foo': 'oof',
 }
+'''
+        web2py_out = self.po2web2py(input_po)
+        assert web2py_out == expected_web2py
+
+            def test_markmin(self):
```
nijel: Messed up indentation?

```suggestion
    def test_markmin(self):
```

nijel:
The new test does fail, can you please check that?

https://travis-ci.org/translate/translate/jobs/635406499#L725

vinyldarkscratch:
Apologies for the failing tests -- I was thinking this was going to be a quickie, but now that I was able to sit down I realized there was a little bit more to it too.  Both directions need to be accounted for, not just one of them.

This PR is now fully ready, I'd say!

nijel:

nijel:
PS: I'm afraid we're not consistent in merge - I do that or rebase on GitHub, but other maintainers merge on command line and accept only fast forward pull requests.

vinyldarkscratch:
Ah, I see, so it depends on who’s reviewing.  I’ll keep that in mind, thank you, and thanks for the merge!  👍

PR #3931, opened 2020-01-09
Properties: Avoid creating comment only units

### LINKED ISSUE #3928: Difference in comment extraction
There is difference in extracting comments from Android and Java properties in regard to nelines between the string and comment.

Java properties:

```
string1=String 1
# This comment will NOT be shown

string2=String 2

#This comment will be shown
string3= String 3
```

Android:

```
    <string name="string1">String 1</string>
    <!-- This comment IS shown -->

    <string name="string2">String 2</string>
    <!-- This comment should be shown -->
    <string name="string3">String 3</string>
```

Probably it should behave consistently.

Originally brought up by @tariver in https://github.com/WeblateOrg/weblate/issues/3295

nijel:
The units with comment only are problematic as they can not be
identified - the getid returns just empty string for all of them. This
makes it hard to process them. Concatenating such comments makes it also
behave consistent with other storages.

Fixes #3928

### DIFF
--- a/translate/convert/test_prop2po.py
+++ b/translate/convert/test_prop2po.py
@@ -205,7 +205,7 @@ do=translate me
         unit = self.singleelement(pofile)
         assert unit.source == "\nvalue\n"
 
-    def test_unassociated_comments(self):
+    def test_header_comments(self):
         """check that we can handle comments not directly associated with a property"""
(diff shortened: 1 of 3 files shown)


unho:
@nijel lgtm. I guess you have checked that roundtrip works ok.

nijel:
Roundtrip is well covered by tests here.

PR #3936, opened 2020-01-17
pypo: Avoid stripping empty lines from comments

nijel:
The empty line inside comments are usually intentional, so avoid dropping them
on round trip.

Issue was brought up by @fujiwarat at https://github.com/WeblateOrg/weblate/issues/3376

PR #3947, opened 2020-02-21
pypo: Avoid adding extra space on empty comment lines

nijel:
Should address https://github.com/translate/translate/pull/3936#issuecomment-589541155


unho:
@nijel lgtm

nijel:
Thanks for review.

PR #3948, opened 2020-02-21
pypo: Performance and code cleanup

nijel:
* Avoid building strings in place
* Use any + chain instead of iterating over temporarily created array
* Remove not needed else statements
* Avoid using len() as boolean condition
* Directly return bool instead of using if


nijel:
Thanks for review.

PR #3952, opened 2020-03-18
ts2: Avoid duplicating plurals definition

nijel:
I see no good reason to duplicate the data.

### DIFF
--- a/translate/storage/test_ts2.py
+++ b/translate/storage/test_ts2.py
@@ -176,6 +176,16 @@ class TestTSfile(test_base.TestTranslationStore):
         assert checkunit.target == [u"Leêr", u"Leêrs"]
         assert checkunit.hasplural()
 
+    def test_nplural(self):
+        tsstr = '''<!DOCTYPE TS>
+<TS version="2.0" language="xx" sourcelanguage="de">
+</TS>
+'''
+        tsfile = ts.tsfile.parsestring(tsstr)
+        assert tsfile.nplural() == 1
+        tsfile = ts.tsfile.parsestring(TS_CONTEXT_QT4)
+        assert tsfile.nplural() == 3
+
     def test_language(self):
         """Check that we can get and set language and sourcelanguage
         in the header"""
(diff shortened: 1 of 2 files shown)


nijel:
This code was introduced in 079b5fc99 and seems only used in virtaal: https://github.com/translate/virtaal/blob/8e883ae96172b2ab1f3bbfded7d19da25185b5a8/virtaal/models/storemodel.py#L225-L237

unho:
@nijel If this is still being used in Virtaal I am afraid we cannot yet land it to prevent breaking existing Virtaal installs with no newer working Virtaal release available. @friedelwolff Any possibility of having a new Virtaal release soon?

nijel:
This doesn't break existing API, it just makes it use shared data we have for plurals. It was just a remark about usage of this function. I've also added test to ensure it behaves correctly.

unho:
Sorry, you are right.

Anyway I did some check and some of the entries in the removed `NPLURALS` dict do not match those from `translate.data.languages` dict:

>     'cy': 5,
>     'ga': 3,
>     'jp': 1,
>     'mk': 3,

nijel:
Good point, going through them now:

* `cy` - Qt actually has 4 plurals, see https://github.com/qt/qttools/blob/fb96051e4200550f5deafae9f2c775d894eb9c7f/src/linguist/shared/numerus.cpp#L84-L88
* `ga` - Qt actually has 2 plurals, see https://github.com/qt/qttools/blob/fb96051e4200550f5deafae9f2c775d894eb9c7f/src/linguist/shared/numerus.cpp#L51-L53
* `jp` - legacy code for Japanese, number of plurals does match otherwise
* `mk` - Qt actually has 2 plurals, see https://github.com/qt/qttools/blob/fb96051e4200550f5deafae9f2c775d894eb9c7f/src/linguist/shared/numerus.cpp#L61-L63

In the end this PR does actually bring it in sync with Qt.

unho:
@nijel lgtm

PR #3953, opened 2020-03-18
ts2: create translation element in _gettargetnode

nijel:
This way all callers can rely on it being present avoiding
errors when operating on None.

Addresses crash described in https://github.com/WeblateOrg/weblate/issues/3603#issuecomment-599529168


unho:
@nijel lgtm

nijel:
Thanks for review.

PR #3954, opened 2020-03-19
Msgfmt to use the same byte order as pocompile

scarabeusiv:
Fixes testing on s390 and others.


nijel:

Issue #3958, opened 2020-03-28
setup.py fails with py2exe


a1s:
I've got this error from setup.py:

```
Traceback (most recent call last):
  File "setup.py", line 526, in <module>
    standardsetup("translate-toolkit", translateversion)
  File "setup.py", line 486, in standardsetup
    translatebashscripts)
  File "setup.py", line 490, in dosetup
    from setuptools import setup
  File "C:\Python27\lib\site-packages\setuptools\__init__.py", line 20, in <module>
    from setuptools.dist import Distribution, Feature
  File "C:\Python27\lib\site-packages\setuptools\dist.py", line 336, in <module>
    _Distribution = get_unpatched(distutils.core.Distribution)
  File "C:\Python27\lib\site-packages\setuptools\monkey.py", line 44, in get_unpatched
    return lookup(item)
  File "C:\Python27\lib\site-packages\setuptools\monkey.py", line 61, in get_unpatched_class
    raise AssertionError(msg)
AssertionError: distutils has already been patched by <class py2exe.Distribution at 0x047BBF48>
```

a1s:
This helps: https://stackoverflow.com/a/45133496

nijel:
I've tried to do that simple change you've suggested, but apparently the py2exe integration is broken way more, so I give up. See https://github.com/translate/translate/pull/3959

a1s:
As far as I understand, what is there now was made for Python2.  Python3 is different.

If there are no plans to get py2exe working with Python3, the whole block could be put under
`if sys.version_info()[0] < 3:`

I have checked out the [nijel:setup-py2exe](https://github.com/nijel/translate/tree/setup-py2exe) branch, and both `bdist_wininst` and `py2exe` commands work without errors in Python2, provided that I revert the changes made to `BuildCommand` and `Distribution` - that's lines 190 and 191.

nijel:
I have exactly zero knowledge of py2exe and I tried to make it work with all supported versions. For Python 2, the compatible py2exe version is even not available in pypi.org. Feel free to pickup where I stopped and make it work at least for Python 2, but I'm not really interested in working on Python 2 only solution.

a1s:
py2exe compatible with Python2 indeed is on PyPI: https://pypi.org/project/py2exe/0.6.9/

nijel:
The release is there, but it has no files attached, so `pip install py2exe==0.6.9` cannot work.

a1s:
[...]

PR #3959, opened 2020-03-30
Fix py2exe

nijel:
Add tests for py2exe and fix invocation. See https://github.com/translate/translate/issues/3958


nijel:
Apparently, this is broken way more than I thought...

PR #3960, opened 2020-03-30
Setup: Import setuptools before py2exe

### LINKED ISSUE #3958: setup.py fails with py2exe

nijel:
Fixes #3958

This time without tests and other changes, see https://github.com/translate/translate/pull/3959 and https://github.com/translate/translate/issues/3958

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ import re
 import sys
 from distutils.sysconfig import get_python_lib
 from os.path import dirname, isfile, join
+from setuptools import setup
 
 try:
     from sphinx.setup_command import BuildDoc
@@ -487,7 +488,6 @@ def standardsetup(name, version, custompackages=[], customdatafiles=[]):
 
 
 def dosetup(name, version, packages, datafiles, scripts, ext_modules=[]):
-    from setuptools import setup
     description, long_description = __doc__.split("\n", 1)
     kwargs = {}
     if py2exe:


nijel:
The ideal approach here would be to include this in the tests. I've tried in #3959, but it's more complex:

* The py2exe on Python 3 needs major changes
* The Python 2 and 3 versions seems incompatible
* The Python 2 version cannot be installed from PyPI, but has to be installed from sourceforge.net

These are addressable, but beyond my interest, so I'll merge this simple fix which should address the Python 2 case.

PR #3963, opened 2020-04-03
json: allow usage of BOM in JSON files

nijel:
Such files are violating RFC, but it suggests that parses should be able
to skip this, see https://tools.ietf.org/html/rfc7159#section-8.1


nijel:
Thanks for review.

PR #3967, opened 2020-04-09
Setup tests

nijel:
Add some basic tests for creating distribution and verifying it is built correctly.

Also fixes several issues found by this.

### DIFF
--- a/.github/workflows/setup.yml
+++ b/.github/workflows/setup.yml
@@ -1,0 +1,46 @@
+name: Distribution
+
+on: [push, pull_request]
+
+jobs:
+  setup:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Checkout submodules
+      shell: bash
+      run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/setup.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        pip install -r requirements/dev.txt
+        pip install twine virtualenv
+    - name: Build sdist
+      run: make build
+    - name: Twine check
+      run: twine check dist/*
+    - name: Install test
+      run: |
+        virtualenv test-ttk-release
```
julen, line 43: Will this use system Python (2.7 I guess) or the 3.7 runtime that was just installed? You may need to use `vitualenv -p python3.7`

nijel: This will use Python defined above in the workflow, so 3.7.

nijel:
As for the indexing, this is code I never touched so far, and I have no clue about that :-). It broke build of the documentation for me, that's why I was fixing it here.

PR #3970, opened 2020-04-14
Android: Workaround broken plural handling

nijel:
In some Android builds, the plurals which do not have "others" are not handled
properly and cause application to crash (android.content.res.Resources$NotFoundException).
Adding "others" as a copy of "many" should address this.


unho:
lgtm

nijel:
Thanks for review.

PR #3973, opened 2020-04-15
Use GitHub actions for tests, lint and docs build

nijel:
This consolidates CI setup from Travis CI and AppVeyor to single service.

The GitHub Actions have much more parallel jobs than other services (20 concurrent jobs compared to 4 on Travis and 1 on AppVeyor), so all the tests can be executed at once, not waiting on consequent runs.


nijel:
The only failing check is from appveyor, which should be disabled in case this PR gets merged as it's configuration is removed.

nijel:
Thanks for review!

@unho Can you please remove https://ci.appveyor.com/project/unho/translate ? It will now produce just failures as its configuration is gone.

julen:
@nijel I'll remove the webhook event on the GitHub side, so hopefully that keeps it away from now on.

nijel:
Thanks, that should do the trick.

unho:
@nijel I never had access to appveyor.

nijel:
Most likely it is registered on you as the project has your username is in the URL, it has GitHub authentication, you can try it... https://ci.appveyor.com/project/unho/translate

unho:
IIRC Dwayne setup this and never shared credentials. That link exists because I somehow granted access to Appveyor on my fork, but I don't have any user on Appveyor.

nijel:
That is not used on your fork, but on this repo :-). AppVeyor doesn't deal with organisations, it's all under person who did enable the repo (for example https://github.com/gammu/gammu/ is built on https://ci.appveyor.com/project/nijel/gammu).

unho:
Then that would be @dwaynebailey IIRC.

PR #3977, opened 2020-04-16
chore: Use built-in submodule fetchin for checkout

nschonni:


nijel:
Good point., I didn't notice they added it back in https://github.com/actions/checkout/commit/422dc4567157f4d62b665a8a288310365b1d194b

PR #3979, opened 2020-04-18
chore: Only build docs on change

nschonni:
Skips the build for PRs that don't touch anything related to the docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,9 +1,23 @@
 name: Docs
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - "Makefile"
+      - ".github/workflows/docs.yml"
+      - "docs/**"
+      - "requirements/**"
+      - "tests/cli/example_test.sh"
+  pull_request:
+    paths:
+      - "Makefile"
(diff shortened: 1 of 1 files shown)


nijel:
I used to play a lot with filtering, but it comes at cost of not detecting some issues. 

For example, in docs there is:

```rst
.. literalinclude:: ../../tests/cli/example_test.sh
```

So in case this file is removed in the PR, the docs build is skipped and nobody notices that it broke docs. That will be noticed at some point later, either when somebody touches the docs or when releasing. And point of CI is to avoid such errors.

So either you can carefully build the paths list and take extra effor to keep it up to date, or just give up and build the docs on every commit and waste some resources here. In most of my projects I gave up on filtering after being hit by such issues :-).

nschonni:
Thanks, looks like that is the only `literalinclude` in the project.
Looking at it, I think that file could be inlined from the docs. It doesn't really have much that someone wouldn't get from copying another test as a starting point.

nschonni:
I've add that imported file into the globs. i didn't find any others doing the same thing.
Not sure what the failure for codecov is about

nijel:

PR #3980, opened 2020-04-18
chore: Only run lint if python files are changed

nschonni:
Also included the setup.cfg which seems to hold the settings.


nijel:

PR #3981, opened 2020-04-18
OS specifc GitHub Actions

nschonni:
Nothing wrong with the current approach, was just playing with the filtering.
Also noticed that the VScode linter doesn't like the interpolation for the `runs-on` value

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,6 +1,12 @@
 name: Test on Linux
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+    - "docs/**"
+  pull_request:
+    paths-ignore:
+    - "docs/**"
 
 jobs:
   test:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -47,10 +47,6 @@ jobs:
         make requirements/min-versions.txt
         cat requirements/min-versions.txt
         pip install -r requirements/min-versions.txt
-    - name: Compile all the files
-      run: python -m compileall -q -f .
-    - name: Setup build
-      run: python setup.py --quiet build
     - name: Install
       run: pip install .
     - name: py.test
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,34 +1,19 @@
-name: Test
+name: Test on Linux
 
 on: [push, pull_request]
 
 jobs:
   test:
 
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu, windows, macos]
         requirements: [latest]
         python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
-        exclude:
-          - os: windows
-            python-version: 2.7
-          - os: windows
-            python-version: 3.5
-          - os: windows
-            python-version: 3.6
-          - os: windows
-            python-version: 3.7
-          - os: macos
-            python-version: 3.6
-          - os: macos
-            python-version: 3.7
         include:
-          - os: ubuntu
+          - os: ubuntu-latest
             python-version: 2.7
             requirements: minimal
-    name: ${{ matrix.os }}, Python ${{ matrix.python-version }}, ${{ matrix.requirements }}
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


nijel:
This duplicates most of the logic into three files, why do you think it's better than having all in one place?

nschonni:
Some parts at the end are duplicated, but since the matrix don't really overlap nor the setup steps, I (personal opinion) find it easier to follow what is going on in each environment

nijel:
I think the matrix should ideally overlap in the ideal case. There were some issues with Windows builds that made me skip some of the Python versions, but ideally we would test all supported Python versions on Windows as well.

nschonni:
True, but at this point it's only ~50% shared at this point. If the windows testing and all the python versions overlap, it would make sense just leaving the brew/apt as the only branching logic in the file.

I'd also recommend pulling out the "minimal" stuff to its' own job. I didn't really understand what it was doing though so I didn't touch it.

The nice part of the Actions vs. Travis, is that you can split this out, and don't have to shoehorn flow logic into yaml 😄

nijel:
The only difference of minimal job is installing minimal required versions.

.github/workflows/linux.yml
```diff
@@ -1,31 +1,24 @@
...
 
 jobs:
   test:
 
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu, windows, macos]
+        os: [ubuntu-latest]
```
nijel, line 17: Why keeping os in the matrix when there is just one?

.github/workflows/win.yml
```diff
@@ -0,0 +1,50 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Update pip
+      run: pip install --upgrade pip wheel
+    - name: Install pip dependencies
+      run: pip install -r requirements/dev.txt
+    - name: Compile all the files
+      run: python -m compileall -q -f .
+    - name: Setup build
+      run: python setup.py --quiet build
```
nijel, line 40: When doing the cleanup here, those are probably not needed in none of the actions. The setup.py build is tested in separate workflow and it should byte compile the files as well (I guess the reasoning behind this was syntax check, but we do have lint job as well).

```suggestion
```

.github/workflows/linux.yml
nschonni: This could definitely be inlined, I just kept to avoid updating the caching and coverage keys

nschonni: Done

.github/workflows/win.yml
nschonni: Interesting, the build looks to be failing after removing this. goin to try and add it back to see if it works again

nschonni: Adding it back fixed it, but here is the failure when both these steps were removed https://github.com/translate/translate/runs/610261179?check_suite_focus=true#step:8:13 if that rings a bell why

nschonni: And failures when only Setup step is removed https://github.com/translate/translate/pull/3981/checks?check_run_id=610316715#step:11:17

nijel: That's certainly bug in the code, but I don't get why byte-compiling hides it. I've tried to address this in https://github.com/translate/translate/pull/3984

nschonni: OK, after that lands I can try and remove the steps again

nijel: I've just merged that.

nschonni: Looks like it's catching something else now https://github.com/translate/translate/pull/3981/checks?check_run_id=610959605#step:10:28

nschonni: OK, add a suggested fix from pylint. Hope it makes sense

translate/misc/selector.py
```diff
@@ -125,7 +125,7 @@ def add(self, path, method_dict=None, prefix=None, **http_methods):
         
         Prefix will override self.prefix for this mapping.
         """
-        # Thanks to S�bastien Pierre 
+        # Thanks to S�bastien Pierre
```
nijel, line 128: This seems wrong, encoding issue?

translate/misc/selector.py
```diff
@@ -282,7 +282,7 @@ def _parse_line(self, line, path, methods):
 
 
 class SimpleParser(object):
-    """Callable to turn path expressions into regexes with named groups.
+    r"""Callable to turn path expressions into regexes with named groups.
```
nijel, line 285: Seems I've missed this one in https://github.com/translate/translate/pull/3984

translate/misc/selector.py
nijel, line 128:
```suggestion
        # Thanks to Sébastien Pierre 
```

nijel:
Can you please squash the fix it to your commit?

nschonni:
Done, not sure what the encoding issue was

translate/misc/selector.py
```diff
@@ -125,7 +125,7 @@ def add(self, path, method_dict=None, prefix=None, **http_methods):
         
         Prefix will override self.prefix for this mapping.
         """
-        # Thanks to S�bastien Pierre 
+        # Thanks to Sébastien Pierre
```
nijel, line 128:
```suggestion
        # Thanks to Sébastien Pierre
```
Whitespace issue? This should not be on the diff :-)

nijel:

PR #3982, opened 2020-04-21
yaml: Utilize round trip features of ruamel.yaml

### LINKED ISSUE #3695: Advanced Yaml support

nijel:
- use parsed file when saving
- this preserves comments and syntax from the original file
- avoid using merged keys, we do not want to duplicate them

Fixes #3695

### DIFF
--- a/docs/formats/yaml.rst
+++ b/docs/formats/yaml.rst
@@ -22,20 +22,6 @@ Non-Conformance
 The following are not yet supported (in most cases these are properly parsed,
 but not saved in round trip):
 
-* Abbreviated lists:
-
-  .. code-block:: yaml
-
-      first-name: [Sun, Mon, Tue, Wed, Thu, Fri, Sat]
-
-
-* Abbreviated dictionaries:
-
-  .. code-block:: yaml
-
-      martin: {name: Martin D'vloper, job: Developer, skill: Elite}
-
-
 * Booleans:
 
   .. code-block:: yaml
--- a/translate/storage/test_yaml.py
+++ b/translate/storage/test_yaml.py
@@ -3,7 +3,6 @@
 from __future__ import unicode_literals
 
 import sys
-from io import BytesIO
 
 import pytest
 import ruamel.yaml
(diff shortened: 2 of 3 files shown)


nijel:
Thanks for review.

docs/formats/yaml.rst
```diff
@@ -22,20 +22,6 @@ Non-Conformance
...
-  .. code-block:: yaml
-
-      first-name: [Sun, Mon, Tue, Wed, Thu, Fri, Sat]
-
-
-* Abbreviated dictionaries:
-
-  .. code-block:: yaml
-
-      martin: {name: Martin D'vloper, job: Developer, skill: Elite}
```
unho, line 36: If these are removed from non-conformance section of docs I guess it is because now Translate Toolkit supports them. If yes, then please add these before the non-conformance section to clearly state this syntax is supported.

nijel: There is nothing like this in the docs so far and in previous fixes I've also just removed sections here (for example in 4012b06506bb99f8275d9a63dc077cc3c132a96f).

I can add something generic like all YAML syntax is supported...

unho: The idea is to have two sections (Conformance and Non-Conformance) for all formats (when deemed necessary) like for http://docs.translatehouse.org/projects/translate-toolkit/en/latest/formats/php.html

nijel: With PHP it makes more sense because there is just limited subset of the language supported (in the end it's generic programming language, so that's expected). With formats like YAML or JSON I see no reason to describe whole language specification that it should work.

unho: Ok, that is a reasonable point. Lets move on.

Issue #3964, opened 2020-04-05
Drop support for Python 2


unho:
Since January 1st 2020 Python 2 is no longer supported. The last release was already really painful to get out due to several dependencies already switched to Python 3 only.

Switching to Python 3 only might allow to simplify code in some places, update and simplify requirements. On the other hand it might require working on providing an alternative for some of the code that works only in Python 2 only (INI format comes to mind).

unho:
@nijel Any problem on your side with this?

nijel:
I'm fine with this - upcoming Weblate release will be Python 3 only as well.

This will break some things. Besides INI you've mentioned the  py2exe integration currently works only on Python 2 (and is not covered by tests), see https://github.com/translate/translate/pull/3960#issuecomment-609009633.

unho:
I am pretty sure more stuff needs work as well.

rffontenelle:
This will break Virtaal too. I don't think this should be an obstacle for dropping it, but just mentioning.

nijel:
The iniparse module now supports Python 3, so that bit should be addressed, see https://github.com/translate/translate/pull/3971

julen:
I'll soon have some spare time to contribute (hi Corona) so I could take on some of the effort here if needed. I recently did a py2 -> py3 migration for [Zing](https://github.com/evernote/zing) and dropped py2 support along the way, so I'm familiar with a good part of the required changes.

nijel:
My proposal would be to release current master this weekend as 2.5.1 (it seems to contain just bufixes compared to previous release, see https://github.com/translate/translate/issues/3966 and https://github.com/translate/translate/pull/3975) and then drop Python 2 support in master, possibly do some other intrusive changes and call next release 3.0.0. For example this can include code formatting, see https://github.com/translate/translate/pull/3989. I would also like to address https://github.com/translate/translate/issues/3819 and related issues https://github.com/translate/translate/issues/3857).

unho:
@nijel I like the plan. I plan to start reviewing PRs today.

PR #3986, opened 2020-04-23
MO: Avoid using deprecated array.tostring

nijel:
This requires Python 3, see https://github.com/translate/translate/issues/3964


julen:
This will immediately break Python 2 support. Could we maybe do such changes in a separate py3 branch?

nijel:
This should be merged once we decide to drop Python 2 support. I've just pushed it because I found the local branch with the change (it was done year ago) and didn't want that to be forgotten.

unho:
lgtm, but I strongly suggest dropping Python 2 support before landing this PR, since it feels correct to have all green before landing.

PR #3988, opened 2020-04-23
[requires.io] dependency update on master branch

unho:


nijel:
I've just deactivated requires.io, it only breaks things for python < 3.0.

unho:
@nijel Great!

unho:
@nijel Disabling it kinda broke the requirements status flag displayed at top of README. Does dependabot provide an alternative to this?

nijel:
Yes, see https://github.com/translate/translate/pull/3992

PR #3989, opened 2020-04-24
chore: Run autopep8

nschonni:
Was looking at pylint, and it pointed it over here for some cleanups. Besides rolling the boolean operators onto the next line, it also seems to re-write `type` checks to `isinstance`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,7 +42,7 @@ extensions = [
 ]
 
 # Display todo notes. See http://sphinx-doc.org/ext/todo.html#directive-todo
-todo_include_todos=True
+todo_include_todos = True
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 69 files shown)


nijel:
I have no experience with autopep8, but using black on most of my projects, do you know differences between these two? Did you have any reasons for choosing autopep8?

nschonni:
I don't have any experience with it either 😄 Was just playing around with it to see if it's useful. I used the "aggressive" option, which may be why it started to make some breaking changes.
If you've got more experience with `black` that might be a better way to go.
I was trying to try out https://www.pylint.org/, but since it doesn't have an autoformatter, i was trying this one auto to do a first pass to see how much it would address

nijel:
I do have experience with black and I like it (especially integrated with [pre-commit](https://pre-commit.com/)). But it will probably create a *lot* of changes on existing code base. Using any code formatter on existing code base which was not really following strict rules will lead to this, so we should clarify what we are going to use to avoid doing such big formatting changes several times.

nschonni:
For sure, I wasn't expecting this to land, just seeing if any of the changes made sense to submit

PR #3993, opened 2020-04-24
lisa: remove not needed assertion

nijel:
The case when there are no language nodes is properly handled in the
code, remove the assertion.


nijel:
The assert did make sense when introduced in https://github.com/translate/translate/commit/029b95195296263aa72eac8bca41aaec738e3ffc#diff-d44ee0984af59a98f193dfd2175508f1R168, because the would not handle this properly, but the code was later changed to work fine in this case, but the assert stayed.

unho:
@nijel lgtm

Issue #3884, opened 2019-03-03
[html2po] translate language version in <html lang="en">


fabianegli:
The `lang` attribute form the html tag is not extracted for translation - I think it is obvious that it should be.

anderskaplan:
I'd be happy to provide a pull request for this change, but I'd like to get the maintainers' opinon on it first. The reason is that there is a unit test to check that the lang tag is *not* extracted for translation. (test_html2po.py line 60)

The decription of the test says "test to ensure that we no longer use the lang attribure" (sic), but does not give any hints to the reason for it.

I have a use case where I want to translate static html pages. It would be useful to swap out the lang attribute at the same time as swapping out the translated content.

PR #3994, opened 2020-04-24
[html2po] Extract the lang attribute from the html tag (#3884).

anderskaplan:
Solves #3884.

While the change itself is simple enough, I'd still like to get the maintainers' opinon on it. The reason is that there is a unit test to check that the lang tag is not extracted for translation. (test_html2po.py line 60)

The decription of the test says "test to ensure that we no longer use the lang attribure" (sic), but does not give any hints to the reason for it.

I have a use case where I want to translate static html pages. It would be useful to swap out the lang attribute at the same time as swapping out the translated content.


nschonni:
The only reason I can think that the `lang` shouldn't be translated would be for non-`html` element uses like `<span lang="fr">Français</span>` for literal alternate language blocks in documents

anderskaplan:
Good point. I'll make sure to handle that distinction in an upcoming update.

anderskaplan:
A better bug fix submitted as a separate PR.

PR #3996, opened 2020-04-25
docs: Update releasing instructions

nijel:
- adjust list of files with version number
- adjust website files to update
- remove no longer used announcement steps


unho:
gtm

PR #3998, opened 2020-04-28
poxliff: support non numeric ids on plurals

nijel:
It is used in the real world and the specification is not clear on this.


translate/storage/test_poxliff.py
```diff
@@ -72,6 +72,23 @@ def test_parse(self):
         assert xlifffile.units[0].source == "cow"
         assert xlifffile.units[0].source == multistring(["cow", "cows"])
 
+    def test_parse_alpha_id(self):
```
unho, line 75: test_parse_alpha_id -> test_parse_plural_alpha_id

PR #3999, opened 2020-04-28
Drop support for Python 2

### LINKED ISSUE #3964: Drop support for Python 2

nijel:
* Update requirements
* Adjust documentation
* Remove `__future__` imports
* Remove six usage

We could certainly do more cleanups, but I wanted to stick with the more straightforward ones here. Please let me know if you want to split up this PR or squash commits somehow.

Fixes #3964


translate/storage/test_pypo.py
```diff
@@ -238,17 +236,14 @@ def test_output_str_unicode(self):
         assert len(pofile.units) == 1
         print(bytes(pofile))
         thepo = pofile.units[0]
-        # On Python 2, str() returns bytestrings while on Python 3, str() returns unicode
-        assert str(thepo) == posource if six.PY3 else posource.encode("UTF-8")
-        assert six.text_type(thepo) == posource
+        assert str(thepo) == posource
+        assert str(thepo) == posource
```
julen, line 240: One of these assertions can be removed now.

nijel: Removed. I really didn't look at most of the code, so it was easy to miss such cases.

unho:
@nijel Amazing job. Just a nit (sorry), can you please also drop `str_prefix` variable in `translate/misc/test_multistring.py` since it is now unnecessary?

PR #4002, opened 2020-04-29
chore: Remove required.txt

nschonni:
Remove references from setup and other requirements.txt files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -397,27 +397,6 @@ else:
         return data_file
 
 
-def parse_requirements(file_name):
-    """Parses a pip requirements file and returns a list of packages.
-
-    Use the result of this function in the ``install_requires`` field.
-    Copied from cburgmer/pdfserver.
-    """
-    requirements = []
-    with open(file_name, 'r') as fh:
-        for line in fh:
-            # Ignore comments, blank lines and included requirements files
-            if re.match(r'(\s*#)|(\s*$)|(-r .*$)', line):
-                continue
-
-            if re.match(r'\s*-e\s+', line):
-                requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
-            elif not re.match(r'\s*-f\s+', line):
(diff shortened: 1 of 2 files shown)


nschonni:
Linux failures seem to be an unrelated `apt` failure, but I can't restart the job. CodeCov also seems incorrect

nijel:
I've restarted the jobs now (it should address the coverage issue as well as coverage from the Linux jobs was missing).

I've kept it in https://github.com/translate/translate/pull/3999 to keep it simple and I also wanted to open up discussion on the deps overall, see https://github.com/translate/translate/issues/4004 for that.

nschonni:
No worries, if you end up going that way, feel free to close this. Just noticed the dead file on that other PR

nijel:
I'm not sure what is best approach in this case, you're welcome to provide feedback in https://github.com/translate/translate/issues/4004

setup.py
```diff
@@ -510,7 +489,6 @@ def dosetup(name, version, packages, datafiles, scripts, ext_modules=[]):
         scripts=scripts,
         ext_modules=ext_modules,
         cmdclass=cmdclass,
-        install_requires=parse_requirements('requirements/required.txt'),
```
unho, line 513: This probably should be kept but instead point to `recommended.txt` like https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies @nijel thoughts on this?

unho: I see that this is kinda proposed in #4004

nijel:
Here is PR for making lxml required: https://github.com/translate/translate/pull/4035 I think it's reasonable as a lot of the code uses it, but this is something what should get broader consensus. I don't see any other dependency which should be promoted, so in case that one is not accepted, let's merge this one.

nijel:
Closing this one as lxml will be required, see https://github.com/translate/translate/issues/4048#issuecomment-642892426 and #4035.

nschonni:
K, not a problem

Issue #3768, opened 2018-03-02
html2po incorrectly strips html


deAtog:
Given a block element that contains inline html, e.g. a paragraph containing links, html2 po will incorrectly strip the inline html in some cases: E.g:

```html
<p><a href="http://example.com">This is a link</a> but this is not. <a href="http://example.com">However this is too</a></p>
```

Will be transformed into the following string for translating:

`This is a link</a> but this is not. <a href="http://example.com">However this is too`

From the comments in the code, it appears this functionality was included to strip out what the author considered to be unnecessary html in the translation text. The regular expression used for this is unable to accurately determine which end tag matches the start tag.

Since html2po does not support the insertion of placeholders, the only solution here is to leave the inline html intact regardless of how unnecessary it may be for a translator.

anderskaplan:
Can you please explain what your use case is, i.e., why doesn't it work for you to have the leading and trailing tags removed?

ngrilly:
Because it makes difficult using html2po to extract strings to be translated and then po2html to replace the untranslated strings by the translated strings. In some languages, the order of words can be different from English, which means the <a>...</a> is moved in a different position in the sentence. With the opening or the closing tag missing, it's not possible to move it.

Issue #3416, opened 2016-03-30
po2html incorrect sub-strings substitution


s-n-g:
This is the situation at hand:

I have two messages

msgid "This is me <b>**Writing in English**</b>"
msgid "**Writing in English**"

Both messages are translated, but po2html result is (within the html file):

This is me (**<b>Translation of "Writing in English**"</b>)
and
(**Translation of "Writing in English**")

As I understand it, there is a sub-string substitution going on for the first message (when it shouldn't).
I have tried to overcome by sorting the po file by message length so that the first message would be substituted first, but no luck...

dwaynebailey:
@s-n-g yes I'm pretty sure what you've observed is correct, we do string substitution.  What would really help is if you could supply a really small test case of the PO and HTML files to replicate the issue.

s-n-g:
Sure, here you are:  [issue3416.zip](https://github.com/translate/translate/files/196399/issue3416.zip)

Commands executed:
`html2po -P html pot`
`mkdir po`
`cp pot/* po`

And after translating **index.po**:
`po2html -t html po elhtml`

PR #4005, opened 2020-05-01
[html2po] multiple bug fixes, parsing makeover

anderskaplan:
A makeover of the HTML parsing to fix issues #3768 (html2po incorrectly strips html),
#3416 (incorrect sub-strings substitution), and a few other issues for which there are no bug reports. Unit tests have been added to expose those cases.

The new parsing algorithm does not use any regex-based parsing on top of the HTML parser. Instead, it puts the content inside localisable HTML elements on a queue which is processed when the end of the localisable content is reached. This means that one can use the DOM of the translatable content when extracting localisable content, which makes it easier to handle edge cases.

Also includes a fix for #3884 (translate language version in <html lang="en">). This fix is better than the PR I submitted earlier, because it does not extract the lang tag from other HTML elements.

Re-enabled disabled unit tests for po2html. Split some unit tests to do one test per method.

Apologies for the large pull request -- it would have been difficult to split this PR into smaller parts.

### DIFF
--- a/translate/convert/test_po2html.py
+++ b/translate/convert/test_po2html.py
@@ -57,27 +57,57 @@ sin.
 </body>'''
         assert htmlexpected.replace("\n", " ") in self.converthtml(posource, htmlsource).replace("\n", " ")
 
-    @mark.xfail(reason="Not Implemented")
+    def test_replace_substrings(self):
+        """Should replace substrings correctly, issue #3416"""
+        htmlsource = '''<!DOCTYPE html><html><head><title>sub-strings substitution</title></head><body>
+<h2>This is heading 2</h2>
+<p>The heading says: <b>This is heading 2</b></p>
+</body></html>'''
+        posource = '#:html.body.h2:2-1\nmsgid "This is heading 2"\nmsgstr "Αυτή είναι μία Ετικέτα 2"\n\n#html.body.p:3-1\nmsgid "The heading says: <b>This is heading 2</b>"\nmsgstr "Η ετικέτα λέει: <b>Αυτή είναι μία Ετικέτα 2</b>"\n'
+        htmlexpected = '''<!DOCTYPE html><html><head><title>sub-strings substitution</title></head><body>
+<h2>Αυτή είναι μία Ετικέτα 2</h2>
+<p>Η ετικέτα λέει: <b>Αυτή είναι μία Ετικέτα 2</b></p>
+</body></html>'''
+        assert htmlexpected in self.converthtml(posource, htmlsource)
+
+    def test_attribute_outside_translatable_content(self):
+        htmlsource = '<img alt="a picture"><p>A sentence.</p>'
+        posource = '''#: html:3\nmsgid "A sentence."\nmsgstr "'n Sin."\n#: html:1\nmsgid "a picture"\nmsgstr "n prentjie"\n'''
+        htmlexpected = '''<img alt="n prentjie"><p>'n Sin.</p>'''
+        assert htmlexpected in self.converthtml(posource, htmlsource)
+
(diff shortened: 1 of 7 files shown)


nijel:
LGTM, but I'd like to get this reviewed by somebody else as well.

anderskaplan:
great, just let me know if you have any questions

translate/convert/test_po2html.py
```diff
@@ -57,27 +57,57 @@ def test_linebreaks(self):
...
 
-    @mark.xfail(reason="Not Implemented")
+    def test_replace_substrings(self):
+        """Should replace substrings correctly, issue #3416"""
+        htmlsource = '''<!DOCTYPE html><html><head><title>sub-strings substitution</title></head><body>
+<h2>This is heading 2</h2>
+<p>The heading says: <b>This is heading 2</b></p>
+</body></html>'''
+        posource = '#:html.body.h2:2-1\nmsgid "This is heading 2"\nmsgstr "Αυτή είναι μία Ετικέτα 2"\n\n#html.body.p:3-1\nmsgid "The heading says: <b>This is heading 2</b>"\nmsgstr "Η ετικέτα λέει: <b>Αυτή είναι μία Ετικέτα 2</b>"\n'
+        htmlexpected = '''<!DOCTYPE html><html><head><title>sub-strings substitution</title></head><body>
```
unho, line 67: So `title` elements are not yet translatable?

translate/storage/html.py
```diff
@@ -202,18 +158,10 @@ def __init__(self, includeuntaggeddata=None, inputfile=None,
...
 
-    ENCODING_RE = re.compile(br'''<meta.*
-                                content.*=.*?charset.*?=\s*?
-                                ([^\s]*)
-                                \s*?["']\s*?>
-                             ''', re.VERBOSE | re.IGNORECASE)
-
     def guess_encoding(self, htmlsrc):
         """Returns the encoding of the html text.
-
```
unho, line 213: This blank line should stay.

unho:
@anderskaplan I am gonna give a lgtm. Anyway left some nits.

It was really painful to review this PR because of that never ending commit. I very much appreciate to review shorter and more focused commits, preferably with a diff below 100 changes, even if there are dozens of commits. I get that this commit was too large because of the massive change, but next time I would suggest placing changes in separate commits (for this type of large commits) like (in this particular order):

* Documentation cleanups.
* Code cleanups.
* Tweaks to existing tests (that do not require altering the tested code).
* Split up existing tests.
* New tests (that do not require altering the tested code).
* New tests (marking them as failing tests since new code providing the functionality hasn't yet been added).
* Actual code changes (and if possible split over several commits, if possible and necessary). Marking tests as passing because of these changes is acceptable in this commit too.
* Documentation changes regarding the new behavior.

Doing it like this will allow for easier, less painful and faster reviews, and also would allow us to merge chunks of commits that have been approved (that is why cleanups should be done first, if possible, before the actual changes).

translate/convert/test_po2html.py
anderskaplan: yes they are; there should be a translation unit for the title as well.
however this test is only checking that we're not mixing up translations when one is a substring of the other.

anderskaplan:
Fair enough. I usually try to avoid big changes like this one for exactly that reason, but in this case it wasn't possible. Thanks for the tip about the commit sequence.

unho:
@anderskaplan Can you please rebase this PR?

translate/storage/html.py
anderskaplan: Fixed

unho:
lgtm @nijel can you merge?

nijel:

PR #4007, opened 2020-05-03
chore: Don't build dependabot branches

nschonni:
This prevents the double build for DependaBot PRs, so they only run when it opens a PR, and not when it creates a branch


nijel:

PR #4010, opened 2020-05-04
PHP improvements

nijel:
* Handle some missing syntax
* Improve keeping quoting style
* Add support for Laravel plurals

### DIFF
--- a/translate/storage/test_php.py
+++ b/translate/storage/test_php.py
@@ -3,6 +3,7 @@
 from pytest import mark
 
 from translate.misc import wStringIO
+from translate.misc.multistring import multistring
 from translate.storage import php, test_monolingual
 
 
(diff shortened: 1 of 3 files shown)


translate/storage/test_php.py
```diff
@@ -484,57 +484,78 @@ def test_parsing_arrays_declared_in_a_single_line(self):
...
+        phpsource = """<?php
+$days = array(
+    'Sunday',
+    'Monday',
+    'Tuesday',
+    'Wednesday',
+    'Thursday',
+    'Friday',
+    "Saturday",
+);
```
unho, line 497: Doesn't it work if it is kept in a single line? I am asking because that is how it was found.

nijel: Parsing works just fine, but serializing formats it this way, that's why I've changed it.

unho: Ok, I guess we will have to allow this to happen. On the other PHP code should be like this, so I guess it is not that bad.

unho:
@nijel lgtm

PR #4012, opened 2020-05-05
Remove StringIO wrapper

nijel:
It is confusing to have StringIO which is actually BytesIO.

### DIFF
--- a/translate/misc/wStringIO.py
+++ b/translate/misc/wStringIO.py
@@ -17,31 +17,17 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, see <http://www.gnu.org/licenses/>.
 
-"""A thin wrapper around BytesIO that accepts and auto-convert non bytes
-input
-"""
-
 from io import BytesIO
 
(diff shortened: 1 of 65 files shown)


translate/misc/wStringIO.py
```diff
@@ -17,31 +17,17 @@
...
-class CatchStringOutput(StringIO):
+class CatchStringOutput(BytesIO):
     """catches the output before it is closed and sends it to an onclose method
     """
 
     def __init__(self, onclose):
         """Set up the output stream, and remember a method to call on closing
         """
-        StringIO.__init__(self)
+        super().__init__()
```
unho, line 30: I guess this is a `super` you missed in the other PR.

unho:
@nijel lgtm

nijel:
Thanks for review.

translate/misc/wStringIO.py
nijel: Probably, it's easy to miss something in those significant changes.

PR #4013, opened 2020-05-07
Avoid using deprecated logging.warn

nijel:


unho:
@nijel lgtm

nijel:
Thanks for review.

PR #4017, opened 2020-05-10
Filters: Use functools.lru_cache instead of own implementation

nijel:
* The cache is now shared between source and target and the size is
  slightly larger.
* The spell checker returns lists instead of being generator. Before the
  target caching was broken as it cached the generator, so it was
  effectively empty for cache hits.


unho:
This touches parts of the codebase I am not familiar with, and for which we do not have testing. I'm  a bit scared about these changes might break something. Can you maybe add tests for the touched code first?

nijel:
I will look into that.

nijel:
I've fixed existing test for that in 3e273b3c3ad4d50fc0ffb26d7acb176b67361d3d, is that good enough?

nijel:
I didn't  find it in the current code, but it seems to be indeed used in older versions, it was removed in https://github.com/translate/pootle/commit/454825975af461bc47c987630474a032b17d827d

unho:
@nijel lgtm despite it would be great to have some test for when enchant is not available (that would increase coverage in `translate/filters/spelling.py`)

PR #4024, opened 2020-05-14
Factory: Accept store as parameter to getobject

nijel:
This is useful to inject existing store as output for convertors.

I've originally made similar change only for one convertor in https://github.com/translate/translate/pull/4022 (now removed), but it turns out to be more generic pattern, so I think placing it into the factory makes more sense.


unho:
lgtm

PR #4025, opened 2020-05-14
idml: Fix conversion and add tests

nijel:


nijel:
This needs setup.py integration, similar as in #4022. I will do it once it lands to avoid conflicts.

Issue #4004, opened 2020-04-30
Conditional imports


nijel:
Right now translate-toolkit doesn't have any requirements (the only one was six and that was dropped for the 3.0 release). There are two levels of deps - recommended and optional.

The code sometimes tries to deal with missing deps by catching `ImportError`, but in many cases it is not (for example lxml imports are not guarded in many cases).

I think this would use some cleanups and revisiting:

* Maybe we should make some "recommended" ones required (lxml comes to my mind).
* Merge "recommended" and "optional" to one list.
* Properly expose these in `setup.py` as `extras_require` so that it's possible to depend on that (for example `translate-toolkit[PHP]` to get PHP support)
  * https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

unho:
I fully agree. Anyway I am thinking that it would be great to have some option to get all the dependencies instead of having to install them one by one.

nijel:
I don't think that pip itself can do it, but the list can be generated in setup.py based on the requirements files and it can include something like "full" option having all extras.

unho:
That's what I was suggesting.

unho:
@nijel I don't think this issue is still addressed.

nijel:
What do you think is missing? The optional deps can be specified since https://github.com/translate/translate/pull/4026, there is also option to do `translate-toolkit[recommended]` or `translate-toolkit[all]`.

unho:
Installing using `translate-toolkit[PHP]` (which I can live without) or the merge of recommended and optional requirements are things mentioned in the first comment on this issue.

nijel:
`translate-toolkit[PHP]` should work since #4026

unho:
Sure. Sorry!

nijel:
As for recommended/optional, I think it kind of makes sense, that's why I didn't change this in the end. I can merge these if you think it's better.

unho:
I think it will be less confusing to have them merged.

phlax:
hi - not sure how helpful this is - but seems worth mentioning...

> The code sometimes tries to deal with missing deps by catching ImportError

When i was profiling pootle/ttk i noticed that these imports were causing a lot of diskio and degrading performance. Not sure if this is still the case, or what the impact was exactly

nijel:
@phlax I guess you talk about https://github.com/translate/translate/pull/3492, I unfortunately didn't get to review that, but I want to take a look.

[...]

PR #4026, opened 2020-05-18
Setup: Allow to specify optional deps using pip

nijel:
The full version can be installed by:

    pip install translate-toolkit[all]

See #4004

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,6 +181,30 @@ classifiers = [
 ]
 
 
+# Generate extras requires
+def update_requires(filename, extras_require, recommended):
+    with open(filename) as requirements:
+        for line in requirements:
+            line = line.strip()
+            # Skip comments, inclusion or blank lines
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -181,6 +181,30 @@
...
+                extras_require[section] = [dependency]
+            else:
+                extras_require[section].append(dependency)
+            extras_require["all"].append(dependency)
+            if recommended:
+                extras_require["recommended"].append(dependency)
+
+
+extras_require = {"recommended": [], "all": []}
+update_reqires("requirements/optional.txt", extras_require, False)
```
unho, line 205: reqires?

nijel: It's called `extras_require` in the setup args, so I kept it same here (see below in the patch).

unho: `reqires` != `requires`

nijel: Sorry, I didn't notice that.

unho:
lgtm

PR #4027, opened 2020-05-18
properties: Fix round trip of empty values without delimiter

### LINKED ISSUE #3620: Fix properties roundtrip
After #3607 some changes need to be introduced in order to fix properties roundtrip. More details in https://github.com/translate/translate/pull/3607#issuecomment-291440437

nijel:
Fixes #3620


unho:
@nijel lgtm

PR #4028, opened 2020-05-19
pypo: Bring line wrapping closer to gettext

nijel:
The line wrapping in Getext is more complex and the regexp based
splitter is always an approximation of that, but this makes it behave
more similar to what the original implementation does.

The PO wrapper is now a class instead of monkey patched instance of
generic wrapper.

Fixes https://github.com/WeblateOrg/weblate/issues/3852


unho:
@nijel Really nice! lgtm

PR #4035, opened 2020-05-27
requirements: make lxml required

nijel:
Lot of code uses it without catching the import error and I think it is
used often enough to be required.

Fixes https://github.com/translate/translate/issues/4004

Alternative approach could be to remove now empty requirements.txt, see https://github.com/translate/translate/pull/4002

### DIFF
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -19,15 +19,11 @@
 
 """Parent class for LISA standards (TMX, TBX, XLIFF)"""
 
-
-try:
-    from lxml import etree
-    from translate.misc.xml_helpers import (getText, getXMLlang, getXMLspace,
-                                            namespaced)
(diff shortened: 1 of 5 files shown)


translate/storage/lisa.py
```diff
@@ -19,14 +19,10 @@
...
+from lxml import etree
 
-try:
-    from lxml import etree
-    from translate.misc.xml_helpers import (getText, getXMLlang, getXMLspace,
-                                            namespaced)
-except ImportError as e:
-    raise ImportError("lxml is not installed. It might be possible to continue without support for XML formats.")
-
+from translate.misc.xml_helpers import (getText, getXMLlang, getXMLspace,
```
unho, line 24: Place `translate.misc` import after `translate.lang` one.

PR #4037, opened 2020-06-01
json: Add support for go-i18n

nijel:


julen:
I have worked with Golang before and I'd suggest not to call this generically as providing Golang support because there's no official/stdlib i18n solution in the language and as such it can be misleading.

There's the [go-i18n package](https://github.com/nicksnyder/go-i18n) which is popular, it has its own flavors to extract strings and this PR looks to be aimed at that. For reference, there are also other packages that solely rely on go files like [go-l10n](https://github.com/loctools/go-l10n).

nijel:
Yes, this is modelled for go-i18n. I'll adjust the names.

translate/storage/test_jsonl10n.py
```diff
@@ -38,6 +38,22 @@
     ]
 }
 """
+JSON_GOLANG = b"""[
```
julen, line 41: `JSON_GOI18N` I guess

PR #4041, opened 2020-06-03
add support for go-i18n json files

uniqx:
I'm working my way through adding support for the JSON variant of [go-i18n translation](https://github.com/nicksnyder/go-i18n) files.


uniqx:
Seems this was already fixed: #4037

Issue #4018, opened 2020-05-10
Identify and remove dead code


nijel:
There is probably quite some dead code lying around. It would be great if long time contributors @unho or @julen could point to such so that it can be properly investigated and possibly removed to avoid need to maintain code nobody uses these days.

Also do you know any other major users of the toolkit besides amaGama, Virtaal and Pootle? Those are probably broken by move to Python 3, but we should not break them more than necessary.

Inspired by https://github.com/translate/translate/pull/3967#pullrequestreview-391292203

unho:
One of the big problems with Translate Toolkit is that it keeps together unrelated code. For example we have:
* the storage (used by third party projects and Translate Toolkit),
* the converters,
* other tools and scripts,
* the language filters (for quality check),
* the language data (including filter customizations),
* code used by some other Translate projects, like search and indexing,
* ...
There was a proposal long ago about splitting these into separate projects, but we never got there. I mean, it is not as much as dropping unused code (which probably is being used anywhere), but about moving out to separate apps parts of the codebase that are usually not used.

nijel:
For some things, it's certainly useful to keep them together (storage and converters) as their code is usually heavily coupled. 

My knowledge of rest is minimal, but I'm pretty sure there are many things not used anywhere. This happens to every bigger project as the code tends to accumulate...

julen:
I'm almost certainly sure the code for L20N support can be considered dead, because [L20N is deprecated](https://github.com/l20n) on its own (later on superseded by [Fluent](https://projectfluent.org/)). Likewise, given that Mozilla is actively moving to Fluent, and the Moz-specific tools were maintained as part of the mozilla.locamotion.org effort (now dead), Mozilla-related converters could be considered dead too.

unho:
I agree with dropping the parts @julen talks about in his last comment because L20N was never actually of widespread usage and those tools are now useless. And if anyone actually needs those then an older version of TTK can be used. But I still would keep storage and converters for DTD and other formats that were widely used by Mozilla to ease converting translations (I sometimes convert translations to TMX so they can be reused among translation projects, or used in any other way like extracting terminology).

nschonni:
Might be good to create a CI job (or jobs) that would pull down and run the dependent projects to look for regressions. Don't think it would need to block merging things or tested across all Python versions

unho:
[...]

PR #3529, opened 2016-11-28
[WIP] Add l20n sections parsing support

ta2-1:
This PR allows to parse units with sections.

TODO: handle section names and section comments
Related to #3516

Issue #3471, opened 2016-07-28
Add l20n checker


ta2-1:
We need to add checker for some l20n format specifics:
- [Variables](http://l20n.org/learn/interpolation-variables)
- [Builtins](http://l20n.org/learn/builtins)
- [Entity refences](http://l20n.org/learn/entity-references)
- [Selectors](http://l20n.org/learn/selectors) and [Advanced selectors](http://l20n.org/learn/advanced-selectors)

dwaynebailey:
@ta2-1 I think you're overthinking this.  In our discussion by checker I literally meant variable checker, using the existing pofilter processing we already have.  Do that and we have varaible checking working.

Certainly the other things can be looked at but I suspect they're a different need.

nijel:
l20n support has been dropped, see https://github.com/translate/translate/pull/4047

Issue #3389, opened 2016-03-08
Add Fluent support to Translate Toolkit


phlax:
Quoting @dwaynebailey : 

> We have some WIP l20n support for l20n2po in ttk. We need to complete this and make sure it works for the minimal use case which is plurals and some other minor pieces.

Refs:

https://github.com/translate/translate/issues/2177

https://github.com/translate/translate/compare/master...dwaynebailey:l20n-db

unho:
More on the format: http://l20n.org/

nijel:
Maybe it could be based on https://github.com/projectfluent/python-fluent now...

PR #4047, opened 2020-06-11
l20n: Drop support for deprecated format

nijel:
It was never completed and was deprecated by Mozzilla in favor of Fluent.

See #4018


nijel:
In case we really agree to drop l20n, related issues and PRs should be closed as well: #3529, #3471, #3389

PR #4049, opened 2020-06-11
Remove u prefix from unicode literals

nijel:
It is not needed on Python 3.


unho:
@nijel So many changes and such a simple change. I won't even check the changes, lgtm.

PR #4053, opened 2020-06-12
Sort all imports

nijel:
Updated isort configuration and sorted imports.

### DIFF
--- a/translate/storage/statsdb.py
+++ b/translate/storage/statsdb.py
@@ -27,9 +27,9 @@ import re
 import stat
 import sys
 from collections import UserDict
-import _thread
 from sqlite3 import dbapi2
 
+import _thread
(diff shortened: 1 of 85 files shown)


translate/storage/statsdb.py
```diff
@@ -27,9 +27,9 @@
 import stat
 import sys
 from collections import UserDict
-import _thread
 from sqlite3 import dbapi2
 
+import _thread
```
unho, line 32: @nijel Is this not from the standard library?

nijel: Not sure, will look into this

nijel: I've rewritten the code to avoid using _thread, see https://github.com/translate/translate/pull/4057

PR #4054, opened 2020-06-12
convert: Fix conversion to multiple formats

nijel:
The dictionaries had duplicate keys, which are silently ignored. Changed
optrecurse to accept iterable and store the formats in tuple instead.

### DIFF
--- a/translate/convert/oo2xliff.py
+++ b/translate/convert/oo2xliff.py
@@ -157,12 +157,12 @@ def convertoo(inputfile, outputfile, templates, pot=False, sourcelanguage=None, 
 
 def main(argv=None):
     from translate.convert import convert
-    formats = {
-        "oo": ("xlf", convertoo),
-        "sdf": ("xlf", convertoo),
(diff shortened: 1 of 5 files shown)


nijel:
Originally these were dicts, where duplicates are silently discared:

```
>>> a = {"a": 1, "a": 2}
>>> a
{'a': 2}
```

So half of the formats were not applied as they had same keys.

translate/convert/oo2xliff.py
```diff
@@ -157,12 +157,12 @@ def convertoo(inputfile, outputfile, templates, pot=False, sourcelanguage=None,
 
 def main(argv=None):
     from translate.convert import convert
-    formats = {
-        "oo": ("xlf", convertoo),
-        "sdf": ("xlf", convertoo),
-        "oo": ("xliff", convertoo),
-        "sdf": ("xliff", convertoo),
```
unho, line 164: Ok, here are the duplicate keys.

PR #4056, opened 2020-06-12
Fix undefined variables

nijel:
~~Work in progress, there are more things to fix.~~


nijel:
The errors were found by flake8, most of the code is not covered by tests...

unho:
lgtm

PR #4066, opened 2020-06-16
tools: Fix execution with Python 3

### LINKED ISSUE #4065: get_moz_enUS.py seems py2 only
Based on https://github.com/translate/translate/blob/master/tools/mozilla/get_moz_enUS.py#L22 this code should not work under python3.

It should be probably ported or killed as now with 3.0.0 it should not work as it is py3 only release.

nijel:
Fixes #4065

I have no clue if these tools are still useful. Maybe it's better to drop tools/mozilla altogether?

### DIFF
--- a/tools/mozilla/get_moz_enUS.py
+++ b/tools/mozilla/get_moz_enUS.py
@@ -19,7 +19,7 @@
 
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


unho:
lgtm

PR #4068, opened 2020-06-24
Support XWiki Formats

surli:
* Informations about XWiki Translation Format are available there: https://dev.xwiki.org/xwiki/bin/view/Community/XWiki%20Translations%20Formats/
  * Add new xwiki dialect to handle special escape of simple quotes
  * Improve propunit to allow handling missing translations serialized
    in comments
  * Add xwiki unit that uses xwiki dialect and a specific comment prefix
    for missing translations
  * Add xwiki file that uses xwiki units

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -134,6 +134,9 @@ from translate.lang import data
 from translate.misc import quote
 from translate.misc.multistring import multistring
 from translate.storage import base
+from copy import deepcopy
+from xml.etree import ElementTree
+from xml.sax.saxutils import escape, unescape
 
 
 labelsuffixes = (".label", ".title")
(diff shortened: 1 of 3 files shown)


nijel:
Can you please add test for parse and serialize in XWikiFullPage? That seems to be not covered at all now (probably the tests are in https://github.com/WeblateOrg/weblate/pull/4093 and should be moved here).

surli:
> Can you please add test for parse and serialize in XWikiFullPage? That seems to be not covered at all now (probably the tests are in WeblateOrg/weblate#4093 and should be moved here).

Yep that was my plan, still working on it :) I switched it to draft until it's ready for review again.

surli:
@nijel so this is now ready for review.

translate/storage/properties.py
```diff
@@ -1067,3 +1183,124 @@ class joomlafile(propfile):
...
+            content = ""\
+                .join(self.root.find("content").itertext())
+            content = unescape(content).encode(self.encoding)
+            super(XWikiPageProperties, self).parse(content)
+
+    def set_xwiki_xml_attributes(self, newroot):
+        for e in newroot.findall("object"):
+            newroot.remove(e)
+        for e in newroot.findall("attachment"):
+            newroot.remove(e)
```
nijel, line 1246: These removals are not covered by tests, are they needed?

translate/storage/properties.py
```diff
@@ -1067,3 +1183,124 @@ class joomlafile(propfile):
...
+        kwargs['personality'] = "xwiki"
+        kwargs['encoding'] = "utf-8"
+        super(xwikifile, self).__init__(*args, **kwargs)
+        self.root = None
+
+    def parse(self, propsrc):
+        if propsrc != b"\n":
+            self.root = ElementTree.XML(propsrc)
+            content = ""\
+                .join(self.root.find("content").itertext())
```
nijel, line 1238:
```suggestion
            content = "".join(self.root.find("content").itertext())
```

translate/storage/properties.py
```diff
@@ -1067,3 +1183,124 @@ class joomlafile(propfile):
...
+    Name = "XWiki Full Page"
+
+    def parse(self, propsrc):
+        if propsrc != b"\n":
+            self.root = ElementTree.XML(propsrc)
+            content = ""\
+                .join(self.root.find("content").itertext())\
+                .replace("\n", "\\n")
+            title = ""\
+                .join(self.root.find("title").itertext())
```
nijel, line 1285:
```suggestion
            title = "".join(self.root.find("title").itertext())
```

translate/storage/properties.py
```diff
@@ -1067,3 +1183,124 @@ class joomlafile(propfile):
...
+            forparsing = ""
+            if content != "":
+                forparsing += "content={}\n"\
+                    .format(unescape(content))
```
nijel, line 1289:
```suggestion
                forparsing += "content={}\n".format(unescape(content))
```

translate/storage/properties.py
```diff
@@ -1067,3 +1183,124 @@ class joomlafile(propfile):
...
+            if title != "":
+                forparsing += "title={}\n" \
+                    .format(unescape(title))
```
nijel, line 1292:
```suggestion
                forparsing += "title={}\n".format(unescape(title))
```

translate/storage/properties.py
surli: They are. Adding tests to cover it.

surli: It's now covered.

nijel:

surli:
Thanks! What release process does translate follows? It's releasing at the same time than Weblate?

nijel:
For last few releases it was released around Weblate release because I am the one doing the release here as well, but there is no rule for that.

surli:
OK, I'm asking since I need those changes for my Weblate PR, so not sure how to pursue until this is released.

nijel:
You don't need a release for that. Weblate is also tested against master branch of translate-toolkit and the integration should be written in a way to disable file format in case it is not available in the translate-toolkit (all you need to do for that is to use string tuple as `loader` instead of importing the class directly).

surli:
> (all you need to do for that is to use string tuple as loader instead of importing the class directly).

Yes already did that changes :) 

> Weblate is also tested against master branch of translate-toolkit 

Nice, I wasn't aware of that. Thanks for the info, I'll take back my work on it then.

nijel:
> Nice, I wasn't aware of that. Thanks for the info, I'll take back my work on it then.

Historically we've seen breakages after some changes in the translate-toolkit, so that's why it was introduced. It's also useful to be able to test support for new formats before it's released in the translate-toolkit.

PR #4069, opened 2020-06-25
storage: Remove version control integration

nijel:
This is supposedly dead code (see #4018).


docs/developers/styleguide.rst
```diff
@@ -210,7 +210,6 @@ Like in :pep:`8`, but:
     from lxml.html import fromstring
 
     from translate.filters import checks
-    from translate.storage import versioncontrol
```
unho, line 213: @nijel This was actually an example on the styleguide. So please replace this with something that fits the example, like `from translate.storage import base`

nijel: Done in https://github.com/translate/translate/pull/4138

PR #4070, opened 2020-06-29
Never use empty translation in po2json, fix #3652

### LINKED ISSUE #3652: po2json puts empty string in resulting JSON if untranslated
Currently `po2json` is populating values with an empty `msgstr` with `""`, instead of falling back to the value of `msgid`. Is this the expecpted behaviour? If so, what do you think about adding a flag such as `--defaultuntranslated`. It could work similar to `--removeuntranslated` except it would instead take the value from `msgid` and put that in the resulting JSON.

Here are the steps to reproduce:

```bash
$ json2po --version
json2po 2.2.3

$ po2json --version
po2json 2.2.3
```

Put the following in `data.json`:
```json
{
	"one": "1 / One",
	"two": "2 / Two",
	"three": "3 / Three",
	"four": "4 / Four"
}
```

```
$ json2po data.json data.pot
$ cp data.pot data.fr.po
```

Then edit the `data.fr.po` file, leaving some translations blank, e.g.:

```
#: .one
msgid "1 / One"
msgstr "1 / Un"

#: .two
msgid "2 / Two"
msgstr "2 / Deux"

#: .three
msgid "3 / Three"
msgstr ""

#: .four
msgid "4 / Four"
msgstr "4 / Quatre"
```

Generating the translated `.json` file results in the following. Note the blank string against `"three"` instead of falling back to the value in `msgid` of `"3 / Three"`:

```bash
$ po2json -t data.json data.fr.po data.fr.json
```

```json
{
    "one": "1 / Un",
    "two": "2 / Deux",
    "three": "",
    "four": "4 / Quatre"
}
```

wojtek555:


nijel:
Thanks, looks good. Would you be able to add some tests to cover this behavior?

wojtek555:
OK, I'll add some tests.

nijel:

wojtek555:
You're welcome.

PR #4074, opened 2020-07-03
fix: line return in XWiki Page Properties format.

surli:
Small fix for the new XWiki Page Properties format.


surli:
@nijel This is a small fix needed for my PR on Weblate.

PR #4076, opened 2020-07-08
Android: Fix escaping double space with HTML

nijel:
Using quotes does not work well with HTML markup, so another escaping is
needed. Using unicode quote seems to be safest in this case.


unho:
lgtm

nijel:
Thanks for review.

PR #4084, opened 2020-07-28
fix: double escaping in XWiki Format.

surli:


surli:
@nijel this is one of the fix I need for Weblate

nijel:

PR #4086, opened 2020-07-28
fix: preserve blank line in properties

surli:


surli:
@nijel this one is ready too

nijel:

PR #4087, opened 2020-07-28
fix: Final fixes for XWiki Format

surli:
- Fix extracting language for XWiki Full Page format
  - Use language or default language
  - Ensure to output properly the unit when serializing by using source
    or translation
  - Ensure to serialize the locale in root node.

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -1229,8 +1229,10 @@ class XWikiPageProperties(xwikifile):
         kwargs['personality'] = "xwiki"
         kwargs['encoding'] = "utf-8"
         self.root = None
-        self.parser = etree.XMLParser(strip_cdata=False, resolve_entities=False)
         super(xwikifile, self).__init__(*args, **kwargs)
+
+    def get_parser(self):
+        return etree.XMLParser(strip_cdata=False, resolve_entities=False)
 
     def extract_language(self):
         language_node = self.root.find("language")
@@ -1243,7 +1245,7 @@ class XWikiPageProperties(xwikifile):
 
     def parse(self, propsrc):
         if propsrc != b"\n":
-            self.root = etree.XML(propsrc, self.parser)
+            self.root = etree.XML(propsrc, self.get_parser())
             content = "".join(self.root.find("content").itertext())
             content = unescape(content).encode(self.encoding)
             self.extract_language()
@@ -1273,7 +1275,7 @@ class XWikiPageProperties(xwikifile):
 
     def serialize(self, out):
         if self.root is None:
-            self.root = etree.XML(self.XWIKI_BASIC_XML, self.parser)
+            self.root = etree.XML(self.XWIKI_BASIC_XML, self.get_parser())
         newroot = deepcopy(self.root)
         # We add a line break to ensure to have a line break before
         # closing of content tag.
@@ -1295,7 +1297,7 @@ class XWikiFullPage(XWikiPageProperties):
 
     def parse(self, propsrc):
         if propsrc != b"\n":
-            self.root = etree.XML(propsrc, self.parser)
+            self.root = etree.XML(propsrc, self.get_parser())
             content = ""\
                 .join(self.root.find("content").itertext())\
                 .replace("\n", "\\n")
@@ -1317,7 +1319,7 @@ class XWikiFullPage(XWikiPageProperties):
         unit_title = self.findid("title")
         unit_content = self.findid("content")
         if self.root is None:
-            self.root = etree.XML(self.XWIKI_BASIC_XML, self.parser)
+            self.root = etree.XML(self.XWIKI_BASIC_XML, self.get_parser())
         newroot = deepcopy(self.root)
         if unit_title is not None:
             newroot.find("title").text = self.output_unit(unit_title)
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -135,7 +135,7 @@ from translate.misc import quote
 from translate.misc.multistring import multistring
 from translate.storage import base
 from copy import deepcopy
-from lxml import etree as ElementTree
+from lxml import etree
 from xml.sax.saxutils import unescape
 
 
@@ -1229,6 +1229,7 @@ class XWikiPageProperties(xwikifile):
         kwargs['personality'] = "xwiki"
         kwargs['encoding'] = "utf-8"
         self.root = None
+        self.parser = etree.XMLParser(strip_cdata=False, resolve_entities=False)
         super(xwikifile, self).__init__(*args, **kwargs)
 
     def extract_language(self):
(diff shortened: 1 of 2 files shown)


surli:
@nijel this one is ready too.

translate/storage/test_properties.py
```diff
@@ -1104,7 +1109,13 @@ def test_cleaning_attributes(self):
         generatedcontent = BytesIO()
         propfile.settargetlanguage("fr")
         propfile.serialize(generatedcontent)
-        expected_xml = properties.XWikiPageProperties.XML_HEADER + """<xwikidoc reference="XWiki.AdminTranslations">
+
+        ## ElementTree changed the way to process the attribute between Python 3.7 and 3.8
+        if sys.version_info.minor >= 8:
+            xml_open = "<xwikidoc reference=\"XWiki.AdminTranslations\" locale=\"fr\">"
+        else:
+            xml_open = "<xwikidoc locale=\"fr\" reference=\"XWiki.AdminTranslations\">"
```
nijel, line 1117: Maybe it's better to compare XML instead of guessing ordering of the attributes? Maybe something like https://stackoverflow.com/a/7060342/225718 would be useful here.

surli: I hesitated about that, could be useful if I had more tests with various XML attributes but here there's only those two (I managed to remove the other ones from the test to avoid the problem). I'm in favor of keeping it like that since it's working properly on different platforms / versions of Python.

nijel: Looking again at this I might realized why other formats do not need this - everything else uses lxml. One of the reasons is that it does better job in preventing some security issues, see https://docs.python.org/3/library/xml.html#xml-vulnerabilities.

Can you please change the XWiki parser to use it as well? It should also disable entities resolving to avoid vulnerability in that. This can be done by using `etree.XMLParser(resolve_entities=False)`. My guess is that this condition will no longer be necessary with lxml.

surli: > Can you please change the XWiki parser to use it as well?

ok I will take a look to use lxml. Hope the switch will be easy :)

nijel:
Looks good, the only thing I'd like to see is to use XML parser which avoids resolving XML entities, for example see:

https://github.com/translate/translate/blob/de82eb28c468a3a7113582ea1b677203a95821e2/translate/storage/flatxml.py#L177

surli:
@nijel so I perform the fix you requested but now I get the following error when I execute Weblate tests: 
> TypeError: cannot pickle 'lxml.etree.XMLParser' object

Not sure what I'm missing there.

nijel:
@surli I have no clue why it should be pickling it, probably something wrong is going on. Can you post here full traceback?

surli:
> Can you post here full traceback?

Here it is: 
```
  File "/home/urli/Github/weblate/weblate/formats/tests/test_formats.py", line 928, in test_translate_file
    self.translate_unit(
  File "/home/urli/Github/weblate/weblate/formats/tests/test_formats.py", line 901, in translate_unit
    unit_to_translate, create = translation_data.find_unit(
  File "/home/urli/Github/weblate/weblate/formats/base.py", line 290, in find_unit
    return self._find_unit_template(context)
  File "/home/urli/Github/weblate/weblate/formats/base.py", line 266, in _find_unit_template
    ttkit_unit = deepcopy(template_ttkit_unit)
  File "/usr/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.8/copy.py", line 296, in _reconstruct
    value = deepcopy(value, memo)
  File "/usr/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/usr/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/usr/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.8/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle 'lxml.etree.XMLParser' object
```

surli:
Hmm actually maybe I shouldn't keep the parser on the object. Sounds better for perf reason, but maybe it's just a bad idea and I should create it only on demand.

nijel:
Yes, that seems to be the way around (and what other storages do).

surli:
@nijel I confirm I don't get this error anymore.

translate/storage/properties.py
```diff
@@ -1231,12 +1231,24 @@ def __init__(self, *args, **kwargs):
...
+
     def parse(self, propsrc):
         if propsrc != b"\n":
-            self.root = ElementTree.XML(propsrc)
+            self.root = etree.XML(propsrc, self.get_parser())
             content = "".join(self.root.find("content").itertext())
             content = unescape(content).encode(self.encoding)
-            self.setsourcelanguage(self.root.find("language").text)
+            self.extract_language()
             super(XWikiPageProperties, self).parse(content)
```
nijel, line 1252: Can be simplified:

```suggestion
            super().parse(content)
```

translate/storage/properties.py
```diff
@@ -135,8 +135,8 @@
 from translate.misc.multistring import multistring
 from translate.storage import base
 from copy import deepcopy
-from xml.etree import ElementTree
-from xml.sax.saxutils import escape, unescape
+from lxml import etree
+from xml.sax.saxutils import unescape
```
nijel, line 139: Is the unescape really needed? It is used in several places, but I see single test for XML entity.

surli: Good catch, it's actually a mistake the way I use it here. Improving the test and fixing the code.

nijel:

Issue #3857, opened 2019-01-14
JSON Serialization breaks the structure


jakub-szczepaniak:
I have noticed that in the last release, JSON serialization is broken for more complex structure.
See below the output from the pytest diff.
```
        def test_serialize_more_complex(self):
            store = self.StoreClass()
            more_complex = b"""
    {
    "foo": {
        "foo_2": {
            "foo_3": "bar",
            "foo_4": "bar2"
        },
        "fizz" : "buzz"
        }
    }
    """
            store.parse(more_complex)
            out = BytesIO()
            store.serialize(out)

>           assert out.getvalue() == more_complex
E           assert b'{\n    "foo...: "buzz"\n}\n' == b'\n{\n"foo": ..."\n    }\n}\n'
E             At index 0 diff: 123 != 10
E             Right contains more items, first extra item: 105
E             Full diff:
E             - (b'{\n    "foo.foo_2.foo_3": "bar",\n    "foo.foo_2.foo_4": "bar2",\n    "foo.'
E             -  b'fizz": "buzz"\n}\n')
E             + (b'\n{\n"foo": {\n    "foo_2": {\n        "foo_3": "bar",\n        "foo_4": '
E             +  b'"bar2"\n    },\n    "fizz" : "buzz"\n    }\n}\n')
```

I have added this test to the pull request here
https://github.com/translate/translate/pull/3856

Basically, the json structure with several nested dictionaries is flattened to the single key "foo.foo_2.foo_3"

I would like to have this fixed, so if you any of you could point me where to start, It would be great.

Thanks, 
Jakub

nijel:
The problem is that neither of these handles all files correctly. Originally there was just `JsonFile` and `po2json` used that. Later I've added `JsonNestedFile`, but it was not integrated into `po2json`.

There really should be just one JSON format which would handle any keys just fine, see https://github.com/translate/translate/issues/3819 and https://github.com/translate/translate/pull/3835 (but I failed to find time to continue developing that for now).

To illustrate the problem, take following JSON structure, it won't be properly handled in neither of current JSON formats:

```json
{
  "key1": {
    "key2": "value1"
  },
  "key3": "value2",
  "key.key4": "value3"
}
```

PR #4090, opened 2020-08-04
Share serializer code for YAML and JSON

nijel:
The code is doing pretty much same thing and both implementations were covering different corner cases. This PR merges both implementations and uses single code for both formats.

### DIFF
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -887,3 +887,34 @@ class TranslationStore:
         :rtype: string
         """
         return "id"
+
+
+class DictUnit(TranslationUnit):
+    def getvalue(self):
+        """Returns dictionary for serialization."""
+        raise NotImplementedError()
+
+
+class DictStore(TranslationStore):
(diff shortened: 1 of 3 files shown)


mkjaxxo:
Does these changes also fix the conversion from JSON array
 
` "dayNames": ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"],`

to
 ```
 "dayNames[0]": "Dimanche",
    "dayNames[1]": "Lundi",
    "dayNames[2]": "Mardi",
    "dayNames[3]": "Mercredi",
    "dayNames[4]": "Jeudi",
    "dayNames[5]": "Vendredi",
    "dayNames[6]": "Samedi",

```
? Thanks in advance.

nijel:
@mkjaxxo This would work with JsonNestedFile right now. Problematic mixed cases (see #3857) are addressed by https://github.com/translate/translate/pull/4091 which is built on top of this.

PR #4093, opened 2020-08-10
Ensure external tools ran successfully in tests

llimeht:
Calls to external tools such as `msgfmt` and `pocompile` need to be checked to ensure that they succeeded. If they were to fail for any reason, then the test will fail at a later line, but for the wrong reason.


unho:
lgtm @nijel can you please review and merge if everything is ok?

PR #4131, opened 2020-09-23
lint: Introduce black for code formatting

nijel:
This introduces usage of [black](https://github.com/psf/black)

* Black is widely used and seems to be the preferred Python code formatter these days
* It enforces consistent code formatting
* The diff is *huge*, but automatically generated
* It will make git blame a bit more tricky (see https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame)
* It will probably make any pending pull requests having merge conflicts, please let me know if something deserves to merge prior this (note that all of them are at least three years old, so they are likely to have merge conflicts anyway right now) 

I'd like to get some feedback prior merging this :-)

See also https://github.com/translate/translate/pull/3989 for related discussion with @nschonni


translate/convert/json2po.py
```diff
@@ -91,24 +92,33 @@ def convert_unit(self, input_unit, commenttype):
...
+        output_store = convertor.convert_store(
+            input_store, duplicatestyle=duplicatestyle
+        )
     else:
         template_store = jsonl10n.JsonFile(template_file)
-        output_store = convertor.merge_store(template_store, input_store,
-                                             blankmsgstr=pot,
-                                             duplicatestyle=duplicatestyle)
+        output_store = convertor.merge_store(
+            template_store, input_store, blankmsgstr=pot, duplicatestyle=duplicatestyle
```
phlax, line 120: nit: i would put 1 per line - at least for `kwargs`

i much prefer the lesser indent tho!

phlax:
@nijel first skim through looks like great changes!

it *is* pretty huge - it will take some time if you want a thorough review - but if you dont mind driveby nits as i have time im happy to review further

nijel:
@phlax Sorry for not being clear enough - I don't want to review the code changes (these are generated and not configurable). I merely want to know opinion on using code formatter and whether black sounds reasonable choice.

translate/convert/json2po.py
nijel: Black puts that into one line if it fits and unless there is a trailing comma.

phlax:
> It enforces consistent code formatting

im a bit opinoniated that i dont like my formatters/linters too opinionated

that said, whatever makes the project easiest to maintain/contribute to is a definite improvement

phlax:
> opinion on using code formatter and whether black sounds reasonable choice.

ah k - well i think i added that as you were responding!

translate/convert/json2po.py
phlax: > these are generated and not configurable

i guess thats a nit for the linter in that case - in terms of readability i think its better to see the args/kwargs per-line (mostly at least - but def where they are mixed)

i havent had a chance to look at black as i mostly have avoided auto-linting like that, but it would be great if it were configurable in this kind of way

phlax (approved):
lgtm if you are happy with it - with above caveats

nijel:
> im a bit opinioniated that i dont like my formatters/linters too opinionated

I used to have similar opinion, but I've changed my mind over the years. IMHO it really helps if the coding standard is well-defined without leaving much space for choices. Otherwise, people end up preferring different choices and the code is less consistent. This is quite well described in https://www.kevinpeters.net/auto-formatters-for-python

tools/mozilla/buildxpi.py
```diff
@@ -50,7 +50,6 @@
 from subprocess import PIPE, CalledProcessError, Popen
 from tempfile import mkdtemp
 
-
```
unho, line 53: Please keep 2 blank lines after imports.

tools/mozilla/buildxpi.py
```diff
@@ -66,21 +65,20 @@ def __init__(self, message=None, **kwargs):
 
     def __str__(self):
         """Format exception message string (avoiding TypeErrors)"""
-        output = ''
+        output = ""
```
unho, line 68: IMHO it would be easier to review if this specific change of quoting happens in a single commit, with no other changes mixed.

unho (changes requested):

tools/mozilla/buildxpi.py
nijel: There are no configurations in black.

tools/mozilla/buildxpi.py
nijel: Okay, I've split this into three commits - adding black, applying black without string normalization and string normalization.

tools/mozilla/buildxpi.py
nijel: ...but this was rather by non inteded changes in isort configuration. Fixed.

unho (commented):
@nijel I am ok with the commit changing string quoting. Regarding the other changes I didn't review all of them, but I have already found some that look really weird to me.

To be honest I dislike the idea of having a tool like black unless it can be configured, but my involvement on this project is now not as strong as it was so I don't want to impose my way in any way. In other words, if you think it is ok I won't oppose, but won't approve either.

setup.py
```diff
@@ -238,7 +255,7 @@ def getcompilecommand(self):
         def chop(self, pathname):
             """returns the path relative to self.dist_dir"""
             assert pathname.startswith(self.dist_dir)
-            return pathname[len(self.dist_dir):]
+            return pathname[len(self.dist_dir) :]
```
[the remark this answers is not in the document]
nijel: Why it would not be legal? For example, PEP-8 says "in a slice the colon acts like a binary operator", so if there is space used around binary operators in other cases, it should be here as well.

tools/mozilla/buildxpi.py
```diff
@@ -217,51 +234,58 @@ def build_xpi(l10nbase, srcdir, outputdir, langs, product, delete_dest=False,
...
 
     p.add_argument(
-        '-L', '--l10n-base',
+        '-L',
+        '--l10n-base',
         type=str,
         dest='l10nbase',
         default='l10n',
-        help='The directory containing the <lang> subdirectory.'
+        help='The directory containing the <lang> subdirectory.',
```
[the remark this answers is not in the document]
nijel: No, it isn't: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

nijel:
I don't want to push something that most of the existing contributors don't like, that's why I've asked for reviews here. Good reasons for using code formatter are described for example in https://www.kevinpeters.net/auto-formatters-for-python followed by description of most popular choices. For me, the preferred choice is black because it works well for many projects where I've contributed and I've recently started to use it for Weblate as well. On the other side formatting of long args would do all of them similarly (`InnoScript.__init__` what @unho commented).

tools/mozilla/buildxpi.py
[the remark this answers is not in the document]
unho: Thanks for the link. I've learnt something new and I won't oppose PEP8 on this particular case.

setup.py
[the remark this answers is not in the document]
unho: I never saw whitespace around colons in slices, but you are right that PEP8 allows it. Nevertheless I dislike it.

julen (approved):
Looks good to me and thanks for doing this 👍

nijel:
Thanks for the feedback, I've just merged this. There is still place to improve the code formatting further, but I think this is good step to more consistent code base and providing better guidance to contributors on that.

PR #4174, opened 2020-11-22
Single-file output option for html2po

anderskaplan:
Sometimes it can be more convenient to work with a single pot/po file for an entire web site, or part of a web site, instead of one per html file. This PR adds that option to html2po.

The function is invoked through the "multifile" command-line option, an option which is provided by some of the other converters for similar purposes.

The PR also removes an obsolete and non-functional option to include untagged content from html2po.

### DIFF
--- a/translate/misc/optrecurse.py
+++ b/translate/misc/optrecurse.py
@@ -98,7 +98,8 @@ class ManHelpFormatter(optparse.HelpFormatter):
 
 
 class StdoutWrapper:
-    out = sys.stdout
+    def __init__(self):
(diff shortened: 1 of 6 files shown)


translate/misc/optrecurse.py
```diff
@@ -98,7 +98,8 @@ def format_option_strings(self, option):
 
 
 class StdoutWrapper:
-    out = sys.stdout
```
anderskaplan, line 101: This is to be able to capture the output from a unit test. Without the constructor, the assignment is done too early and no output is captured.

PR #4179, opened 2020-11-24
Html2po: option for single file output

anderskaplan:
Sometimes it can be more convenient to work with a single pot/po file for an entire web site, or part of a web site, instead of one per html file. For example, this is a workflow familiar to authors of WordPress plugins. This PR adds that option to html2po.

The function is invoked through the "multifile" command-line option, an option which is provided by some of the other converters for similar purposes. It might be better to create a new option for it. Please let me know what you think.

The PR also removes an obsolete and non-functional option to include untagged content from html2po.

### DIFF
--- a/translate/misc/optrecurse.py
+++ b/translate/misc/optrecurse.py
@@ -102,7 +102,8 @@ class ManHelpFormatter(optparse.HelpFormatter):
 
 
 class StdoutWrapper:
-    out = sys.stdout
+    def __init__(self):
+        self.out = sys.stdout
 
     def __getattr__(self, name):
         return getattr(self.out, name)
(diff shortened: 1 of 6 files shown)


translate/misc/optrecurse.py
```diff
@@ -102,7 +102,8 @@ def format_option_strings(self, option):
 
 
 class StdoutWrapper:
-    out = sys.stdout
```
anderskaplan, line 105: This is to be able to capture the output from a unit test. Without the constructor, the assignment is done too early and no output is captured.

nijel:

alexsegura:
@nijel would it be possible to make a new release to have this option, please? 🙏

nijel:
Yes, I plan to do it soon, but first I'd like to merge some more PRs like https://github.com/translate/translate/pull/4188 or https://github.com/translate/translate/pull/4206

PR #4182, opened 2020-11-27
[html2po] undo unintended file mode change in #4179

anderskaplan:
When checking out master after #4179 had been merged, I noticed that html2po.py had lost the executable bit by accident in that PR. This PR undoes that change. Sorry for the fuss.


nijel:

PR #4184, opened 2020-12-01
xliff: set xliff language on all file nodes

nijel:
Setting it only on first one will produce mixed language xliff,
what is usually not desired.

Fixes https://github.com/WeblateOrg/weblate/issues/4945


bryceco:
Thanks!

Issue #4177, opened 2020-11-23
huge temporary file


Jibec:
translate toolkit uses 21,6GB of space on my drive:
```
[jb@localhost .translate_toolkit]$ ls -lah
total 21G
drwxrwxr-x.  2 jb jb 4,0K 23 nov.  19:00 .
drwx--x---+ 83 jb jb  12K 23 nov.  12:37 ..
-rw-r--r--.  1 jb jb  21G 23 nov.  19:00 stats.db
```

is this temporary file really needed in home folder?
maybe storing it in tmp would make it faster and more efficient?

as an example, this also create issues when using translate toolkit in automation: https://pagure.io/fedora-docs/translations-scripts/pull-request/18# 
because of using pocount there: https://pagure.io/fedora-docs/translations-scripts/blob/master/f/stats/compute.sh#_11

nijel:
The pocount scripts you are using is relying on [statsdb](https://github.com/translate/translate/blob/master/translate/storage/statsdb.py) which stores this sqlite database in `~/.translate_toolkit/stats.db`. It seems to cache stats for all files ever counted, but that's what I could figure out from quick look at the code and I've seen it for the first time now.

nijel:
Fixed by removing statsdb in  #4188.

Issue #3369, opened 2015-11-25
[RFC] Support XLIFF2.0


rvanlaak:
Tried to wrap my head around the changes that XLIFF2.0 has to offer, but can't really summarize it in actionable steps:

https://www.localizationworld.com/lwdub2014/feisgiltt/slides/Yves_Differences.pdf

The most important topics:

* 1.2 has many features that aren't included in 2.0 by default anymore
* 2.0 is a BC break on 1.2
* 2.0 allows to add specialized modules, but also for constraints and processing requirements

So is there a plan for translate toolkit to support the 2.0 version?

phlax:
Hi @rvanlaak this is something we have just begun looking at - although we are really only at the research and dev stage atm.

You can see my notes and links to progress so far here - https://github.com/phlax/pootle_liff/wiki

As this will take quite a bit of development to implement its likely to take a few months - but it is one of our development priorities in the new year.

rvanlaak:
@phlax you have some updates on this?

Symfony started supporting 2.0 completely lately: http://symfony.com/blog/new-in-symfony-2-8-translator-improvements

phlax:
hi @rvanlaak i made a little progress with this last month - https://github.com/translate/pyliff - but in terms of support in Translate Tool Kit we are still some way off - and dealing with some other priorities.

I can ping you when we make progress with this.

rvanlaak:
A ping now and then over here would be nice, but I think eventually weblate will benefit about it by integrating it. So maybe @nijel can say something about whether weblate and translation-toolkit priorities are in line :+1:

btaens:
Any progress on this feature?

zonky2:
another five years later and still no sign of life? :(

phlax:
as im no longer working on translate code im gonna remove my assignment

PR #4188, opened 2020-12-02
remove statsdb

Jibec:
I'm unsure to understand the use case for statsdb file, but I'm sure it creates issues I faced: https://github.com/translate/translate/issues/4177

I assume this feature may have been used by external tools who didn't want to implement a proper cache?
I did only test the unit test still worked, and don't really know how to test it from the command line without destroying my system (any idea welcome)


Jibec:
how do I make `make test-functional` use the new scripts instead of the ones installed by my package manager?

same question for make test

> (venv) [jb@localhost translate]$ lang="en" make test
> ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
> py.test: error: unrecognized arguments: --boxed
>   inifile: /home/jb/Projects/translate/pytest.ini
>   rootdir: /home/jb/Projects/translate
> 
> make: *** [Makefile:84 : test] Erreur 4

nijel:
Create a Python virtualenv and install current code to it using `pip install -e .` (the `.` represents path to your translate-toolkit checkout).

nijel:
Going back to topic of this pull request. I'm not sure that statsdb brings reasonable value to pocount besides sharing implementation. The caching itself is not that beneficient here and there is definitely issue with it as it only grows and old entries are never pruned. So removing statsdb from the pocount script seems reasonable to me.

On the other side, removing the module completely will break third party users.

Virtall seems to use statsdb, but on the other side current translate-toolkit releases are not compatible with it anyway (dropped Python 2 support and some legacy interfaces), so that one is probably not a blocker. If virtaal would still need it, it can integrate it to it's code base. Anyway mentioning @friedelwolff for this.

Pootle does not depend on statsdb itself, but uses wordcount function from the module.

Jibec:
thanks, make test-functional was easy, but I can't run "make test", it tells me:

```
(venv) [jb@localhost translate]$ make test
ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --boxed
  inifile: /home/jb/Projects/translate/pytest.ini
  rootdir: /home/jb/Projects/translate

make: *** [Makefile:84 : test] Erreur 4
```

I'm using pytest 6.1.2 with python 3.9

Jibec:
thanks, make test-functional was easy, but I can't run "make test", it tells me:

```
(venv) [jb@localhost translate]$ make test
ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --boxed
  inifile: /home/jb/Projects/translate/pytest.ini
  rootdir: /home/jb/Projects/translate

make: *** [Makefile:84 : test] Erreur 4
```

I'm using pytest 6.1.2 with python 3.9

I also see codecov says this test isn't covered: "if isinstance(unit.source, multistring):"
but what is a multistring? I read [this](https://github.com/translate/translate/blob/bca1340b72bf212b2da6f6759e23fc98ae941927/translate/misc/multistring.py), but have trouble to understand how it may looks like in a po file.

nijel:
`mutlisting` is used for storing plurals. You can use `pytest` directory for running the testsuite. The `Makefile` seems outdated and does not work with current pytest.

Jibec:
Thank you Michal.

I understand there is no reason to have the "test_plurals" function 
empty.
I'll try to cover that tonight, I'll let you know if I'm stuck somehow.

tests/cli/data/test_pocount_po_file/stdout.txt
```diff
@@ -5,5 +5,3 @@ Fuzzy:            0 (  0%)          0 (  0%)             n/a
 Untranslated:     0 (  0%)          0 (  0%)             n/a
 Total:            1                 3                      3
 
-Unreviewed:       1 (100%)          3 (100%)               3
```
nijel, line 8: Can this be added back? I don't feel like we should remove functionality here.

Jibec:
Le 2020-12-10 14:33, Michal Čihař a écrit :
> @nijel commented on this pull request.
> -------------------------
> 
> In tests/cli/data/test_pocount_po_file/stdout.txt [1]:
> 
>> @@ -5,5 +5,3 @@ Fuzzy:            0 (  0%)          0 (  0%)
>     n/a
>  Untranslated:     0 (  0%)          0 (  0%)             n/a
>  Total:            1                 3                      3
> 
> -Unreviewed:       1 (100%)          3 (100%)               3
> 
> Can this be added back? I don't feel like we should remove
> functionality here.

I feel like it's a Virtaal feature, not a pofile feature:
https://github.com/translate/virtaal/blob/6371d5bf9bb75fb79077c94d013972789ff46697/virtaal/controllers/unitcontroller.py#L80
https://github.com/translate/virtaal/blob/6371d5bf9bb75fb79077c94d013972789ff46697/virtaal/views/propertiesview.py#L37

Should I just say Unreviewed = Translated?
But what about other statuses?

Shouldn't we just remove these statuses from pocount output?

nijel:
Even if the tool is calle PO count, it can IMHO count strings in any file format. And some do support storing review state.

Jibec:
I'll try to find out how it looks like in an actual po file if I can find a real world example.
--
Jean-Baptiste

Jibec:
I found this file:
[example.txt](https://github.com/translate/translate/files/5675839/example.txt)

It provides this output:
```
[jb@localhost packages]$ pocount bluefish/po/el.po
Processing file : bluefish/po/el.po
Type               Strings      Words (source)    Words (translation)
Translated:      58 (  6%)         86 (  2%)              92
Fuzzy:          456 ( 52%)       1138 ( 33%)             n/a
Untranslated:   353 ( 40%)       2181 ( 64%)             n/a
Total:          867              3405                     92

Empty:          353 ( 40%)       2181 ( 64%)               0
Needs-Work:     456 ( 52%)       1138 ( 33%)               0
Unreviewed:      58 (  6%)         86 (  2%)              92
```

which leads to: https://github.com/translate/translate/blob/master/translate/tools/pocount.py#L276
does it sounds the right place to you?

I am unsure it is a feature, as I can't find any po file with "statuses" that may looks like those "extended" statuses.

nijel:
Seems like Unreviewed = Translated for gettext.

Jibec:
Because of the name of the command, and the test only using po files, I 
forgot it supports more than po files: 
http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/pocount.html

I think I should first add in a new PR:
* a test with xliff statuses
* a test with fuzzy strings with po (like the example I gave in previous 
comment)
* a test with plurals

Do you feel like additional scenarios are required to be added to tests?

nijel:
Covering these would be great addition, I don't thing more is needed right now.

Jibec:
I am almost done with this, but states still doesn't work when the attribute "approved=yes" is set.
Do you handle differently XLIFF 1.2 and 2.1 ? Because statuses are not the same (they simplified it quite a lot, which is the good approach)

nijel:
translate-toolkit doesn't support xliff 2, see https://github.com/translate/translate/issues/3369

translate/tools/pocount.py
```diff
@@ -116,12 +189,38 @@ def calcstats_old(filename):
...
+        sourcewords_count = state_stats.get("sourcewords", 0)
+        targetwords_count = state_stats.get("targetwords", 0)
+
+        unit_count += 1
+        sourcewords_count += wordcounts[id(unit)][0]
+        targetwords_count += wordcounts[id(unit)][1]
+
+        state_stats["units"] = unit_count
+        state_stats["sourcewords"] = sourcewords_count
+        state_stats["targetwords"] = targetwords_count
```
nijel, line 219: This all could be heavily simplified by using defaultdict (the patch would need importing it as well):

```suggestion
        state_stats = stats.get(extended_state_strings[state], defaultdict(int))

        state_stats["units"] += 1
        state_stats["sourcewords"] += wordcounts[id(unit)][0]
        state_stats["targetwords"] += wordcounts[id(unit)][1]
```

Jibec:
one commit at a time, I will survive this PR ;)

I understand that the counting have to behave depending on file type.
if file is a PO, what I implemented applies
but if file is an xliff, then it won't work

either I change my logic to have something generic
either I check the file type (which is the storage class used?)

nijel:
Did this work with the pocount before removing statsdb?

Jibec:
yes it did.

no worries, I'll fix that, i am just doing it slowly because not much time to allocate and not much dev skills.
--
Jean-Baptiste

Jibec:
ok, I give up on flake8 things, it's not understandable for me :/

nijel:
The changes in 0477084ccf94b4439990d937f2728d3c6ea1d2ba really should not be needed. That's existing code which does pass the lint...

Jibec:
> The changes in [0477084](https://github.com/translate/translate/commit/0477084ccf94b4439990d937f2728d3c6ea1d2ba) really should not be needed. That's existing code which does pass the lint...

is it harmful? I tried to respect pycharm recommendations, it took me quite some time to go through all requests from the many existing linters.

nijel:
It's kind of harmful because it introduces unrelated changes to this PR - for example all `ConsoleColor` related changes have nothing to do with statsdb removal.

The `handledir` -> `handle_dir` change (and similar ones) is probably not desired as the whole translate-toolkit uses this way of naming methods. Even though that might not be the best pratice, it's the practice current code base uses a lot. In case this should  be changed it should be discussed before and introduced separately - it would break external API as well.

Jibec:
I think I'm done with this PR, I decided to rebase it the multiple commits were not really useful to understand the change.

nijel:

PR #4202, opened 2020-12-11
test po fuzzy & plurals, and xlif states

Jibec:
in continuation of https://github.com/translate/translate/pull/4188#issuecomment-743056848

could you please tell me what are realistic combination in xliff files? what does approved means in a trans-unit?
the current output results doesn't seems correct to me, I'm a little bit confused.

### DIFF
--- a/tests/cli/data/test_pocount_po_fuzzy/one.po
+++ b/tests/cli/data/test_pocount_po_fuzzy/one.po
@@ -1,0 +1,6 @@
+msgid "Translated string"
+msgstr "Une chaîne traduite"
+
+#, fuzzy
+msgid "Fuzzy string"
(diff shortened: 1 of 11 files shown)


nijel:
The relation between [approved](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#approved) and [state](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#state) is not obvious to me. Weblate applies additional logic on top what translate-toolkit does based on the user feedback and it would be good to consolidate it, but I'm not really sure whether Weblate approach is better (it adds `state="final"` when string is approved in addition to `approved="yes"`).

Anyway the tests look good as they are now.

Issue #4183, opened 2020-12-01
po2tmx ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters


Jibec:
Hello,

here is how I use po2tmx: https://pagure.io/fedora-localization-statistics/blob/less_subprocess/f/build_tm.py#_121
here is the faulty file: https://jibecfed.fedorapeople.org/partage/ar.po.gz
here is the stacktrace:

```
Traceback (most recent call last):
  File "/home/jb/Projects/fedora-localization-statistics/./build_tm.py", line 201, in <module>
    main()
  File "/home/jb/Projects/fedora-localization-statistics/./build_tm.py", line 70, in main
    process_tmx(lang_code, compendium_file, tmx_file)
  File "/home/jb/Projects/fedora-localization-statistics/./build_tm.py", line 122, in process_tmx
    tmx_file_result = po2tmx.convertpo(
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/convert/po2tmx.py", line 70, in convertpo
    convertor.convertfiles(inputfile, outputfile.tmxfile, sourcelanguage,
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/convert/po2tmx.py", line 62, in convertfiles
    tmxfile.addtranslation(source, sourcelanguage, translation,
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/tmx.py", line 152, in addtranslation
    unit = self.addsourceunit(source)
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/lisa.py", line 297, in addsourceunit
    newunit = self.UnitClass(source)
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/lisa.py", line 59, in __init__
    super().__init__(source)
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/base.py", line 125, in __init__
    self.source = source
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/lisa.py", line 112, in source
    self.setsource(source, sourcelang='en')
  File "/home/jb/Projects/fedora-localization-statistics/venv/lib/python3.9/site-packages/translate/storage/lisa.py", line 117, in setsource
[...]

PR #4213, opened 2020-12-21
add control cleanup

Jibec:
fix: #4183
some pofiles to convert may contain unexpected content.
Instead of loosing it all, just clean the faulty string


translate/storage/tmx.py
```diff
@@ -64,7 +76,6 @@ def addnote(self, text, origin=None, position="append"):
     def _getnotelist(self, origin=None):
         """Returns the text from notes.
 
-        :param origin: Ignored
```
nijel, line 67: This seems like unintended change.

translate/storage/tmx.py
```diff
@@ -18,13 +18,19 @@
...
+
 from lxml import etree
 
 from translate import __version__
 from translate.misc.xml_helpers import setXMLlang
 from translate.storage import lisa
 
 
+def remove_control_characters(s):
+    return "".join(ch for ch in s if unicodedata.category(ch)[0] != "C")
```
nijel, line 31: This removes way too much for example right-to-left/left-to-right marks are C as well and you definitely do not want these to be removed.

The actual code in libxml rejects just a few ASCII control chars, see https://github.com/GNOME/libxml2/blob/46837d47d59c7b8c9bd1d08a6a717a90a7f1ceb6/include/libxml/chvalid.h#L108-L110

Also, I think this code is better placed to `translate/misc/xml_helpers.py` because it is probably useful for other XML based formats as well.

translate/storage/tmx.py
Jibec: yep, indeed

translate/storage/tmx.py
Jibec: ok, makes sense. I think I'll finish work and then open a new PR with everything at the right spot.
Is it ok like this?

nijel: Yes, sounds good.

PR #4214, opened 2020-12-22
xliff: Ensure xml:space="preserve" is set when updating target

nijel:
Otherwise the whitespace might be lost.

Fixes https://github.com/WeblateOrg/weblate/issues/5026


apourche:
Thanks @nijel  !

PR #4219, opened 2020-12-27
prevent invalid chars to break tmx conversion

Jibec:
fix: #4183
this is a rewrite of #4213


nijel:

PR #4224, opened 2021-01-01
Po2html: option for single PO file input

anderskaplan:
Sometimes it can be more convenient to work with a single pot/po file for an entire web site, or part of a web site, instead of one per html file. For example, this is a workflow familiar to authors of WordPress plugins. The option to extract translatable content from an entire web site into a single po file was added to html2po in #4179. This pull request adds the inverse operation to po2html.

The function is activated when a single PO file is given as input, and a directory of HTML files given as template. Instead of iterating over the input files, the converter will iterate over the template files and produce one output file per template file.

The pr also includes a little cleanup of the documentation: the entries on the Converters page are sorted alphabetically to make them easier to find.

### DIFF
--- a/docs/commands/html2po.rst
+++ b/docs/commands/html2po.rst
@@ -2,10 +2,10 @@
 .. _html2po:
 .. _po2html:
 
-html2po
-*******
+html2po, po2html
+****************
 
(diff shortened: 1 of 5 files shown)


nijel:

PR #4225, opened 2021-01-01
test: Remove condtion prior to assert

nijel:
It is not really useful - pytest shows the output only in case of failure,
so better to remove the condition and avoid having failure only branch
which would be never tested and could contain bugs.


nijel:
...and this reveals bugs: https://github.com/translate/translate/pull/4226

Issue #4232, opened 2021-01-12
pocount infinite loop


Jibec:
hi,

here is a code creating an infine loop with pocount:
```
#| identified as a comment
#|raise an infinite loop bug!
msgid "text"
msgstr "texte"
```

I feel like to happen somewhere there: https://github.com/translate/translate/blob/master/translate/storage/poparser.py#L93
and https://github.com/translate/translate/blob/master/translate/storage/poparser.py#L123

I understand the translate toolkit is able to detect changes of msgid?
changing the test looks breaks previous msgid detection.
but I am unsure to understand the usecase here.
We normally have the fuzzy strings to handle previous values, why would a translator keep the previous string id as a comment?

```
#. type: verbatim
#: darktable.pod:12
#, no-wrap
#| msgid ""
#|"    -d {all,cache,camctl,camsupport,control,dev,fswatch,imageio,input,\n"
#|"        ioporder,lighttable,lua,masks,memory,nan,opencl,params,perf,\n"
#|"        pwstorage,print,signal,sql,undo}\n"
#| "    --disable-opencl\n"
#| "    --library <library file>\n"
#| "    --datadir <data directory>\n"
#| "    --moduledir <module directory>\n"
#| "    --tmpdir <tmp directory>\n"
#| "    --configdir <user config directory>\n"
#| "    --cachedir <user cache directory>\n"
#| "    --localedir <locale directory>\n"
#| "    --luacmd <lua command>\n"
#| "    --conf <key>=<value>\n"
#| "    --noiseprofiles <noiseprofiles json file>\n"
#| "    --help\n"
#| "    --version\n"
#| "\n"
msgid ""
"    --cachedir <user cache directory>\n"
"    --conf <key>=<value>\n"
"    --configdir <user config directory>\n"
"    -d {all,cache,camctl,camsupport,control,dev,fswatch,imageio,input,\n"
"        ioporder,lighttable,lua,masks,memory,nan,opencl,params,perf,\n"
"        pwstorage,print,signal,sql,undo}\n"
"    --datadir <data directory>\n"
"    --disable-opencl\n"
"    -h, --help\n"
"    --library <library file>\n"
"    --localedir <locale directory>\n"
"    --luacmd <lua command>\n"
"    --moduledir <module directory>\n"
"    --noiseprofiles <noiseprofiles json file>\n"
"    -t <num openmp threads>\n"
"    --tmpdir <tmp directory>\n"
"    --version\n"
"\n"
msgstr ""
[...]

PR #4234, opened 2021-01-15
pocount: prevent infinite loop if space is missing

Jibec:
fix: #4232

### DIFF
--- a/translate/storage/poparser.py
+++ b/translate/storage/poparser.py
@@ -90,9 +90,14 @@ def read_prevmsgid_lines(parse_state):
     """
     prevmsgid_lines = []
     next_line = parse_state.next_line
-    while startswith(next_line, "#| ") or startswith(next_line, "| "):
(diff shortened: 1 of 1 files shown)


Jibec:
sorry, but I have no idea how to write the unit test!

nijel:
I still think that translate-toolkit should fail on parsing invalid files and not treat invalid syntax lines as comments (what I guess will happen here).

Jibec:
I know how to prevent failure and continue, but not really how to properly raise an error.

When processing 200 000 po files, I just would like to prevent an endless loop.

By the way, when patching Translate, I have no idea how to activate some debug level. Would you mind documenting it somewhere?

Why should it fail when a comment doesn't have the correct format and I want to get translation statistics? May it inpact the result?

nijel:
There is no logging inside the po parser, so there is nothing to enable, sorry.

I've made fix that will raise an error in https://github.com/translate/translate/pull/4235

PR #4236, opened 2021-01-16
cpo: Fix raising exceptions from ctypes

nijel:
It is not possible to raise an exception from ctype callback this way.

Store the exception in (thread local) variable and raise it when leaving the call.

Discovered when working on https://github.com/translate/translate/pull/4235, see https://stackoverflow.com/a/42443453/225718 for more details.

This causes tests to fail, I'd like to get feedback first whether such change looks okay.

### DIFF
--- a/translate/convert/test_po2dtd.py
+++ b/translate/convert/test_po2dtd.py
@@ -518,7 +518,7 @@ msgid "&Bookmarks"
 msgstr "Dipu&kutshwayo1"
 
 #: bookmarksItem.title
-msgctxt "bookmarksItem.title
+msgctxt "bookmarksItem.title"
 msgid "Bookmarks"
 msgstr "Dipukutshwayo2"
 
--- a/translate/storage/cpo.py
+++ b/translate/storage/cpo.py
@@ -33,6 +33,7 @@ import os
 import re
 import sys
 import tempfile
+import threading
 from ctypes import (
     CFUNCTYPE,
     POINTER,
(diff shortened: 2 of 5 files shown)


Jibec:
I installed this using `pip install git+https://github.com/nijel/translate@cpo` and did not discover any regression.
I would have expect the scripts to stop randomly, but no. See [log.txt](https://github.com/translate/translate/files/5840446/log.txt)

Here is the function call I do:
https://pagure.io/fedora-localization-statistics/blob/main/f/build_stats.py#_126

But in the log, I still have other output content that looks like error but don't raise anything.
The number of po files tested is a little bit more than 200 000 files.

This process takes a 110 minutes to run, I can re run it if needed.

nijel:
It internally raises an exception, but tools like pocount turn that into a warning.

Jibec:
Oh, indeed it's there: 
https://github.com/translate/translate/blob/master/translate/tools/pocount.py#L158

but why using "return" while the function calling calcstat is already 
handling errors?
https://github.com/translate/translate/blob/master/translate/tools/pocount.py#L468

would it be ok to remove the try/catch in calcstats?
and in handlefile function, display in log the filename would help when 
we launch pocount on a directory.

nijel:
I have no clue about the intents and history of that code, sorry. On the first look, the warning prints just warning, so that's something what was gracefully handled, while the `handlefile` prints an exception with stacktrace. Most likely this is consistent with how other covert tools behave...

PR #4239, opened 2021-01-19
Fix columnorder argument for csv2po/po2csv

slide333333:
Currently the `--columnorder` argument isnot correctly parsed for `csv2po` and `po2csv`. For example `po2csv --columnorder=source,target,location ...` is parsed as columns `['s', 'o', 'u', 'r', 'c', 'e', ',', 't', 'a', 'r', 'g', 'e', 't', ',', 'l', 'o', 'c', 'a', 't', 'i', 'o', 'n']`. The changes in this PR apply the correct parsing so the reulst is `["source", "target", "location"]`.

### DIFF
--- a/translate/convert/csv2po.py
+++ b/translate/convert/csv2po.py
@@ -248,6 +248,10 @@ def convertcsv(
     return 1
 
 
+def columnorder_callback(option, opt, value, parser):
+    setattr(parser.values, option.dest, value.split(','))
+
+
 def main(argv=None):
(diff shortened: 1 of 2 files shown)


nijel:
Can you please add tests to make sure this really works? 

PS: The black code formatter complains on using single quotes instead of double ones.

nijel:
Okay, I've added the tests and fixed coding style myself in https://github.com/translate/translate/pull/4260

Issue #3573, opened 2017-02-07
Simplify converters code


unho:
Most converters follow the same pattern so their code can be aligned to look the same and then simplified by putting the common code in a `BaseConverter` class.

Issue #3550, opened 2017-01-19
Check if po2json should output empty string for untranslated and fuzzy units


unho:
Right now `po2json` outputs empty strings for untranslated and fuzzy units. This doesn't align with the behavior I expected, which is the same `po2prop` has: output source text for the untranslated and fuzzy units.

Issue #3456, opened 2016-06-23
po2prop ignoring fuzzy translations with --fuzzy --removeuntranslated


anatolb:
##### The parameter *--fuzzy* does not work
when used together with *--removeuntranslated*, any translations marked fuzzy are not written to f_en.properties.

##### IMHO this is a bug, because 
1. a fuzzy translation is not 'untranslated', so should not be affected by *--removeuntranslated*
2. the help text for *--fuzzy* says "use translations marked fuzzy", which is not happening
``` ASCIIDOC
command line: po2prop -t f.properties f_en.po f_en.properties --fuzzy --removeuntranslated

verified with
  po2prop 1.14.0-alpha1
  po2prop 2.0.0b2
```

anatolb:
##### proposed solution (seems to work for me but not sure if it breaks something else)
``` diff
--- a/translate/convert/po2prop.py
+++ b/translate/convert/po2prop.py
@@ -166,7 +166,7 @@ class reprop(object):
                 delimiter = prespace + delimiter_char + postspace
             if key in self.inputstore.locationindex:
                 unit = self.inputstore.locationindex[key]
-                if not unit.istranslated() and bool(unit.source) and self.remove_untranslated:
+                if not unit.istranslated() and bool(unit.source) and self.remove_untranslated and not unit.isfuzzy():
                     returnline = u""
                 else:
                     if unit.isfuzzy() and not self.includefuzzy or len(unit.target) == 0:
```

PR #4257, opened 2021-02-02
po2json: Make sure that --fuzzy and --removeuntranslated can be used together.

rv0:
Currently when you combine the options --fuzzy and --removeuntranslated together, the fuzzy translations are also removed.
Both of these options work fine on their own, the problem lies in the check itself, that currently ignores the fuzzy parameter and will skip any fuzzy strings when --removeuntranslated is passed on.

The following PR will improve the logic in the "skip" check, to make sure that:
- When --removeuntranslated is passed, the string must be empty or missing to be skipped
or
- When --fuzzy is passed, the string will be included

### DIFF
--- a/translate/convert/po2json.py
+++ b/translate/convert/po2json.py
@@ -39,9 +39,8 @@ class rejson:
         self.inputstore.makeindex()
         for unit in self.templatestore.units:
             inputunit = self.inputstore.locationindex.get(unit.getid())
-            skip_unit = self.remove_untranslated and (
-                inputunit is None or inputunit.isfuzzy() or not inputunit.istranslated()
-            )
+            skip_unit = (self.remove_untranslated and (inputunit is None or not inputunit.istranslated())) or (
+                    not self.includefuzzy and inputunit.isfuzzy())
             if skip_unit:
                 continue
             if inputunit is not None:


rv0:
I made a mistake, will fix :)

rv0:
Not sure how to fix/handle/interpret the flake8 test that is failing.

Also, would love to add a unittest for these changes, but I haven't figured out yet how to run them

nijel (commented):
* Your changes break several `po2json` tests, can you please look at it?
* The flake8 issues can be easier addressed by running black code formatter, see http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html
* Testing is documented at http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/testing.html

rv0:
@nijel 
Thanks for providing me the information, got unit tests to work eventually and noticed the flaws of my initial patch. 

However, I don't think the test cases reflect the expected behaviour, so I changed them too. This might be radical so allow me to explain the logic used,

given that we start with po input:
```
#: .foo
msgid "foo"
msgstr "oof"

#: .bar
#, fuzzy
msgid "bar"
msgstr "rab"

#: .baz
#, fuzzy
msgid "baz"
msgstr ""

#: .qux
msgid "qux"
msgstr ""
"""
```

And json file:
```
{
    "foo": "foo",
    "bar": "bar",
    "baz": "baz",
    "qux": "qux"
}
```

I adapted the following tests:

### include_fuzzy=false and remove_untranslated=false

This was the expected output:
```
{
    "foo": "oof",
     "bar": "bar",
     "baz": "baz",
     "qux": "qux"
}
```
As both bar and baz are fuzzy, they should be removed if include_fuzzy is false:

```
{
    "foo": "oof",
     "qux": "qux"
}
```

### include_fuzzy=true and remove_untranslated=true

This was the expected output:
```
{
    "foo": "oof"
}
```

As bar is fuzzy and translated, imo it should also be included:

```
{
    "foo": "oof",
     "bar": "bar",
}
```


These changes seem logical to me, however doing a search in the issue queue I see they have caused confusion to others in the past, some examples:
https://github.com/translate/translate/issues/3456
https://github.com/translate/translate/issues/3550

Outside of the scope of this change however, I also think the code could be simplified a lot by adding an "allow_fuzzy=False" parameter to the global is_translated function, or even better just removing the fuzzy check from that function all together (because it's an assumption that is only clear when you read the source code), "not translated" and "fuzzy" mean two completely different things to me, but then again there may be conceptual or historical reasons behind all this that I'm completely missing.

nijel:
> Outside of the scope of this change however, I also think the code could be simplified a lot by adding an "allow_fuzzy=False" parameter to the global is_translated function, or even better just removing the fuzzy check from that function all together (because it's an assumption that is only clear when you read the source code), "not translated" and "fuzzy" mean two completely different things to me, but then again there may be conceptual or historical reasons behind all this that I'm completely missing.

This depends on how you interpret "is translated". Right now it means "the translation is completed", which definitely should exclude fuzzy (and other needing review states, as in xliff). I don't think it's reasonable to change the semantics as that could easily silently break third-party code. So, the unit is either untranslated, fuzzy or translated (or approved, but that's out of scope here).

rv0:
> > Outside of the scope of this change however, I also think the code could be simplified a lot by adding an "allow_fuzzy=False" parameter to the global is_translated function, or even better just removing the fuzzy check from that function all together (because it's an assumption that is only clear when you read the source code), "not translated" and "fuzzy" mean two completely different things to me, but then again there may be conceptual or historical reasons behind all this that I'm completely missing.
> 
> This depends on how you interpret "is translated". Right now it means "the translation is completed", which definitely should exclude fuzzy (and other needing review states, as in xliff). I don't think it's reasonable to change the semantics as that could easily silently break third-party code. So, the unit is either untranslated, fuzzy or translated (or approved, but that's out of scope here).

There's not really any indication that it means that a translation is _completed_, the comments in code say:

> Indicates whether this unit is translated.

As for the semantics, the GNU documentation does more or less say [a fuzzy entry is a translated entry](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html), for must purposes, only from the viewpoint of the translator one that needs additional review.
I think the fact that the difference is not clear is potentially a source of confusion and bugs.

In any case, the changes in my PR are made with the idea in mind that a fuzzy translation is a translation. Any comments on that?

nijel:
I have no clue about the original intents behind this behavior, I'm just describing how it behaves. Most of the things in the translate-toolkit are modelled around XLIFF, so it might be also the reason for this behavior.

Another issue with `--removeuntranslated` and `--no-fuzzy` is that each convertor implements this on its own and the behavior is not consistent. This is described in issues you've referenced.

> include_fuzzy=false and remove_untranslated=false
> As both bar and baz are fuzzy, they should be removed if include_fuzzy is false:

The question is whether they should be removed (what you implement) or source used instead (what is current implementation). `po2prop` and `po2dtd` both include source string in this case.

> include_fuzzy=true and remove_untranslated=true
> As bar is fuzzy and translated, imo it should also be included:

`po2prop` doesn't include fuzzy in this case and `po2dtd` does.

rv0:
I understand. So how to continue from here?

In my use case, I need to make sure no source strings end up in the localized json, as the strings are used with speech synthesis and this will make for example an English word to be spoken with a French voice. Not very good ;) If I don't include the source string, the application will automatically fall back to the English word and speak it with an English voice. That's fine.
On the other hand some words are machine translated and thus fuzzy, I want to include these in the localized json as in most cases the translation is good or good enough (using generic short sentences and machine learning driven translations is surprisingly effective nowadays).

I'd be happy to perform additional work on this PR if needed, I think my use-case is not that uncommon so a solution would benefit others, but I'm going to need some guidance on the approach to take. Perhaps additional parameters should be added?

nijel:
In the end the "correct" behavior for the first case depends on usage of the resulting file - whether all keys should be present there or not. And I think `--removeuntranslated` is there to indicate this and in case it's not present, then all keys should be present in the output, even with the source strings. Therefore, I'd prefer reverting this part of the PR.

For fuzzy inclusion, I think the behavior you've made makes more sense - when you explicitly specify `--fuzzy` you want these strings to be present (as `--nofuzzy` is the default behavior). Even with `--removeuntranslated`.

Don't get me wrong, I just don't want to merge change which will end up breaking existing workflows and this easily can do that. The perfect solution would be to have a single logic used by all the convertors (see https://github.com/translate/translate/issues/3573).

rv0:
Does any of the other converters use some sort of parameter to disable the inclusion of source strings? If so I can add that as well (either here or in another PR)

nijel:
I think `--removeuntranslated` is really the parameter which controls this, but it's supported only by few converters.

PR #4266, opened 2021-02-05
fix: do not modify XWikiPageProperties source file

### LINKED ISSUE #4265: XWiki Page Properties format does not handle correctly source edition
As explained in https://dev.xwiki.org/xwiki/bin/view/Community/XWiki%20Translations%20Formats/#HFileFormat: 
>  source translation files need to contain a XWiki.TranslationDocumentClass xobject as documented in Localization Extension

Right now editing the source file translation has the effect to remove this special class. You can see the result when using this format on Weblate here: https://github.com/xwiki-contrib/application-activitypub/pull/22/files#diff-0134e6af64b31dc0fc0e04d39cece9438a0cc36b411b5f474aee2f90ff5e162cL113

surli:
* Ensure that when performing a change in a XWikiPageProperties source
    file the XML is not modified. Fixes #4265

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -1257,6 +1257,9 @@ class XWikiPageProperties(xwikifile):
         self.root = None
         super(xwikifile, self).__init__(*args, **kwargs)
 
+    def is_source_file(self):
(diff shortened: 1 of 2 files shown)


nijel:

PR #4271, opened 2021-02-12
po2php: Use phpfile as a serializer

### LINKED ISSUE #3909: po2php fills translation files with source values rather than translations
Version 2.4.0

Hello, 

I have two translation folders (en and fr) with files following this structure:
```
<?php return [
	'key'    => 'value',
	'key2'       => 'value2',
];
```
When I do 

> php2po -P lang/en pot/ 

and 

> php2po -t lang/en lang/fr po-fr/

I get my pot and my po files where the English values are correctly translated to French.

However, when I do 

> po2php - t lang/en po-fr/ lang/fr 
it replaces the french values in the french translation files with the english source values.

Any idea what I'm doing wrong?

nijel:
* Avoid reimplementing parser and serializer and use phpfile for that.
* This ensures compatibility with php2po output (fixes #3909)
* This breaks some of the existing behaviour though:
    * Additional PHP statements are stripped from the file
    * Some whitespace is not preserved

### DIFF
--- a/translate/convert/po2php.py
+++ b/translate/convert/po2php.py
@@ -22,10 +22,7 @@ See: http://docs.translatehouse.org/projects/translate-toolkit/en/latest/command
 for examples and usage instructions.
 """
 
-import re
-
 from translate.convert import convert
-from translate.misc import quote
 from translate.storage import php, po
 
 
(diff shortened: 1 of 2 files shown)


nijel:
Merging as this seems better to me than the previous state.

PR #4303, opened 2021-03-25
Added --preserveplaceholders documentation to podebug.rst

asmosoinio:
Documented the feature from https://github.com/translate/translate/pull/3522

I could not find this anywhere in the docs.


asmosoinio:
ah this feature **is** documented in `--help` - should use same text in the .rst probably: 

```
  --preserveplaceholders
                        attempt to exclude characters that are part of
                        placeholders when performing character-level rewrites
                        so that consuming applications can still use the
                        placeholders to generate final output
```

nijel:

PR #4378, opened 2021-07-04
Add basic Fluent storage support

### LINKED ISSUE #3389: Add Fluent support to Translate Toolkit

str4d:
This uses `fluent.syntax` for parsing and serialization.

Implemented:
- Messages and terms.
- Comments (both attached to messages/terms and standalone).
- Errors when parsing Fluent files are recorded.
- Attributes from parsed Fluent files are exposed, and preserved on reserialization.

Part of #3389.

### DIFF
--- a/translate/storage/fluent.py
+++ b/translate/storage/fluent.py
@@ -138,6 +138,11 @@ class FluentUnit(base.TranslationUnit):
                 this._type = ast.Message
                 this._set_value(source_from_entry(node))
                 self.generic_visit(node)
+
+            def visit_Term(self, node):
+                this._type = ast.Term
+                this._set_value(source_from_entry(node))
+                self.generic_visit(node)
         Parser().visit(entry)
 
     def to_entry(self):
--- a/translate/storage/fluent.py
+++ b/translate/storage/fluent.py
@@ -61,6 +61,7 @@ class FluentUnit(base.TranslationUnit):
     def __init__(self, source=None, entry=None):
         super().__init__(source)
         self._id = None
+        self._errors = {}
         # The source and target are equivalent for monolingual units.
         self.target = source
         if source is not None:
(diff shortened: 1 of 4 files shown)


str4d:
Things I don't know how to handle yet (because I'm not familar with this library's API):

- How should Fluent's concept of plurals be exposed? (They are just specific usages of the general selector mechanism.)
- How should Fluent attributes be handled? Each attribute is an additional string that should be translated, so they aren't like XLIFF segments, but we also shouldn't represent them as separate translation units.

translate/storage/fluent.py
```diff
@@ -0,0 +1,191 @@
...
+        value = fp.maybe_get_pattern(FluentParserStream(self.source))
+
+        attributes = [ast.Attribute(
+            ast.Identifier(id),
+            fp.maybe_get_pattern(FluentParserStream(value)),
+        ) for (id, value) in self._attributes.items()]
+
+        comment = None
+        if self.getnotes():
+            comment = ast.Comment(self.getnotes())
```
nijel, line 160: Can you please add tests covering this as well?

nijel (commented):
Thanks, looks good. There are some coding style issues reported. Also, I'd like to have the comment conversion covered by the tests.

nijel:
@str4d Will you find time to look into this?

nijel:
Closing in favor of https://github.com/translate/translate/pull/4412 where I'm doing some fixes.

str4d:
Thanks! Sorry I didn't have time to pick this back up, been a very busy couple of months for me.

Issue #3238, opened 2014-12-12
Add support for Localized Property List File (OS X >= 10.9 and iOS >= 7)


nijel:
Some time ago [Apple introduced "Localized property list file"](https://developer.apple.com/library/ios/releasenotes/Foundation/RN-Foundation/#//apple_ref/doc/uid/TP30000742-CH2-SW56).

You can find more [here](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html) and another explanations [here](https://github.com/objcio/articles/blob/master/2014-02-10-string-localization.md#plural-and-gender-support).
Some nice guy also set up a test repository [here](https://github.com/ObjectiveToast/LocalizeTesting), demonstrating the usage of the new file (`.stringsdict`).

Any future plan to support this?

romanr:
This is a current official best practice for non plurals localization for macOS/iOS: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html#//apple_ref/doc/uid/10000171i-CH5-SW10    
Without support for `stringsdict` proper Localization of macOS/iOS can not be achieved. This is a show stopper level issue.

nijel:
The base format has support in Python stdlib: https://docs.python.org/3/library/plistlib.html

PR #4398, opened 2021-07-22
stringsdict: Add support for stringsdict storage

x-sheep:
#3238 

This implements Stringsdict as defined by Apple. While it can be used for regular strings, it's recommended to only use this format to store plurals.

Apple deviates from the CLDR Language Plural Rules standard in the following ways:
* All languages support the 'zero' category. This implementation prepends the 'zero' category in every language that doesn't define it already.
* The 'other' category is always used when a category has a missing string, including in languages that doesn't have 'other' defined in the list of tags. This is currently unsupported by this implementation, and the 'other' string is not preserved when parsing a file in a language that doesn't define it.

### DIFF
--- a/translate/storage/stringsdict.py
+++ b/translate/storage/stringsdict.py
@@ -68,7 +68,7 @@ class StringsDictFile(base.TranslationStore):
             return data.cldr_plural_categories
 
         locale = target_lang.replace("_", "-").split("-")[0]
-        tags = data.plural_tags.get(locale, data.cldr_plural_categories)
+        tags = data.plural_tags.get(locale, data.cldr_plural_categories).copy()
         if "zero" not in tags:
             tags.insert(0, "zero")
         return tags
--- a/translate/storage/stringsdict.py
+++ b/translate/storage/stringsdict.py
@@ -13,19 +13,18 @@ class StringsDictId(base.UnitId):
 
     def __str__(self):
         s = super().__str__()
-        if s[0] == ":":
+        if s.startswith(":"):
             return s[1:]
         return s
 
 
 class StringsDictUnit(base.DictUnit):
-    IdClass = StringsDictId
-
     """A single entry in a .stringsdict file.
     One entry represents either a localized format string, or a variable used
     within another string.
     """
 
+    IdClass = StringsDictId
     format_value_type = ""
 
     def __init__(self, source=None):
--- a/translate/storage/stringsdict.py
+++ b/translate/storage/stringsdict.py
@@ -8,7 +8,19 @@ from translate.misc.multistring import multistring
 from translate.storage import base
 
 
-class StringsDictUnit(base.TranslationUnit):
+class StringsDictId(base.UnitId):
+    KEY_SEPARATOR = ":"
+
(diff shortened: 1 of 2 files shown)


translate/storage/stringsdict.py
```diff
@@ -0,0 +1,154 @@
...
+        Note that 'zero' is always supported.
+        """
+        target_lang = self.gettargetlanguage()
+        if target_lang is None:
+            return data.cldr_plural_categories
+
+        locale = target_lang.replace("_", "-").split("-")[0]
+        tags = data.plural_tags.get(locale, data.cldr_plural_categories)
+        if "zero" not in tags:
+            tags.insert(0, "zero")
```
nijel, line 73: This modifies the list, please create a copy before changing it.

translate/storage/test_stringsdict.py
```diff
@@ -0,0 +1,138 @@
...
+        </dict>
+    </dict>
+</plist>"""
+        store = self.StoreClass()
+        store.parse(content)
+
+        assert store.units[0].source == "shopping-list"
+        assert store.units[1].source == "shopping-list[apple]"
+        assert store.units[2].source == "shopping-list[orange]"
+        assert store.units[3].source == "other-string"
```
nijel, line 73: Can you please add assertions on target as well?

translate/storage/stringsdict.py
```diff
@@ -0,0 +1,154 @@
...
+                    spec_type = value.get("NSStringFormatSpecTypeKey", "")
+                    if spec_type and spec_type != "NSStringPluralRuleType":
+                        raise ValueError(
+                            f"{innerkey} in {key} is not of NSStringPluralRuleType"
+                        )
+
+                    plural_tags = self.target_plural_tags
+                    plural_strings = [value.get(tag, "") for tag in plural_tags]
+
+                    u = self.UnitClass(f"{key}[{innerkey}]")
```
nijel, line 104: Can the key also contain [] characters? In case it can, it can break easily. DictUnit / DictStore classes handle this better.

nijel (commented):
Thanks, looks good, but needs a bit of polishing and extending the tests.

translate/storage/stringsdict.py
x-sheep: I've changed the base class to DictUnit, and added a UnitId subclass using `:` as a separator (in line with the `plistbuddy` command line tool). The parser now sets the unitid directly.

nijel:

nijel:
@x-sheep  Not sure what was your motivation for this, but do you plan to integrate this in Weblate as well (https://github.com/WeblateOrg/weblate/issues/623)? I'd just like to avoid wasting time when more people would be working on that ;-).

x-sheep:
Yes, I did write this because Weblate was missing support, but I don't think I'll be trying to fork Weblate and implement it myself.

nijel:
Okay, I've quickly written https://github.com/WeblateOrg/weblate/commit/4c86a2f38998b6c7117a35d31489a9ea53866a1f, feedback from real life testing is welcome.

Issue #4429, opened 2021-09-08
Comments break Fluent (FTL) serialization


flaki:
The current implementation seems to completely choke on any [free-standing Fluent comments](https://projectfluent.org/fluent/guide/comments.html) during serialization. Weblate imports the FTL seemingly fine, but exporting in any way (repository commits or file downloads) results in a fatal error:

- `TypeError - 'NoneType' object is not subscriptable`
- Traced to `"/home/weblate/weblate-env/lib/python3.8/site-packages/fluent/syntax/stream.py", line 13, in get return self.string[offset]`

Detailed error log:

<details>

```
Internal Server Error: /download/rustfest-cfp/cfp-form/en/

TypeError at /download/rustfest-cfp/cfp-form/en/
'NoneType' object is not subscriptable

Request Method: GET
Request URL: http://weblate.rustfest.global/download/rustfest-cfp/cfp-form/en/
Django Version: 3.2.7
Python Executable: /home/weblate/weblate-env/bin/uwsgi
Python Version: 3.8.10
Python Path: ['/', '/home/weblate/weblate-env/lib/python3.8/site-packages/git/ext/gitdb', '.', '', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/weblate/weblate-env/lib/python3.8/site-packages', '/home/weblate/weblate-env/lib/python3.8/site-packages/gitdb/ext/smmap']
Server time: Wed, 08 Sep 2021 17:59:26 +0000
Installed Applications:
['weblate.addons',
 'weblate.auth',
 'weblate.checks',
 'weblate.formats',
 'weblate.glossary',
 'weblate.machinery',
 'weblate.trans',
 'weblate.lang',
 'weblate_language_data',
 'weblate.memory',
 'weblate.screenshots',
 'weblate.fonts',
 'weblate.accounts',
 'weblate.configuration',
 'weblate.utils',
 'weblate.vcs',
 'weblate.wladmin',
 'weblate.metrics',
 'weblate',
 'weblate.gitexport',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.admin.apps.SimpleAdminConfig',
 'django.contrib.admindocs',
 'django.contrib.sitemaps',
 'django.contrib.humanize',
 'social_django',
 'crispy_forms',
 'compressor',
[...]

PR #4431, opened 2021-09-09
Fluent: Fix serialization of standalone comments

str4d:
Closes translate/translate#4429.


nijel:

PR #4533, opened 2022-01-24
ci: Add yaml formatter to pre-commit

nijel:
This ensures that yaml configurations for CI are consistently formatted.


phlax:
@nijel i used to format yaml, this way, but these days, having worked with some really complex yaml, i now always use least-indent formatting

just my 2p

nijel:
I used this one as it was most similar to existing formatting, I have no preference in this...

PR #4567, opened 2022-02-05
chore: unpin core GitHub Actions

nschonni:
Pointin to v2 will reduce some Dependabot noise, till there is a v3 release

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         submodules: recursive
     - name: Setup Python
-      uses: actions/setup-python@v2.3.2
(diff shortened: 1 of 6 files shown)


nijel:
I'm not sure this is a good move. It can make builds randomly fail without knowing reason. Investigating what has changed then takes quite some time, so for me it seems better to have extra PRs and reliable CI environment...

nschonni:
Up to you. I find that it makes sense for some third party Actions, but the core ones seem to follow semver pretty well

nijel:
Okay, let's give it a try. I'm not convinced that either of approaches is better, and having less commits in this area might be a good thing.

PR #4568, opened 2022-02-05
feat: drop Python 3.6

nschonni:
Maybe you want to keep it for a bit, but it hit EOL at the end of December https://devguide.python.org/devcycle/#end-of-life-branches


nijel:

PR #4587, opened 2022-03-04
setup: Rewrite to use current setuptools practices

### LINKED ISSUE #4582: Generated wheels contain duplicate files, breaking installer
The wheels generated by this package's `setup.py` contain duplicate copies of the Python files, e.g.:

```
$ python setup.py bdist_wheel
[...]
$ unzip -l dist/translate_toolkit-3.6.0-py3-none-any.whl | grep translate/__init__\.py
     1816  02-25-2022 23:20   translate/__init__.py
     1816  02-25-2022 23:20   usr/lib/python3.10/site-packages/translate/__init__.py
```

or:

```
$ python -m build -w
[...]
$ unzip -l dist/translate_toolkit-3.6.0-py3-none-any.whl | grep translate/__init__\.py
     1816  02-25-2022 23:20   translate/__init__.py
     1816  02-25-2022 23:20   translate_toolkit-3.6.0.data/data/lib/python3.10/site-packages/translate/__init__.py
```

My guess is that the .py files get picked up as data files somehow.

This is problematic not only because it fattens up wheels with duplicate files but also breaks installer, effectively making it impossible to switch to PEP517 builds on Gentoo:

```pytb
$ python -m installer --destdir=/tmp/z2 dist/translate_toolkit-3.6.0-py3-none-any.whl 
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/installer/__main__.py", line 84, in <module>
    main(sys.argv[1:], "python -m installer")
  File "/usr/lib/python3.10/site-packages/installer/__main__.py", line 80, in main
    installer.install(source, destination, {})
  File "/usr/lib/python3.10/site-packages/installer/_core.py", line 109, in install
    record = destination.write_file(
  File "/usr/lib/python3.10/site-packages/installer/destinations.py", line 207, in write_file
    return self.write_to_fs(scheme, path_, stream, is_executable)
  File "/usr/lib/python3.10/site-packages/installer/destinations.py", line 167, in write_to_fs
    raise FileExistsError(message)
FileExistsError: File already exists: /tmp/z2/usr/lib/python3.10/site-packages/translate/__init__.py
```
```

### PR DESCRIPTION
**Pull request by nijel**
- drop py2exe support, it is broken anyway (see #4169)
- drop a lot of custom code
- use setup.cfg to define static metadata
- fixes double presence of some files in wheels (fixes #4582)

TODO:

- [x] shipping some of the data is broken (eg. language models are not found)

### CODE REVIEW

File: pyproject.toml [hunk 1]
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel"
```
mgorny, line 4: Adding `wheel` was never necessary, and setuptools docs were fixed not to do that anymore recently.

mgorny:
Unfortunately, I can't figure out a way to install `share` without moving it inside `translate/`.

nijel:
Moving it is IMHO not a big issue. I just got busy with other stuff and won't get back to this in next weeks...

nijel:
It works well for me now. Reviews are welcome as this is quite big change.

Makefile
```diff
@@ -6,6 +6,7 @@ all: help
 
 build: docs
 	python setup.py sdist --formats=gztar
+	python setup.py bdist_wheel
```
mgorny, line 9: Not saying this is wrong or insisting you change it but most of the new projects use [build](https://pypi.org/project/build/) for this. Might be worth a shot since it does the build in a venv, so it verifies that all the build-time deps are pulled in correctly. Then it uses sdist to build wheel, so you have an extra check that sdist isn't missing anything. But this obviously implies that it needs to be installed prior.

```suggestion
	python -m build
```

(with no extra args it builds both sdist & wheel into `dist/`)

mgorny:
I think you need to add `__init__.py` to these directories, i.e.:

```
	translate/share/__init__.py
	translate/share/langmodels/__init__.py
```

nijel:
That will probably remove the warning, but there is no need for this directory to be importable, it contains only data...

mgorny:
I'll try and see if I can come up with some other solution.

mgorny:
One other solution I was able to get working is to use [automatic discovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery), i.e. remove `packages = ...` entirely. However, the docs say it's experimental, so you probably shouldn't be using it at this point.

I'm going to try asking upstream for help.

mgorny:
Opened: https://github.com/pypa/setuptools/discussions/3323

mgorny:
Ok, so apparently an alternative is to use:

```
[options]
packages = find_namespace:

[options.packages.find]                                                          
include =
    translate*
```

You need to explicitly specify the root package (the `include`), as otherwise it grabs pretty much everything found in the repo.

nijel:
Thanks for looking into this! I've switched to find_namespace, and it seems to work well.

mgorny:
Thank you for working on this! That's one problem out of my plate ;-).

Issue #4617, opened 2022-04-29
Backslash are not exported correctly for .strings files


NathanFallet:
This issue comes from https://github.com/WeblateOrg/weblate/issues/7562

When exporting strings in Android XML files, this line is present:
https://github.com/translate/translate/blob/e01d8579fbde01c0c347865283d8800c5358a883/translate/storage/aresource.py#L239

But for iOS Strings files, there is no equivalent. I think that it should be added here to solve the issue:
https://github.com/translate/translate/blob/e01d8579fbde01c0c347865283d8800c5358a883/translate/storage/properties.py#L541

PR #4618, opened 2022-04-29
Fixing backslash problems for iOS strings

NathanFallet:
Fixing #4617, and then https://github.com/WeblateOrg/weblate/issues/7562

### DIFF
--- a/translate/storage/properties.py
+++ b/translate/storage/properties.py
@@ -538,7 +538,7 @@ class DialectStrings(Dialect):
 
     @staticmethod
     def encode(string, encoding=None):
-        return string.replace("\n", r"\n").replace("\t", r"\t")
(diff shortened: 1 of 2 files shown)


nijel (commented):
You've changed `encode` method which takes care of serializing, but the test covers only parsing. Can you please extend the test to cover serialization as well?

NathanFallet:
You're right, I'm going to add it as well

NathanFallet:
@nijel it's done, you can review again and launch tests if it's okay

NathanFallet:
@nijel it's done, you can review again and launch tests if it's okay.
I've launched tests with and without the modification, and it works as expected (fails without, and succeed with)

nijel:

PR #4639, opened 2022-06-01
rc: Fix how identifiers are found after concatenated string

jon-turney:
Fix how identifiers are found after a concatenated string in MENU and
DIALOG resources.

After extract_test() has consumed one or more adjacent quoted string
tokens to concatenate, the following token is the identifier.

This fixes an oversight in 2c581f31.

### DIFF
--- a/tests/cli/data/test_rc2po/one.rc
+++ b/tests/cli/data/test_rc2po/one.rc
@@ -77,7 +77,7 @@ BEGIN
     PUSHBUTTON      "See license",IDC_LICENCIA,10,162,85,15
     RTEXT           "If you don't have internet, please use magic.",IDC_STATIC,23,105,120,18
     ICON            IDR_MAINFRAME,IDC_STATIC,44,74,20,20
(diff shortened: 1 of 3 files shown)


jon-turney:
e.g if we have res.rc

```
IDD_DIALOG DIALOG 0, 0, 340, 180
CAPTION "Caption"
BEGIN
    LTEXT           "Concatenated " "Strings",IDC_STATIC_HEADER,7,0,258,8,NOT WS_GROUP
END
```

then without this change, the output of `rc2po res.rc` contains:

```
#: DIALOG.IDD_DIALOG.LTEXT.%22Strings%22
msgid "Concatenated Strings"
msgstr ""
```

with this change, the source location is more correctly formed:

```
#: DIALOG.IDD_DIALOG.LTEXT.IDC_STATIC_HEADER
msgid "Concatenated Strings"
msgstr ""
```

nijel:
Can you please add this as a testcase for rc2po?

nijel:

PR #4652, opened 2022-06-13
Make 'rc2po --charset' do something

jon-turney:
At the moment, the specified encoding gets passed down via the kwargs of rcfile.__init__(), but that then ignores it when calling rcfile.parse(), which goes off and does it's autodetection based on `#pragma code_page`, or (if that is absent) utf-16-le or cp1252,

### DIFF
--- a/translate/storage/rc.py
+++ b/translate/storage/rc.py
@@ -308,7 +308,7 @@ class rcfile(base.TranslationStore):
     """This class represents a .rc file, made up of rcunits."""
 
     UnitClass = rcunit
-    default_encoding = "cp1252"
+    default_encoding = "auto"
 
     def __init__(self, inputfile=None, lang=None, sublang=None, **kwargs):
         """Construct an rcfile, optionally reading in from inputfile."""
@@ -319,7 +319,7 @@ class rcfile(base.TranslationStore):
         if inputfile is not None:
             rcsrc = inputfile.read()
             inputfile.close()
-            self.parse(rcsrc)
+            self.parse(rcsrc, self.encoding)
 
     def add_popup_units(self, pre_name, popup):
         """Transverses the popup tree making new units as needed."""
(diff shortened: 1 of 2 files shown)


jon-turney:
It seems that even after #4191, explicitly specifying the encoding on the command line doesn't do anything.

translate/storage/rc.py
```diff
@@ -308,7 +308,7 @@ class rcfile(base.TranslationStore):
     """This class represents a .rc file, made up of rcunits."""
 
     UnitClass = rcunit
-    default_encoding = "cp1252"
+    default_encoding = "auto"
```
nijel, line 311: Won't this cause regression for files that are actually cp1252? The autodetection between iso-8859-1 and cp1252 is not much reliable...

jon-turney: I don't think so, because I don't think this value does much currently.

It seems to be passed as a kwarg to rcfile.init() when rc.py is invoked by test_rc2po.py (but not rc2po.py, which has it's own setting for this), but is just ignored as it's not passed into rcfile.parse(), so we're currently always relying on the autodetection there.

jon-turney: Looking at this a bit more, I think I see what you mean: this does possibly change the encoding automatically selected if rcfile.__init__() is called without an encoding kwarg.

I adjusted things so 'cp1251' is the default_encoding to be tried if charset_normalizer can't determine one.

nijel: Going again through the code, `default_encoding` should be real encoding:

https://github.com/translate/translate/blob/4d12d3737241e6de6d8cc180179b06af5f752bda/translate/storage/base.py#L771

jon-turney: I hadn't noticed that usage, but I guess that explains something that had confused me about how some of the tests were behaving.

There seems to be a confusion of meanings of `default_encoding`.  Is it "the encoding used when one isn't specified" or "the fallback encoding to use when one isn't specified, we tried autodetection, and that failed"?

PR #4664, opened 2022-06-28
rc2po: Honor passed --charset

nijel:
- sync default with rcfile
- make rcfile honor passed encoding for parsing

Alternative fix for https://github.com/translate/translate/pull/4652

@jon-turney can you please check it works for your use case as well?

### DIFF
--- a/translate/convert/rc2po.py
+++ b/translate/convert/rc2po.py
@@ -138,13 +138,12 @@ def main(argv=None):
     parser = convert.ConvertOptionParser(
         formats, usetemplates=True, usepots=True, description=__doc__
     )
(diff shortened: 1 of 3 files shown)


jon-turney:
> can you please check it works for your use case as well?

Yes, this fixes my problem.  Thanks.

I notice that the output of `rc2po -h` still claims that the default for `--charset` is cp1252.  Is that actually accurate, or is the default to do some autodetection?

nijel:
Hmm, maybe we should make the autodetection default instead (it actually was before, but only thanks to ignoring the charset parameter).

PR #4667, opened 2022-06-29
Handle '\r' in rc file strings similarly to '\n'

jon-turney:


jon-turney:
According to the [lexer in windres](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=binutils/rclex.c;h=2dcca7dff660f73956109663667fac9ab9012490;hb=HEAD#l281) the set of escapes which can appear in a .rc file string are `\a\b\r\f\n\r\t\v', some of which still aren't handled after this...

nijel:

PR #4697, opened 2022-08-06
Allow numeric resource IDs in RC files

derula:
# Background

When using tools like Resource Hacker to extract an RC file from a compiled binary EXE, resource identifiers are not `#define`d because the name cannot be reproduced from the binary. Instead, the numeric IDs are used directly. Furthermore, in the case of string tables, there is a comma after the resource ID. Running `rc2po` on such files produces an empty result.

# The fix

Simply allow numeric IDs, and an optional comma after `id_control`. Everything else seems to fall into place nicely.

A comma after the resource ID seems to be a documented feature for string tables: https://docs.microsoft.com/en-us/windows/win32/menurc/stringtable-resource#examples
Whether this causes any issues with other resource types, I cannot say. Tests all pass, however.

# Not covered

Converting such PO files back to RC was not tested, as it's out of scope for me. I am mostly interested in the RC file parsing aspect. However, if that's an issue, I could look at that as well.

### DIFF
--- a/translate/storage/rc.py
+++ b/translate/storage/rc.py
@@ -206,6 +206,7 @@ def rc_statement():
     undefined_control = (
         Group(
             name_id.set_results_name("id_control")
+            + Optional(",")
             + delimited_list(
                 concatenated_string ^ constant ^ numbers ^ Group(combined_constants)
             ).set_results_name("values_")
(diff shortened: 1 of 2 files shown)


nijel:

PR #4714, opened 2022-09-02
Add new rewriter to spot easily what is not part of the translation file 

pfouque:
Looking for a tool to test my translation I found (remind) podebug with rewriter was a really good tool.
But It was missing an option to spot strings that have not been integrated yet.
I tried with Chef and Flipped, but it was still not enough obvious.

That's why I created the NSA rewriters 

Based on flipped it keeps variables but obfuscates all alphanumeric characters with a black square 

Perhaps it can serve the community, so I made this PR...

### DIFF
--- a/docs/commands/option_rewrite.rst
+++ b/docs/commands/option_rewrite.rst
@@ -143,3 +143,19 @@ messages are still readable.
    Run :doc:`pofilter` as a quick method to fix up incorrect changes, or
    upgrade to version 1.4.
 
+.. _option_rewrite#nsa:
+
+nsa
+=======
+
+.. versionadded:: dev
+
+Rewrites the source text with black square that looks like NSA folder.
+
+.. code-block:: po
+
+  msgid "English"
+  msgstr "▮▮▮▮▮▮▮"
+
+This allows a translator or programmer to check what is not part of the translation. 
+This way it's easy to spot what is not translated yet.
--- a/docs/commands/option_rewrite.rst
+++ b/docs/commands/option_rewrite.rst
@@ -143,19 +143,21 @@ messages are still readable.
    Run :doc:`pofilter` as a quick method to fix up incorrect changes, or
    upgrade to version 1.4.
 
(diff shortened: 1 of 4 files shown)


docs/commands/option_rewrite.rst
```diff
@@ -143,3 +143,19 @@ messages are still readable.
    Run :doc:`pofilter` as a quick method to fix up incorrect changes, or
    upgrade to version 1.4.
 
+.. _option_rewrite#nsa:
+
+nsa
+=======
```
nijel, line 149: The underline should have same lenth as the text:

```suggestion
===
```

docs/commands/option_rewrite.rst
```diff
@@ -143,3 +143,19 @@ messages are still readable.
...
+
+.. versionadded:: dev
```
nijel, line 151:
```suggestion
.. versionadded:: 3.7.3
```

docs/commands/podebug.rst
```diff
@@ -58,7 +58,7 @@ Options:
 -S, --timestamp       skip conversion if the output file has newer timestamp
 -f FORMAT, --format=FORMAT     specify format string
 --rewrite=STYLE        the translation rewrite style: :doc:`xxx, en, blank,
-                       chef  (v1.2), unicode (v1.2) <option_rewrite>`
+                       chef  (v1.2), unicode (v1.2), nsa (dev) <option_rewrite>`
```
nijel, line 61: This label is missing more styles, maybe use something more generic as label here would a better approach.

nijel: It links to this: https://translate--4714.org.readthedocs.build/projects/translate-toolkit/en/4714/commands/option_rewrite.html

pfouque: Thanks @nijel for the review! I renamed it "classified", and addressed the changes you pointed.

nijel:

Issue #4485, opened 2021-12-07
Support needs to be added for i18next JSON format v4


NoahAndrews:
https://www.i18next.com/misc/json-format
https://github.com/WeblateOrg/weblate/issues/6689

PR #4720, opened 2022-09-22
Add support for the i18next v4 format

cotillion:
This should fix translate/translate#4485

I don't really know what I'm doing in python so it would be great if someone could review this.

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -338,7 +338,6 @@ class I18NextUnit(JsonNestedUnit):
                 self.storevalue(output, value, override_key=self._item[i])
 
 
-
 class I18NextFile(JsonNestedFile):
     """A i18next v3 format, this is nested JSON with several additions.
 
@@ -410,7 +409,8 @@ class I18NextFile(JsonNestedFile):
             yield from parent
 
 
-I18NextV4Suffixes = [ "zero", "one", "two", "few", "many", "other" ]
+I18NextV4Suffixes = ["zero", "one", "two", "few", "many", "other"]
+
 
 class I18NextV4Unit(JsonNestedUnit):
     """A i18next v4 format, JSON with plurals.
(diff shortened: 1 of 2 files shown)


translate/storage/jsonl10n.py
```diff
@@ -409,6 +409,122 @@ def _extract_units(
             yield from parent
 
 
+I18NextV4Suffixes = ["zero", "one", "two", "few", "many", "other"]
```
nijel, line 412: I'd use `translate.lang.data.cldr_plural_categories` instead of redefining it here, it is based on CLDR anyway.

translate/storage/jsonl10n.py
```diff
@@ -409,6 +409,122 @@ def _extract_units(
...
+                # Check already processed items
+                if k in processed:
+                    continue
+
+                plurals = []
+                suffix = ""
+                plural_base = ""
+
+                if "_" in k:
+                    plural_base, suffix = k.rsplit("_", 1)
```
nijel, line 495: This will be more effective:

```suggestion
                plural_base, _sep, suffix = k.rpartition("_")
```

translate/storage/jsonl10n.py
```diff
@@ -409,6 +409,122 @@ def _extract_units(
...
+                plural_base, suffix = item[0].rsplit("_", 1)
+
+                if suffix in I18NextV4Suffixes:
+                    return plural_base
+            return item[0]
+
+        def get_plurals(count, base):
+            if count <= 2:
+                return [base + "_one", base + "_other"][:count]
+            return [f"{base}_{I18NextV4Suffixes[i]}" for i in range(count)]
```
nijel, line 439: Shouldn't this be language-dependent and match CLDR? `translate.lang.data.plural_tags` contains the per language mapping, see for example `GoI18NJsonFile` how it can be handled.

nijel:
I've done some fixes in https://github.com/translate/translate/pull/4802

PR #4728, opened 2022-09-30
[WIP] fix: remove extra `return ->` in laravel keys

### LINKED ISSUE #4563: Laravel PHP strings adds extra content to source lang keys
string adds the return -> (if php contains return array() or return[]-> (if php contains return [) to the key string.

example file:
```php
<?php
return [
    'welcome' => 'Welcome to our application',
    'apples' => 'There is one apple|There are many apples',
];
```

Returned keys:
```php
return[]->'welcome'
return[]->'apples'
```

should be:
```php
welcome
apples
```

related to https://github.com/WeblateOrg/weblate/issues/5633

Simoneu01:
Close #4563
 
Close https://github.com/WeblateOrg/weblate/issues/5633

### DIFF
--- a/translate/storage/test_php.py
+++ b/translate/storage/test_php.py
@@ -1201,3 +1201,65 @@ return [
         assert bytes(phpfile).decode() == phpsource
         phpunit.source = multistring(["There is an apple", "There are many apples"])
         assert bytes(phpfile).decode() == phpsource.replace("one apple", "an apple")
+
+    def test_comments(self):
+        # https://github.com/laravel/laravel/blob/e87bfd60ed4ca30109aa73c4d23250c113fe75ff/lang/en/auth.php
+        phpsource = r"""<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Authentication Language Lines
+    |--------------------------------------------------------------------------
+    |
+    | The following language lines are used during authentication for various
+    | messages that we need to display to the user. You are free to modify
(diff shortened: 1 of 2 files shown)


translate/storage/test_php.py
```diff
@@ -1182,22 +1182,62 @@ def phpparse(self, phpsource):
         return self.StoreClass(dummyfile)
 
     def test_plurals(self):
-        phpsource = r"""<?php
-return [
-    'welcome' => 'Welcome to our application',
-    'apples' => 'There is one apple|There are many apples',
-];
-"""
+        phpsource = open("./laravel/plurals.php").read()
```
nijel, line 1185: You can not make assumptions about current working directory. Either derive the filename from ``__file__`` and make sure the test files are included in the package, or inline it as it was done before.

nijel:
@Simoneu01 Any progress on this?

jekkos:
I tried this patch on weblate 3.14 and it resolves the issue with Laravel language import. What is holding us back to merge it?

Simoneu01:
> @Simoneu01 Any progress on this?

Sorry, I've been busy IRL. 

> I tried this patch on weblate 3.14 and it resolves the issue with Laravel language import. What is holding us back to merge it?

This PR has some issues when the language file has comments on it, and also breaks the indentation. I'll look into it in the next week.

nijel:
Fixed by https://github.com/translate/translate/pull/5815 instead.

PR #4734, opened 2022-10-16
doc: fix GPL link to GPLv2.0

### LINKED ISSUE #4733: GPL 2 ? GPL 3 ?
The github site shows GPL2, but the link from http://docs.translatehouse.org/projects/translate-toolkit/en/latest/license.html goes to the GPL3 page.

rffontenelle:
Fixes #4733

### DIFF
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -6,5 +6,5 @@ License
 *******
 
 The Translate Toolkit documentation is released under the `GNU General Public
-License (GPL) <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>`_.
+License version 2, or later (GPL-2.0-or-later) <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>`_.
 
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -6,5 +6,5 @@ License
 *******
 
 The Translate Toolkit documentation is released under the `GNU General Public
-License (GPL) <http://www.gnu.org/licenses/gpl.html>`_.
+License (GPL) <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>`_.
 


rffontenelle:
@nijel LGTM so enabling auto-merge, but pinging you just in case you have second opinion.

nschonni:
Probably good to clarify the 'or later' part too at the same time https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.html#VersionTwoOrLater

rffontenelle:
Disabled auto-merge to solve this discussion.

If we're adding "or later", then I believe the version should be specified. But I don't see how we can link without being hard to read. For instance, linking "or later":

> The Translate Toolkit documentation is released under the [GNU General Public License (GPL) version 2.0](http://www.gnu.org/licenses/gpl.html) [or later](https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.html#VersionTwoOrLater)

nijel:
When doing changes in this, it should cover all occurrences to be consistent. I can see:

* The copyright header for most of the files state version 2 or later.
* The Python package is GPL2 or later:

  https://github.com/translate/translate/blob/b0ffce8ae485870014b866e974d7abbe19115272/setup.cfg#L17

  That was introduced in https://github.com/translate/translate/pull/4587 without really thinking about that, but it seems to match the code.

* GitHub shows what is in https://github.com/translate/translate/blob/master/COPYING, updating the file will change what is displayed on GitHub.

So for now, I'd change the docs to state version 2 or later (it can still be a link to version 2 IMHO). We can update the licensing to version 3 or later if desired.

docs/license.rst
```diff
@@ -6,5 +6,5 @@ License
 *******
 
 The Translate Toolkit documentation is released under the `GNU General Public
-License (GPL) <http://www.gnu.org/licenses/gpl.html>`_.
+License (GPL) <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>`_.
```
nijel, line 9: Update to GPL 2 or later:

```suggestion
License version 2, or later (GPL-2.0-or-later) <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>`_.
```

PR #4751, opened 2022-11-05
feat: avoid xml self-closing tags

warlof:
Business Central is not using self-closing tags while generating and maintaining XLIFF files.

As a result, when used together with Weblate, there are huge diff which are generated after captions beeing translated - often leading to invalid merge conflict. In order to avoid this case, we swap serialize library from LXML to standard Xml which allow us to skip self-closing tag when an element is empty using `short_empty_elements` parameter from the ToString helper.

Example :
- https://github.com/StefanMaron/MSDyn365BC.Code.History/tree/fr-20/Email%20-%20Current%20User%20Connector/Source/Email%20-%20Current%20User%20Connector/Translations (especially with `note` tag)

Related to rvanbekkum/vsc-xliff-sync#97
Related to WeblateOrg/weblate#8352

### DIFF
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -362,7 +362,7 @@ class LISAfile(base.TranslationStore):
             )
         else:
             if self.XMLdoctype:
-                out.write(b"" + self.XMLdoctype + "\n")
+                out.write(f"{self.XMLdoctype}\n".encode(self.encoding))
             stream = io.BytesIO()
             ElementTree(root).write(
                 stream,
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -333,11 +333,29 @@ class LISAfile(base.TranslationStore):
     def serialize(self, out=None):
         """Converts to a string containing the file's XML"""
         root = self.document.getroot()
+        xml_declaration = ""
+
         if self.XMLdoublequotes:
             if self.XMLuppercaseEncoding:
-                out.write(b'<?xml version="1.0" encoding="UTF-8"?>\n')
+                xml_declaration = (
+                    '<?xml version="1.0" encoding="' + self.encoding.upper() + '"?>\n'
+                )
             else:
-                out.write(b'<?xml version="1.0" encoding="utf-8"?>\n')
+                xml_declaration = (
+                    '<?xml version="1.0" encoding="' + self.encoding.lower() + '"?>\n'
+                )
+        else:
+            if self.XMLuppercaseEncoding:
+                xml_declaration = (
+                    "<?xml version='1.0' encoding='" + self.encoding.upper() + "'?>\n"
+                )
+            else:
+                xml_declaration = (
+                    "<?xml version='1.0' encoding='" + self.encoding.lower() + "'?>\n"
+                )
+
+        out.write(xml_declaration.encode(self.encoding))
+
         if self.XMLindent:
             reindent(root, **self.XMLindent)
         # When XMLSelfClosingTags flag is active, relay
@@ -347,18 +365,16 @@ class LISAfile(base.TranslationStore):
             treestring = etree.tostring(
                 self.document,
                 pretty_print=not self.XMLindent,
-                xml_declaration=not self.XMLdoublequotes,
                 encoding=self.encoding,
                 doctype=self.XMLdoctype,
             )
         else:
             if self.XMLdoctype:
-                out.write(b'' + self.XMLdoctype + '\n')
+                out.write(b"" + self.XMLdoctype + "\n")
             ElementTree.register_namespace("", self.namespace)
             treestring = ElementTree.tostring(
                 root,
                 encoding=self.encoding,
-                xml_declaration=not self.XMLdoublequotes,
                 short_empty_elements=False,
             )
         treestring = self.serialize_hook(treestring)
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -333,26 +333,16 @@ class LISAfile(base.TranslationStore):
     def serialize(self, out=None):
         """Converts to a string containing the file's XML"""
         root = self.document.getroot()
-        xml_declaration = ""
+        xml_quote_format = "'"
+        xml_encoding = self.encoding.lower()
 
         if self.XMLdoublequotes:
-            if self.XMLuppercaseEncoding:
-                xml_declaration = (
-                    '<?xml version="1.0" encoding="' + self.encoding.upper() + '"?>\n'
-                )
-            else:
-                xml_declaration = (
-                    '<?xml version="1.0" encoding="' + self.encoding.lower() + '"?>\n'
-                )
-        else:
-            if self.XMLuppercaseEncoding:
-                xml_declaration = (
-                    "<?xml version='1.0' encoding='" + self.encoding.upper() + "'?>\n"
-                )
-            else:
-                xml_declaration = (
-                    "<?xml version='1.0' encoding='" + self.encoding.lower() + "'?>\n"
-                )
+            xml_quote_format = '"'
+
+        if self.XMLuppercaseEncoding:
+            xml_encoding = self.encoding.upper()
+
+        xml_declaration = f"<?xml version={xml_quote_format}1.0{xml_quote_format} encoding={xml_quote_format}{xml_encoding}{xml_quote_format}?>\n"
 
         out.write(xml_declaration.encode(self.encoding))
 
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -18,6 +18,8 @@
 
 """Parent class for LISA standards (TMX, TBX, XLIFF)"""
 
+from xml.etree import ElementTree
+
 from lxml import etree
 
 from translate.misc.xml_helpers import (
@@ -265,6 +267,8 @@ class LISAfile(base.TranslationStore):
     XMLdoublequotes = True
     XMLdoctype = None
     XMLuppercaseEncoding = True
+    # Determine how empty tags should be serialized (<note></note> or <note />)
+    XMLSelfClosingTags = True
 
     namespace = None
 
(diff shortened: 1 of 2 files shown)


nijel (changes requested):
Switching to a different serializer breaks 63 tests, can you please make the change less intrusive?

warlof:
sorry about that, I spent the overall day to figure where and how to make that change and once done, running tests was failing against I/O exception :(

what about doing the following changes :
- adding an extra property into LISA class called `UseSelfClosingTag` - based on that new flag, determine which serializer and format to use, default to `True` which will avoid breaking changes
- adding an alternative implementation of XliffFile which set the `UseSelfClosingTag` to false and override the existing `XliffFile` class ?

warlof:
@nijel which way you prefer I handle the xliff without self closing tag part ?

Currently, I already made changes which prevent other formatters to be impacted.

Now, I can either
- add a XliffFileWithoutSelfClosingTags, which override XliffFile and set the new flag to true
- update xliff test file which contain self-closing tags to use the non self-closing format instead

nijel:
I'd prefer to avoid changes in existing behavior. It's used by thousands of users and changing the output might break their workflows.

When Weblate is the main focus, it might be enough to add the class attribute and change the value in the customization add-on, similarly like we do for YAML or JSON (see https://docs.weblate.org/en/latest/admin/addons.html#customize-json-output).

translate/storage/lisa.py
```diff
@@ -329,20 +333,50 @@ def serialize_hook(treestring):
...
+        else:
+            if self.XMLuppercaseEncoding:
+                xml_declaration = (
+                    "<?xml version='1.0' encoding='" + self.encoding.upper() + "'?>\n"
+                )
             else:
-                out.write(b'<?xml version="1.0" encoding="utf-8"?>\n')
+                xml_declaration = (
+                    "<?xml version='1.0' encoding='" + self.encoding.lower() + "'?>\n"
+                )
```
nijel, line 355: When doing all this manually, it would be better to set `quote` and `encoding` based on the attributes and then build the XML declaration just once:

```suggestion
        xml_declaration = f"<?xml version={quote}1.0{quote} encoding={quote}{encoding}{quote}?>\n"
```

warlof:
I've updated the xml_declaration including your changes which avoid a chit load of if/else stuff and revert the change applied to the xliff formatter - so default is still self-closing tags.

I'll take a look at the addon thing and submit changes for the xliff formatting against the Weblate repository.
That would definitely did the trick since it might be an optional setup for a specific component setup in Weblate.

warlof:
Hi, I had clean a bit the branch history
I also confirm XML Addon has been implemented onto weblate repository and is working like a charm.

Thank you for your guidelines :)

translate/storage/lisa.py
```diff
@@ -329,20 +333,40 @@ def serialize_hook(treestring):
...
+            treestring = etree.tostring(
+                self.document,
+                pretty_print=not self.XMLindent,
+                encoding=self.encoding,
+                doctype=self.XMLdoctype,
+            )
+        else:
+            if self.XMLdoctype:
+                out.write(b"" + self.XMLdoctype + "\n")
+            ElementTree.register_namespace("", self.namespace)
```
nijel, line 364: This worries me as the registry is global and can influence other ElementTree uses. It is nearly impossible to test that this doesn't break anything inside complex applications like Weblate.

Maybe it would be better to use `default_namespace` of `tostring`? Unfortunately, it is Python 3.8+ only.

See https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.register_namespace and https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.tostring

translate/storage/lisa.py
```diff
@@ -329,20 +333,40 @@ def serialize_hook(treestring):
...
+            treestring = ElementTree.tostring(
+                root,
+                encoding=self.encoding,
+                short_empty_elements=False,
+            )
```
nijel, line 369: Actually `default_namespace` is present in `write` for all supported Python versions, and `tostring` is just a thin wrapper around it, so we could use `write` directly as well.

Something like this (not tested at all):

```suggestion
            stream = io.BytesIO()
            ElementTree(root).write(stream, self.encoding, short_empty_elements=False)
            treestring = stream.getvalue()
```

translate/storage/lisa.py
warlof: My python knowledge does not allow me to make such arbitrary 😶

both translate and Weblate are requiring python 3.7 as a minimum version which is mostly the reason why I didn't use this way.

translate/storage/lisa.py
```diff
@@ -329,20 +333,38 @@ def serialize_hook(treestring):
...
+                pretty_print=not self.XMLindent,
+                xml_declaration=False,
+                encoding=self.encoding,
+                doctype=self.XMLdoctype,
+            )
+        else:
+            if self.XMLdoctype:
+                out.write(b"" + self.XMLdoctype + "\n")
+            stream = io.BytesIO()
+            ElementTree(root).write(stream, self.encoding, short_empty_elements=False)
```
nijel, line 366: Sorry, seems I've missed the namespace from my suggestion:

```suggestion
            ElementTree(root).write(stream, self.encoding, short_empty_elements=False, default_namespace=self.namespace)
```

translate/storage/lisa.py
```diff
@@ -329,20 +333,38 @@ def serialize_hook(treestring):
...
+            treestring = etree.tostring(
+                self.document,
+                pretty_print=not self.XMLindent,
+                xml_declaration=False,
+                encoding=self.encoding,
+                doctype=self.XMLdoctype,
+            )
+        else:
+            if self.XMLdoctype:
+                out.write(b"" + self.XMLdoctype + "\n")
```
nijel, line 364: Better to use encode to correct encoding (just in case somebody will use non-ascii in the doctype):

```suggestion
                out.write(f"{self.XMLdoctype\n".encode(self.encoding))
```

translate/storage/lisa.py
warlof: Shit append 🙃

translate/storage/lisa.py
```diff
@@ -18,6 +18,8 @@
 
 """Parent class for LISA standards (TMX, TBX, XLIFF)"""
 
+from xml.etree import ElementTree
```
nijel, line 21: Missing import for the new code:

```suggestion
import io

from xml.etree import ElementTree
```

nijel (commented):
Thanks for the changes, I think the code now looks good. What is missing is test case that would verify that the feature works as intended.

warlof:
I had an exception while running those changes with Python 3.9

Made two changes :

- switch from `ElementTree(root).write(` to `ElementTree.ElementTree(root).write(`
- also using `default_namespace` is generating an exception - so I revert this to namespace registration => https://stackoverflow.com/questions/18338807/cannot-write-xml-file-with-default-namespace

Also added a simple test

translate/storage/test_xliff.py
```diff
@@ -643,3 +645,29 @@ def test_preserve():
...
+  </file>
+</xliff>
+"""
+        xlifffile = xliff.xlifffile.parsestring(xlfsource)
+        xlifffile.XMLSelfClosingTags = False
+        xlifffile.XMLuppercaseEncoding = False
+
+        a = BytesIO()
+        xlifffile.serialize(a)
+        assert a.getvalue().decode("utf-8") == f"{xlfsource}"
```
nijel, line 673: No need for BytesIO here:

```suggestion
        assert bytes(xlifffile).decode("utf-8") == xlfsource
```

translate/storage/lisa.py
```diff
@@ -329,20 +334,45 @@ def serialize_hook(treestring):
...
+        else:
+            if self.XMLdoctype:
+                out.write(f"{self.XMLdoctype}\n".encode(self.encoding))
+            stream = io.BytesIO()
+
+            ElementTree.register_namespace("", self.namespace)
+            ElementTree.ElementTree(root).write(
+                stream,
+                self.encoding,
+                short_empty_elements=False,
```
nijel, line 372: This didn't work?

```suggestion
            ElementTree.ElementTree(root).write(
                stream,
                self.encoding,
                short_empty_elements=False,
                default_namespace=self.namespace,
```

warlof: negative, it seems it fall in a kind of caveat

```
test_xliff.py:671: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
base.py:734: in __bytes__
    self.serialize(out)
xliff.py:870: in serialize
    super().serialize(out)
lisa.py:369: in serialize
    ElementTree.ElementTree(root).write(
C:\Python39\lib\xml\etree\ElementTree.py:746: in write
    qnames, namespaces = _namespaces(self._root, default_namespace)
C:\Python39\lib\xml\etree\ElementTree.py:863: in _namespaces
    add_qname(key)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

qname = 'version'

    def add_qname(qname):
        # calculate serialized qname representation
        try:
            if qname[:1] == "{":
                uri, tag = qname[1:].rsplit("}", 1)
                prefix = namespaces.get(uri)
                if prefix is None:
                    prefix = _namespace_map.get(uri)
                    if prefix is None:
                        prefix = "ns%d" % len(namespaces)
                    if prefix != "xml":
                        namespaces[uri] = prefix
                if prefix:
                    qnames[qname] = "%s:%s" % (prefix, tag)
                else:
                    qnames[qname] = tag # default element
            else:
                if default_namespace:
                    # FIXME: can this be handled in XML 1.0?
>                   raise ValueError(
                        "cannot use non-qualified names with "
                        "default_namespace option"
                        )
E                       ValueError: cannot use non-qualified names with default_namespace option

C:\Python39\lib\xml\etree\ElementTree.py:840: ValueError
```

translate/storage/lisa.py
```diff
@@ -329,20 +334,45 @@ def serialize_hook(treestring):
...
+            )
+            treestring = stream.getvalue()
+
```
nijel, line 375: Okay, I think we need to revisit this. `ElementTree.register_namespace` will definitely have tendency to break behavior in some situations (at least in any async or threaded processes). We should stick to using parameters only, not altering the global state.

According to https://stackoverflow.com/a/34177354/225718 `method="c14n"` should do the trick for lxml.

So, how about this?

```suggestion
        if self.XMLSelfClosingTags and self.encoding == "utf-8":
             kwargs = {"method": "c14n"}
        else:
             kwargs = {"encoding": self.encoding}

        treestring = etree.tostring(
            self.document,
            pretty_print=not self.XMLindent,
            xml_declaration=False,
            doctype=self.XMLdoctype,
            **kwargs,
        )
```

warlof: I think it does
I'll make further check tomorrow afternoon

nijel:
Closin in favor of https://github.com/translate/translate/pull/4776

PR #4776, opened 2022-12-06
feat(lisa): avoid xml self-closing tags

nijel:
- ability to override the XML format for XLIFF files
- self-closing tag is set to true by default to prevent conflict with current implementation
- capability will be defer to Weblate as XLIFF Addon like JSON Format (https://docs.weblate.org/en/latest/admin/addons.html#customize-json-output)
- always generate XML declaration in translate-toolkit for consistency

Co-authored-by: Michal Čihař <michal@cihar.com>

### DIFF
--- a/translate/misc/xml_helpers.py
+++ b/translate/misc/xml_helpers.py
@@ -184,6 +184,23 @@ def reindent(
             elem.tail = i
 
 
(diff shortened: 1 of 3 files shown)


warlof:
I confirm it's working as expected.
Thank you @nijel :)

PR #4798, opened 2023-01-11
json: support gotext JSON format

### LINKED ISSUE #4450: Add support for gotext format
Please consider adding a support of gotext translation format. It is a golang's native translation format based on JSON. Unfortunately it is poorly documented.

References:
- [An intro article](https://www.alexedwards.net/blog/i18n-managing-translations), not comprehensive, just to make an idea.
- [golang.org/x/text/message/pipeline](https://pkg.go.dev/golang.org/x/text/message/pipeline#Messages) package documentation. It provides data structures serialized to JSON.
- [golang.org/x/text/feature/plural](https://pkg.go.dev/golang.org/x/text/feature/plural#Selectf) package documentation, contains description of how messages are pluralized.

SamWhited:
I started experimenting with this as I'd very much like this to be supported in Weblate, however my Python is extremely rusty, so let me know if anything is amiss.

Fixes #4450

Signed-off-by: Sam Whited <sam@samwhited.com>

### DIFF
--- a/translate/storage/jsonl10n.py
+++ b/translate/storage/jsonl10n.py
@@ -416,6 +416,132 @@ class I18NextFile(JsonNestedFile):
             yield from parent
 
 
+class GoTextJsonUnit(BaseJsonUnit):
+    ID_FORMAT = "{}"
+
+    def __init__(
+        self,
+        source=None,
+        item=None,
+        notes=None,
+        placeholders=None,
+        comment=None,
+        message=None,
+        meaning=None,
+        key=None,
+        fuzzy=None,
+        position=None,
+        **kwargs,
+    ):
+        super().__init__(source, item, notes, placeholders)
+        self.comment = comment
+        self.message = message
+        self.meaning = meaning
+        self.key = key
+        self.fuzzy = fuzzy
+        self.position = position
+
+    def getvalue(self):
+        target = self.target
+        if isinstance(target, multistring):
+            strings = list(target.strings)
+            if len(self._store.plural_tags) > len(target.strings):
+                strings += [""] * (len(self._store.plural_tags) - len(target.strings))
+            target = {
+                "select": {
+                    "feature": "plural",
+                }
+            }
+            if self.placeholders:
+                target["select"]["arg"] = self.placeholders[0]["id"]
+            target["select"]["cases"] = {
+                plural: {"msg": strings[offset]}
+                for offset, plural in enumerate(self._store.plural_tags)
+            }
+        value = {"id": self.getid()}
+        if self.message:
+            value["message"] = self.message
+        if self.notes:
+            value["translatorComment"] = self.notes
+        if self.comment:
+            value["comment"] = self.comment
+        if self.key:
+            value["key"] = self.key
+        if self.fuzzy:
+            value["fuzzy"] = self.fuzzy
+        if self.position:
+            value["position"] = self.position
+        value["translation"] = target
+        if self.placeholders:
+            value["placeholders"] = self.placeholders
+        return value
+
+
+class GoTextJsonFile(JsonFile):
+    """gotext JSON file
+
+    See following URLs for doc:
+
+    https://pkg.go.dev/golang.org/x/text/cmd/gotext
+    https://github.com/golang/text/tree/master/cmd/gotext/examples/extract/locales/en-US
+    """
+
+    UnitClass = GoTextJsonUnit
+
+    @property
+    def plural_tags(self):
+        return plural_tags.get(self.gettargetlanguage(), plural_tags["en"])
+
+    def _extract_units(
+        self,
+        data,
+        stop=None,
+        prev=None,
+        name_node=None,
+        name_last_node=None,
+        last_node=None,
+    ):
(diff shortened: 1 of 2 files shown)


translate/storage/jsonl10n.py
```diff
@@ -416,6 +416,94 @@ def _extract_units(
...
+
+    https://pkg.go.dev/golang.org/x/text/cmd/gotext
+    https://github.com/golang/text/tree/master/cmd/gotext/examples/extract/locales/en-US
+    """
+
+    UnitClass = GoTextJsonUnit
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        locale = self.gettargetlanguage()
```
nijel, line 458: You should implement gettargetlanguage/settargetlanguage to operate on `data['language']`.

translate/storage/jsonl10n.py
```diff
@@ -416,6 +416,94 @@ def _extract_units(
...
+    def _extract_units(
+        self,
+        data,
+        stop=None,
+        prev=None,
+        name_node=None,
+        name_last_node=None,
+        last_node=None,
+    ):
+        for value in data:
```
nijel, line 477: Shouldn't this process only messages?

```suggestion
        for value in data["messages"]:
```

Just guessing, as I saw the test failure.

SamWhited: Fixed, thanks!

translate/storage/jsonl10n.py
SamWhited: Working on this now; thank you!

SamWhited: I pushed a possible fix; it doesn't work in the way you suggested, but LMK what you think. When trying to override gettargetlanguage I couldn't figure out if we had the whole file parsed already or not, so I thought it might make sense to allow it to be file based (it's normally stored in language name folders anyways, so this will work), but then override it if we detect a language in the file when we parse.

SamWhited:
Looking at the current test run it doesn't seem to preserve other unknown elements in the JSON. I'm not really sure what to do about that; eg. "message" is the source string (it's not really necessary in the file as far as I can tell, it's just in the file you send off to translators so that the source string is readily available), or the placeholder stuff that defines variables (I didn't support this deliberately and just always used a default variable name, more complicated support can come later if necessary).

nijel:
There is existing support for storing placeholders (see WebExtension classes), so that can be easily used for that. As for `message` it's good to keep it if it's present, so you will need an additional attribute to store and serialize that as well.

translate/storage/jsonl10n.py
```diff
@@ -416,6 +416,90 @@ def _extract_units(
...
+                        cases.get(key, {}).get("msg")
+                        for key in cldr_plural_categories
+                        if key in cases
+                    ]
+                )
+            unit = self.UnitClass(
+                translation,
+                value.get("id", ""),
+                value.get("translatorComment", ""),
+            )
```
nijel, line 489: You can store placeholders on Unit this way:

```suggestion
                placeholders=value.get("placeholders", []),
            )
```

SamWhited: I've started experimenting with this, but I still won't have a way to know which placeholder is the plurals one wgeb cakkubg `getvalue`, I might just remove support for now as I think this is a bit beyond my knowledge of how this library works; we can always add it back later and just have support for simple files to start, I think

SamWhited: Update: pushed the various bits that can be part of message that we might want to preserve including the placeholder stuff. I feel like there *really* has to be a better way to preserve bits we don't care about for the sake of translations, but which should still be preserved across import/export of the file though?

translate/storage/jsonl10n.py
```diff
@@ -416,6 +416,120 @@ def _extract_units(
...
+        if isinstance(target, multistring):
+            strings = list(target.strings)
+            if len(self._store.plural_tags) > len(target.strings):
+                strings += [""] * (len(self._store.plural_tags) - len(target.strings))
+            target = {
+                "select": {
+                    "feature": "plural",
+                }
+            }
+            if self.placeholders and len(self.placeholders) > 0:
```
nijel, line 451: Checking for length is not needed, these conditions are equivalent for a list.

```suggestion
            if self.placeholders:
```

translate/storage/jsonl10n.py
```diff
@@ -416,6 +416,128 @@ def _extract_units(
...
+
+    def __init__(
+        self,
+        comment=None,
+        message=None,
+        meaning=None,
+        key=None,
+        fuzzy=None,
+        position=None,
+        **kwargs,
```
nijel, line 430: Please keep existing positional args before new ones, otherwise something will break as that changes API.

SamWhited: I don't know what this means, sorry, none of these new arguments are positional, they're all named?

SamWhited: Nevermind, I found an example and I see what you mean. Done.

nijel:

PR #4805, opened 2023-01-13
docs: update the year in the copyright footer

### LINKED ISSUE #4804: Outdated footer on the documentation website
The copyright notice in the docs webpage still state 2022 instead of 2023.

fabianegli:
Closes #4804

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,7 +21,7 @@ sys.path.insert(0, os.path.abspath(".."))
 # -- Project information -----------------------------------------------------
 
 project = "Translate Toolkit"
-copyright = "2002-2022, Translate"
+copyright = "2002-2023, Translate"
 
 # The short X.Y version.
 version = "3.8.1"


nijel:
I'd instead remove the year, there is no good reason for keeping it around. See for example https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

fabianegli:
I left a note on the website you linked about why I think it is valuable despite not being necessary to retain copyright.

> Copyright is a way of protecting intellectual property so a person may reap the benefit of their work. But as mentioned above, it expires. While a copyright notice may not be required to assert the copyright, it will allow someone to assess that the work is not subject to copyright any more after it expired. It will otherwise be hard in to assess when the last changes have been made after 70 years of inactivity in the project - also because much of the metadata associated with files may have been lost.
>
>While appreciating that this is a hassle, I think the long term benefits may outweigh the cost. After all, it's not a quick change to make and only has to be done if work is being done anyway.

nijel:
Most copyright notices here are not properly maintained, updating the year does not improve it.

When talking about this particular one, what is actually the copyright holder "Translate" in it?

fabianegli:
I just wanted to relieve some of the burden on the devs. I have an opinion, but it is not a loosely held one and I won't have any hard feelings if this PR is closed and forgotten. I don't think there's truly a right or wrong way to do this.

nijel:
Okay, let's merge this, as I currently don't have time to make more changes in this area.

Issue #4649, opened 2022-06-08
Translate Toolkit should expose attributes in Fluent's storages


pan93412:
Related: https://github.com/WeblateOrg/weblate/issues/7742#issuecomment-1147283794

nijel:
To clarify - the attributes are now exposed as a dictionary:

https://github.com/translate/translate/blob/650b3d28d8a739beef827e0261dc409e4add799c/translate/storage/fluent.py#L93-L94

This way it really doesn't fit into way translatable strings are exposed. Each attribute should become a separate `Unit` and some folding when serializing will be needed.

emmapeel2:
We want to move Tor Browser localization to Fluent, but we also have problems to load this file: https://gitlab.torproject.org/tpo/translation/-/blob/basebrowser-newidentityftl/en-US/new-identity.ftl

nijel:
I currently do not have any plans on this.

henry-torproject:
> Each attribute should become a separate `Unit` and some folding when serializing will be needed.

@nijel would it be possible to keep the attributes grouped together within the same unit, rather than split attributes into separate units? Or split them, but still associate them somehow?

Ideally, weblate's interface would be able to show the attributes grouped together along with their attribute names, because they can provide important context to each other. Similar to how  pontoon handles it https://pontoon.mozilla.org/es-ES/firefox/browser/browser/aboutDialog.ftl/?string=192297

I guess weblate has something similar set up for handling plurals, which it groups together in the same page.

nijel:
There is nothing in Weblate that would fit this. Looking at the example posted in https://github.com/WeblateOrg/weblate/issues/7742#issue-1261056973 I really think the best approach is to translate each string separately. These are really independent strings and should be handled that way in Weblate.

henry-torproject:
> There is nothing in Weblate that would fit this.

How much work would it take to enable this? Would it be a matter of a few tweaks, or is it outside of weblate's current structure?

> Looking at the example posted in [WeblateOrg/weblate#7742 (comment)](https://github.com/WeblateOrg/weblate/issues/7742#issue-1261056973) I really think the best approach is to translate each string separately. These are really independent strings and should be handled that way in Weblate.

When fluent is used in a HTML context the attributes correspond to the HTML attributes. So the strings are related in that they correspond to the same HTML element. E.g.

```
my-button = Open File
  .accesskey = F
  .title = Choose a file from your computer
```

[...]

Issue #4448, opened 2021-10-08
Fluent-backed units are not syntax-checked and may completely break Weblate


flaki:
I've spent the better part of the day debugging a 500 error from Weblate that manifested pretty much out of the blue in random places. As it turns out the root of the issue was a honest mistake from one of the Chinese translators, who translated the string `Note to organizers ({ -optional })` to `给组织者的附注 ({ -可选 })`. Notice the `{ -optional }` bit which is a [term](https://projectfluent.org/fluent/guide/terms.html) in fluent, an internal reference. Should this have happened with a latin language, we might have never noticed, but this caused a syntax error in the parser later on.

The worst issue (that probably belongs to Weblate in the first place) is allowing such invalid syntax to be entered and saved in the first place. The issue was so severe that even discoverability issues aside, the problem could only be fixed by manually editing the PostgreSQL database itself, trying to edit a previously saved invalid sequence like this resulted in 500 error as well. The issue can be reproduced when repo operations are attempted or the "Download translation" button hit. This also means that in the case of linked repositories in other components, the 500 error can even surface in perfectly valid repositories (or, in my case, even in components backed by a different file format, not Fluent).

The reason I'm filing this bug primarily here is because it was `translate/storage` that eventually made this issue harder to debug than it needed to be. This is the error that is reported:

```
File "[..]/site-packages/fluent/syntax/stream.py", line 258, in take_id_start
  raise ParseError('E0004', 'a-zA-Z')

Exception Type: ParseError at /file-sync/content/blogposts/
Exception Value: a-zA-Z
Request information:
USER: Weblate Admin
```

Before the broken ftl file gets handed to `fluent.syntax` there is a serialization step:

```
  File "/home/weblate/weblate-env/lib/python3.8/site-packages/translate/storage/fluent.py", line 197, in serialize
    body = [unit.to_entry() for unit in self.units]
  File "/home/weblate/weblate-env/lib/python3.8/site-packages/translate/storage/fluent.py", line 197, in <listcomp>
    body = [unit.to_entry() for unit in self.units]
  File "/home/weblate/weblate-env/lib/python3.8/site-packages/translate/storage/fluent.py", line 149, in to_entry
    value = fp.maybe_get_pattern(FluentParserStream(self.source))
```

The primary issue debugging the problem is that none of the error message, or the full backtrace, or the extended call chain information shows anything useful to pinpoint *where* did the broken parsing occur. Because the ftl breaks in serialization there's also no temporary file created.

The full backtrace:

[...]

PR #4822, opened 2023-01-27
Refactor Fluent parsing and serializing

henry-torproject:
We refactor the parsing and serializing of FluentFiles with a number of
improvements:

+ Fluent Term's keep a prefix "-" id so they can be better identified as  Terms, rather than Messages.
+ Fluent Terms and Messages expose all of their Attributes within a single FluentUnit in a fluent-ish syntax.
+ Multi-line values are now properly handled. In particular, the indent has been removed from the FluentUnit's source.
+ We no longer rely on methods from fluent.syntax sub-modlues that are not part of its public API (serialize_pattern, serialize_placeable, and FluentParser.maybe_get_pattern).
+ We throw ValueErrors if the FluentFile's FluentUnits do not serialize to valid fluent, with improved messaging (see issue https://github.com/translate/translate/issues/4448).
+ No longer include all fluent Placeables as placeholders since they may contain localized text. In particular, no longer include Placeables for SelectExpressions, FunctionReferences, NumberLiterals and StringLiterals.
+ We escape special characters "*", "[" and "." if they appear at the start of a value, for consistency with multiline values.
+ We include ResourceComments and GroupComments in the comments of FluentUnits under their scope.

We also refactor the test code and add coverage for these features and more.

Fixes https://github.com/translate/translate/issues/4649

Partially addresses https://github.com/translate/translate/issues/4448

### DIFF
--- a/translate/storage/fluent.py
+++ b/translate/storage/fluent.py
@@ -22,163 +22,425 @@ It is a monolingual base class derived format with :class:`FluentFile`
 and :class:`FluentUnit` providing file and unit level access.
 """
 
-from fluent.syntax import FluentParser, ast, parse, serialize, visitor
-from fluent.syntax.serializer import serialize_pattern, serialize_placeable
-from fluent.syntax.stream import FluentParserStream
+import re
+import textwrap
+
+from fluent.syntax import FluentSerializer, ast, parse, serialize, visitor
 
 from translate.storage import base
 
 
-def id_from_source(source):
-    # If the caller does not provide a unit ID, we need to generate one
-    # ourselves. A valid Fluent identifier has the following EBNF grammar:
-    #     Identifier          ::= [a-zA-Z] [a-zA-Z0-9_-]*
-    #
-    # This means we can't simply use the source string itself as the identifier
-    # (as e.g. PO files do). Instead, we hash the source string with a
-    # collision-resistant hash function.
-    import hashlib
-
-    return "gen-" + hashlib.sha256(source.encode()).hexdigest()
-
-
-def source_from_entry(entry):
-    # Serialized patterns come in two forms:
-    # - Single-line patterns, which have a leading space we need to strip (for
-    #   consistency with the expectations of what callers will set).
-    # - Multiline patterns, which have a leading newline we need to preserve.
-    return serialize_pattern(entry.value).lstrip(" ")
-
-
 class FluentUnit(base.TranslationUnit):
-    """A Fluent message."""
+    """Represents a single fluent Message, Term, a ResourceComment, a
+    GroupComment or a stand-alone Comment.
+    """
 
-    def __init__(self, source=None, entry=None):
+    def __init__(
+        self,
+        source=None,
+        unit_id=None,
+        comment="",
+        fluent_type="Message",
+        placeholders=None,
+    ):
+        """
+        :param source: The serialized fluent value, or None.
+        :type source: str or None
+        :param unit_id: An optional id to set (see :meth:`~FluentUnit.setid`),
+            otherwise an id is generated from the given `source`.
+        :type unit_id: str or None
+        :param str comment: An optional comment for the unit.
+        :param str fluent_type: The fluent type this unit corresponds to.
+        :param placeholders: An optional list of sub-strings of the source that
+            should be highlighted as placeholders. A translation of this unit
+            would be expected to contain the same sub-strings.
+        :type placeholders: list[str] or None
+        """
+        if fluent_type not in (
+            "Message",
+            "Term",
+            "ResourceComment",
+            "GroupComment",
+            "DetachedComment",
+        ):
+            raise ValueError(f'Unknown value "{fluent_type}" for fluent_type')
         super().__init__(source)
-        self._type = None
-        self._id = None
-        self._attributes = {}
-        self._placeables = []
-        if source is not None:
-            self._type = ast.Message
-            self._set_value(source)
-            # Default to source string
-            self._id = id_from_source(self.source)
-        if entry is not None:
-            self.parse(entry)
+        self._fluent_type = fluent_type
+        self._is_comment = fluent_type.endswith("Comment")
+        self.placeholders = placeholders or []
+        self.target = source
+        if unit_id is None and fluent_type == "Message" and source:
+            unit_id = self._id_from_source(source)
+        self.setid(unit_id)
+        self.addnote(comment)
 
-    def _set_value(self, value):
-        self.source = value
-        # The source and target are equivalent for monolingual units.
-        self.target = self.source
+    @staticmethod
+    def _id_from_source(source):
+        # If the caller does not provide a unit ID, we need to generate one
+        # ourselves.
+        # The set of valid ids is restricted, so we cannot use the source string
+        # as the identifier (as e.g. PO files do). Instead, we hash the source
+        # string with a collision-resistant hash function.
+        # By default, we choose an id that indicates that this represents a
+        # fluent Message.
+        import hashlib
+
+        return "gen-" + hashlib.sha256(source.encode()).hexdigest()
 
     def getid(self):
         return self._id
 
+    # From fluent EBNF.
+    _FLUENT_ID_PATTERN = r"[a-zA-Z][a-zA-Z0-9_-]*"
+    _FLUENT_ID_REGEXES = {
+        "Message": _FLUENT_ID_PATTERN,
+        "Term": r"-" + _FLUENT_ID_PATTERN,
+    }
+
     def setid(self, value):
-        self._id = value
+        """Set the id of the unit.
+        A valid fluent identifier is [a-zA-Z][a-zA-Z0-9_-]*
+        For a FluentUnit that represents a fluent Message, the id must be a
+        valid fluent identifier.
+        For a fluent Term, the id must be a valid fluent identifier prefixed by
+        a "-".
+        For a fluent Comment, GroupComment or ResourceComment, the id is unused.
+        """
+        regex = self._FLUENT_ID_REGEXES.get(self._fluent_type, "")
+        if not re.fullmatch(regex, value or ""):
+            raise ValueError(
+                f"Invalid id for a {self._fluent_type} FluentUnit: {value}"
+            )
+        self._id = value or None
 
     def isheader(self):
-        return self._type in [
-            ast.Comment,
-            ast.GroupComment,
-            ast.ResourceComment,
-        ]
+        return self._is_comment
 
-    def getattributes(self):
-        return self._attributes
+    @property
+    def fluent_type(self):
+        """The fluent type this unit corresponds to."""
+        return self._fluent_type
 
     def getplaceables(self):
-        return self._placeables
+        """Still called in Weblate. Returns :attr:`~FluentUnit.placeholders`."""
+        return self.placeholders
 
-    def parse(self, entry):
-        this = self
+    @classmethod
+    def new_from_entry(cls, fluent_entry, comment=None):
+        """Create a new unit corresponding to the given fluent AST entry.
 
-        class Parser(visitor.Visitor):
-            _found_id = False
+        :param fluent_entry: A fluent Entry to convert.
+        :type fluent_entry: Entry
+        :param comment: A comment to set on the unit. For fluent Comments the
+            comment is taken from the object instead.
+        :type comment: str or None
 
-            @staticmethod
-            def visit_Attribute(node):
-                this._attributes[node.id.name] = source_from_entry(node)
+        :return: A new FluentUnit.
+        :rtype: FluentUnit
+        """
+        if isinstance(fluent_entry, ast.ResourceComment):
+            return cls(comment=fluent_entry.content, fluent_type="ResourceComment")
+        if isinstance(fluent_entry, ast.GroupComment):
+            return cls(comment=fluent_entry.content, fluent_type="GroupComment")
+        if isinstance(fluent_entry, ast.Comment):
+            return cls(comment=fluent_entry.content, fluent_type="DetachedComment")
+        if isinstance(fluent_entry, ast.Term):
+            return cls._create_from_fluent_pattern(
+                fluent_entry, "Term", f"-{fluent_entry.id.name}", comment
+            )
+        if isinstance(fluent_entry, ast.Message):
+            return cls._create_from_fluent_pattern(
+                fluent_entry, "Message", fluent_entry.id.name, comment
+            )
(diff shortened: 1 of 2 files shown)


henry-torproject:
This is my first contribution, so I'm not sure on all the conventions, or exactly how this integrates with weblate.

I split the merge request into 4 separate commits:

1. Commit 1. Small clean up to removed unused errors.
2. This is the refactoring. This was done in one large step. I needed to change the logic to address https://github.com/translate/translate/issues/4649 but I ended up addressing a lot of other edge cases. I kept the changes to `test_fluent.py` minimal to show that the overall behaviour is mostly the same as before except fluent Message Attributes now become separate FluentUnits. I can try and split this further in retrospect if it would help.
3. Adding tests. I replaced the previous tests with new ones since the previous tests repeated the same pattern. I also added tests for all the edge cases I addressed.
4. Added extra support for GroupComment and ResourceComment so they aren't simply ignored.

Let me know if you want any of these moved into separate merge requests or merged together.

henry-torproject:
@nijel  Since I don't know exactly how Translate integrates with Weblate, could you give me an overall view of how Weblate would use a TranslationStore and TranslationUnit. I want to know how it stores the original locale's localization file input, and how it saves and exports the file for other locales with transaltions, just to make sure that this is going to work as expected with future and existing fluent files that use Weblate. Links to the relevant weblate code is ok.

nijel:
I will get to review this later, but to answer Weblate related questions for now:

Weblate adds a thin layer on top of translate-toolkit to work with files and units, this is all in https://github.com/WeblateOrg/weblate/blob/main/weblate/formats/ttkit.py. It makes it possible to support different backends in Weblate, and it also hides some differences between formats in translate-toolkit.

1. Once the file is parsed, all (translatable) units are stored in the database.
2. Any translations happen in the database only.
3. Once the changes should be written out, the file is again parsed, and matching units get updated (typically by setting `target` attribute).
4. Serialization is called.

So typically, Weblate should be resistant to changes in the translate-toolkit, but as translate-toolkit is not really consistent in the API, there can be always surprises.

Fluent specific code in Weblate right now:

https://github.com/WeblateOrg/weblate/blob/a249d75fcbf8a2e05ef71284022808e14060077f/weblate/formats/ttkit.py#L1998-L2040

henry-torproject:
> Weblate adds a thin layer on top of translate-toolkit to work with files and units, this is all in https://github.com/WeblateOrg/weblate/blob/main/weblate/formats/ttkit.py. It makes it possible to support different backends in Weblate, and it also hides some differences between formats in translate-toolkit.
> 
>1. Once the file is parsed, all (translatable) units are stored in the database.
> 
>2. Any translations happen in the database only.
> 
>3. Once the changes should be written out, the file is again parsed, and matching units get updated (typically by setting `target` attribute).

Thanks for the description. Are they matched based on if they have the same `getid`? I changed the ids for [fluent Terms](https://projectfluent.org/fluent/guide/terms.html) to include a prefix "-" to keep them distinct from fluent Messages. Is this going to cause a problem?

nijel:
Yes , it is used for matching, so Weblate will see those as new units. If there is a way to detect this without looking at the file, we can do a database migration to minimize the impact.

henry-torproject:
> Yes , it is used for matching, so Weblate will see those as new units. If there is a way to detect this without looking at the file, we can do a database migration to minimize the impact.

I'm not sure what you mean by "without looking at the file", what can we work with?

Before the patch, the term "-example-term" would be saved as a `FluentUnit` `unit` with

1. `unit.getid() == "example-term"`.
2. `unit._type == fluent.syntax.ast.Term`.

Where `fluent.syntax.ast` is from the fluent python library. So I guess if you can access `unit._type` then you can prepend a "-" to the id. I split this id change for Terms into a separate commit so you could land just that small change to do the migration.

To give you some context: a fluent Term and a fluent Message have different roles. A Message is like a typical translation. A Term is generally like a name (like "Firefox") that can be used within messages. As such, a Term may have different language properties (like beginning with a vowel, or gender) and may need information about the context it is being used in. As an example, here is the same Term in firefox for [en-US](https://searchfox.org/mozilla-central/rev/861fb9abfcaff123aab45f6ac56a0106b116dc14/browser/branding/official/locales/en-US/brand.ftl#20) and [cs](https://searchfox.org/l10n/rev/456a0d206f9338c30cf028cf2afe58da8bdfe0ef/cs/browser/branding/official/brand.ftl#41-51). So unlike Mesages, Terms are more locale-dependant in their structure. As such, it can be important for a translator to know whether they are translating a Message or a Term. Putting a prefix "-" in the id is one way to identify a Term for someone who is familiar with fluent. It also makes it more obvious that when the user writes `{ -example-term }` then they are referencing the other translation unit.

nijel:
Okay, in that case migration without parsing the file is not really possible and I don't think it's worth of the complexity to parse the files on upgrade. So all term units will be seen as new strings upon this change.

henry-torproject:
> Okay, in that case migration without parsing the file is not really possible, and I don't think its worth of the complexity to parse the files on upgrade. So, all term units will be seen as new strings upon this change.

What information is stored in the database? Is it just the id and source/target values, or is there some extra info?

Would it be possible to duplicate all units with ids, and then clean up the unmatched units that aren't in the template after the change? I.e. for any unit with id `example-id`, we create another with the id `-example-id` with the same source (and target).

I know of one file in tor project's weblate that defines Terms, but we could probably clean it up manually with some coordination. I'm not sure about other projects though.

nijel:
Duplicating the units including all related objects (history, comments, ...) sounds like a lot of work, but it would be the way to go. The issue is that there will be both of them visible until the file is parsed, what is something we can't do directly from the migration.

henry-torproject:
> Duplicating the units including all related objects (history, comments, ...) sounds like a lot of work, but it would be the way to go. The issue is that there will be both of them visible until the file is parsed, what is something we can't do directly from the migration.

ok. I really don't want to mess with other people's projects if possible.

Alternatively, would it be possible to add an additional flag "fluent-unit-is-term" to the unit here https://github.com/WeblateOrg/weblate/blob/693d70041c55d74295fd09e3e8546b0bfde30e11/weblate/formats/ttkit.py#L2006 based on the current `unit._type`, and use that to migrate the id at some later stage. As far as I can tell, any flag text can be set, and this is stored in the database. The main thing I'm unsure of is at what point the flag would be saved to the database, I guess it would require re-saving the unit.

So the steps would be (in different commits, if needed):

1. In weblate, include "fluent-unit-is-term" in the flags if `unit._type.__class__.name == "Term"`.
2. Update the database to save the new flags (this is the step I'm unsure of).
3. Incorporate the changes to Term ids from translate toolkit, with a migration of the old ids based on the presence of the "fluent-unit-is-term" flag.
4. Clean up the flag if we don't need it anymore.

nijel:
When considering the two-step migration, I'd instead modify Weblate to change id to the future format with current translate-toolkit using the logic you suggest for using flags. I'm still not sure whether it is worth of the effort.

henry-torproject:
> When considering the two-step migration, I'd instead modify Weblate to change id to the future format with current translate-toolkit using the logic you suggest for using flags. I'm still not sure whether it is worth of the effort.

I'm not sure how much effort it would be. I also don't know what other projects use fluent, so I cannot know whether they use Terms. But I would like to not silently break other project's translations.

So if a migration is not going to happen, then I think I would keep the ids for Terms as they are now (without the "-" prefix) and find some other way to distinguish between Terms and Messages on weblate. Maybe with something under "Flags".

nijel:
I think it makes sense to use the same identifier in Weblate as is in the file.

In the end, you probably can have following fluent file:

```
-something = Fizz Buzz
something = I can code { -something }!
```

Which cannot be represented with the current code because both end up being called `something`.

As for the migration, we've been there for other formats as well, and it turns out to be tricky. To do it properly, Weblate would need to reparse the original file during the migration. That has several issues that would lead to massive complexity of the migration code, which would be hard to test. I think it's a valid approach to document this in the upgrade notes, and let users deal with that. On Hosted Weblate we currently have only four components using Fluent, so the scope doesn't seem to be massive here.

nijel (approved):
Looks good to me. Will merge this next week unless there is some negative feedback.

henry-torproject:
> Looks good to me. Will merge this next week unless there is some negative feedback.

@nijel nice! I have two questions to double-check with you.

## Q1

For a fluent Message, I currently treat its value as a distinct unit, and each Attribute as a separate unit. When dealing with a translation, we expect *every* part of the Message to be translated. As such, when serializing the translation, I wanted to ensure that the translation includes the value, as well as all attributes for a Message. If one or more is missing, the entire Message should be considered incomplete and no part of it should appear in the serialized ".ftl" file for the translation. E.g. for the message `my-button` with attributes `title` and `accesskey`, we expect a source/target value for the units `my-button`, `my-button.title` and `my-button.accesskey`.

I tried to enforce this in the current patch, but I'm not if it will work with weblate. Currently I check for empty units in `to_fluent` [here](https://github.com/translate/translate/pull/4822/files#diff-da94ddb16cd123953b692cc6dd020564666354fe0bac97622a0f19ecbb37c729R326-R328) and I filter out Messages with empty parts [here](https://github.com/translate/translate/pull/4822/files#diff-da94ddb16cd123953b692cc6dd020564666354fe0bac97622a0f19ecbb37c729R680-R687). But this will only work if non-translated weblate units have empty `source` values *and* are still included in the weblate storage object.

Is this assumption correct in weblate? Otherwise I can add some extra metadata to the `FluentFile` object to track missing attributes more directly.

## Q2

In contrast to fluent Messages, a fluent Term's Attributes are expected to be different for different locales. Therefore, I have allowed the source for Terms to include fluent-ish syntax for attributes. So a translating user on weblate could enter something like

```
My Term
.starts-with-vowel = no
```

where the "." line starts an attribute. The translating user would have to have some fluent knowledge to be able to use this.

In principle, I could treat fluent Messages the same, so a translating user would enter

```
Remove
.title = Remove the selected contact from the list
.accesskey = R
```

and we just add some checks to weblate to ensure that we have lines for each expected attribute.

This would make it more consistent with fluent Terms, but I feel like it could be confusing syntax for translators.

What do you think is better for translators on weblate? Each attribute is in a separate unit (the current patch). Or we have one unit per Message, and the user translates all attributes within that one unit?

henry-torproject:
> Q1: Weblate assumes that the file is a complete representation of the translation. So, if saving discards some translated strings, Weblate will discard it later as well.

ok, I misunderstood then. I thought only the template file was a complete representation, and translations would be derived from it using the database only. But what you are describing would not work.

> Is there a way to store an incomplete translation?

Messages with missing values or attributes can be stored in a fluent file, but that file cannot be used safely in the final product because it will lead to empty parts of the UI or even coding errors because an attribute is missing.

However, I think the firefox build system has some additional checks it performs for missing attributes, so we might be able to clean up the file on our end instead. I'm not sure what the other fluent projects on weblate would want though.

In principle we could come up with a way to comment these out in a recognisable way so they can be read back into weblate, but that seems a bit hacky.

I guess this might be another reason to follow the suggestion in "Q2" above and expose the Message and its Attributes within a single weblate unit. And basically require translators to translate the entire Message in one batch, with an error if an attribute was missing.

Note that it should be possible, using the `source` and `id` alone to migrate from Messages and Attributes being combined in a single unit to them being separate units. So we could always do that in the future.

> Do you know, how is this handled in Pontoon? Maybe it's better to look how they do it before reinventing that ourselves…

I'm not sure what they do in the back end for pontoon, but their approach is quite different because they group the Message and its Attributes as separate inputs, but shown on the same page, so they are inherently linked.

nijel:
> I thought only the template file was a complete representation, and translations would be derived from it using the database only. B

It works this way, but if Weblate saves an unit, it expects it to be saved in the file. Not disappearing because of other (related) unit is not translated. That,s how I understood this:

> Message should be considered incomplete and no part of it should appear in the serialized ".ftl" file for the translation. 


> I'm not sure what they do in the back end for pontoon, but their approach is quite different because they group the Message and its Attributes as separate inputs, but shown on the same page, so they are inherently linked.

Okay, I think they way to go then is to expose these as single unit then. For the future, we might consider adding UI to edit these in Weblate so that it is more user friendly. I guess this will be what Pontoon does.

nijel:
I quickly looked at their source, and they emit entity (what matches unit in terms of translate-toolkit) for each `ast.Message` and `ast.Term`. The text being what `serialize_entry` produces.

In general, their code seems quite simple compared what we have, see https://github.com/mozilla/pontoon/blob/master/pontoon/sync/formats/ftl.py

henry-torproject:
> Okay, I think they way to go then is to expose these as single unit then. For the future, we might consider adding UI to edit these in Weblate so that it is more user friendly. I guess this will be what Pontoon does.

Yes, that eventual UI would be ideal. For now, I'll make that change to the current patch now, and use a single unit.

In the short-term on the weblate side, I was thinking of adding some fluent-specific features:

1. Syntax highlighting for the fluent references, selectors and the ".attribute = " parts. Is there some existing code in weblate for this?
2. Add some fluent-specific verifications, like ensuring there are no missing attributes, missing references, and verifying html-parts within a message. I could expose some more API on the FluentUnits to do this. I noticed there is a "weblate/checks" directory which does verifications, but is it ok to do format-specific checks here?

nijel:
>     Syntax highlighting for the fluent references, selectors and the ".attribute = " parts. Is there some existing code in weblate for this?

Weblate uses https://prismjs.com/ for syntax highlighting. Fluent doesn't seem to be supported there right now. 

>    Add some fluent-specific verifications, like ensuring there are no missing attributes, missing references, and verifying html-parts within a message. I could expose some more API on the FluentUnits to do this. I noticed there is a "weblate/checks" directory which does verifications, but is it ok to do format-specific checks here?

Yes, that is the correct place. 

Flags are used to toggle both syntax highlighting and checks, these can be added via file format.

emmapeel2:
> > Looks good to me. Will merge this next week unless there is some negative feedback.
> 
> Can we delay this being used in _weblate_ until the 20th February or so?
> 
> @emmapeel2 is away right now and is needed to handle the change in id for fluent Terms. And in case there is some unexpected fallout from the other changes.

Bring it on!

henry-torproject:
@nijel I've started writing some fluent checks for weblate, using some more API in translate toolkit. But they can come later.

I think the current changes are ok to land in translate and used in weblate. Can you just give it another quick re-review since I made the Message Attribute changes and a few minor tweaks after the last one https://github.com/translate/translate/pull/4822#pullrequestreview-1288981871

henry-torproject:
> I need to do one more review and I will merge this if I don't encounter any issues. I'm just too busy in the last days, so give me a few more days for that.

ok,  sounds good! :)

nijel (approved):

nijel:

PR #4846, opened 2023-02-26
json: Add support for go-i18n v2

KuSh:
go-i18n v2 supports other data-serialization format (TOML, YAML) with the same structure but I don't know if a mutualization is possible, nor how to do it. But if desirable I'm willing to try with some advices.


nijel:

TOML is currently not supported in translate-toolkit, YAML is, but the format would have to be reimplemented there (the JSON serializer always writes a complete file, while YAML tries to preserve comments).

PR #4858, opened 2023-03-05
 json: Fix go-i18n v2 unit format on plurals without description

KuSh:
While trying to implement go-i18n v2 format handling in weblate I saw there was a bug when unit has plurals without a description
I've also simplified the code by relying as much as possible on parent classes


nijel:

Issue #4879, opened 2023-04-11
JUnit reporting support for pocount


last-partizan:
Hello, i'm adding translation validation for my CI, and looking for options to report untranslated strings.

Your project already have `pocount` for this, but it doesn't include junit output.

I have added this feature internally to my project, would you like to include it in your code? I can make PR with my changes.

<details>
<summary>(It looks like this)</summary>

```diff
diff --git a/.ci/check-missing-translations b/.ci/check-missing-translations
index a420217d7c..3da5e0827e 100755
--- a/.ci/check-missing-translations
+++ b/.ci/check-missing-translations
@@ -47,7 +47,7 @@ state_strings = {
 logger = logging.getLogger(__name__)

 # define style constants
-style_full, style_csv, style_short_strings, style_short_words = range(4)
+style_full, style_csv, style_short_strings, style_short_words, style_junit = range(5)

 # default output style
 default_style = style_full
@@ -157,7 +157,9 @@ def calcstats(filename):
     targetwords = lambda elementlist: sum(
         wordcounts[id(unit)][1] for unit in elementlist
     )
-    stats = {}
+    stats = {
+        "filename": filename,
+    }

     # units
     stats["translated"] = len(translated)
@@ -316,7 +318,7 @@ def summarize(title, stats, style=style_full, indent=8, incomplete_only=False):
                 + ConsoleColor.ENDC(),
             )
         )
-    else:  # style == style_full
+    elif style == style_full:
         print(
             "Processing file : " + ConsoleColor.HEADER() + title + ConsoleColor.ENDC()
         )
@@ -406,6 +408,7 @@ def untranslatedmessages(units):
 class summarizer:
     def __init__(self, filenames, style=default_style, incomplete_only=False):
         self.totals = {}
+        self._stats = []
         self.filecount = 0
         self.longestfilename = 0
         self.style = style
@@ -442,11 +445,13 @@ Review Messages, Review Source Words"""
                 summarize("TOTAL:", self.totals, incomplete_only=False)
             print("File count:   %5d" % (self.filecount))
             print()
+        if self.style == style_junit:
+            self.print_junit()

     def updatetotals(self, stats):
         """Update self.totals with the statistics in stats."""
         for key in stats.keys():
-            if key == "extended":
+            if key in ["extended", "filename"]:
                 # FIXME: calculate extended totals
                 continue
             if key not in self.totals:
@@ -456,6 +461,7 @@ Review Messages, Review Source Words"""
     def handlefile(self, filename):
[...]

PR #4880, opened 2023-04-12
feat(junitmsgfmt): Rewrite in python; add --untranslated option

last-partizan:
I started to add junit output to pocount, but figured out it's better to rewrite existing junitmsgfmt to report untranslated messages.

Default behaviour is the same as the old one, and with `--untranslated` option it will add failures for files with untranslated messages.

I would like to hear some feedback, especially for tests as this is my first PR here.

Refs #4879

### DIFF
--- a/translate/tools/junitmsgfmt.py
+++ b/translate/tools/junitmsgfmt.py
@@ -1,0 +1,99 @@
+from __future__ import annotations
+
+import subprocess
+import xml.etree.ElementTree as etree
+from argparse import ArgumentParser
+from os.path import basename
+from time import time
+from typing import Iterable, NamedTuple
+
+
+class MsgfmtTester:
+    def __init__(self, files: Iterable[str], untranslated=False):
+        self._detect_untranslated = untranslated
+        self._files = files
+
+    def run(self):
+        results = list(map(self._run_msgfmt, self._files))
+        self._print_results(results)
+
+    def _run_msgfmt(self, pofile: str):
+        start_time = time()
+        process = subprocess.Popen(
+            ["msgfmt", "-cv", "-o", "/dev/null", pofile],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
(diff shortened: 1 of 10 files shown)


translate/tools/junitmsgfmt.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import annotations
+
+import subprocess
+import xml.etree.ElementTree as etree
```
last-partizan, line 4: I see you're using lxml instead of xml.etree and deepsource marked this as error.

Should i use it here too? I used it becouse it is typed. And it's safe for generating xml. But i can switch to lxml.

Would be great to add types-lxml as dependency for type-checking support.

nijel: lxml is safer for parsing untrusted XML, that's why it's used in most places. You are using it for generating, so it does not matter much.

nijel (commented):
Can you please also update documentation in `docs/commands/junitmsgfmt.rst`?

last-partizan:
What's the preferred way to add tests?

I see codecov is detecting only python tests, maybe it's better to replace shell tests with pytest tests? But i haven't found existing way to tests expected output.

Can i add pytest-syrupy for snapshot testing? it would be much easier to compare program outputs.

translate/tools/junitmsgfmt.py
```diff
@@ -0,0 +1,99 @@
...
+                classname="msgfmt",
+                name=f"check[{result.file}]",
+                file=result.file,
+                time=f"{result.time:.4f}",
+            )
+            for failure in result.failures:
+                etree.SubElement(
+                    case, "failure", message=failure.message
+                ).text = failure.text
+        etree.indent(root)
```
nijel, line 68: This was added in Python 3.9, so it fails on older versions. Maybe this would be a reason to use lxml instead? :-)

last-partizan: Looks like that's my only option :)

PR #4881, opened 2023-04-12
feat(tests): Add snapshot testing

last-partizan:
Hello, i would like to add snapshot testing for this app.

This should make easier to write and update functional tests from tests/cli.

I'm preparing rewrite of pocount, and current cli tests are incomplete. This ensures output will not change for all option combinations after refactoring.

Snapshots can be generated/updated with `pytest --snapshot-update`.


translate/tools/tests.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+
+from syrupy.assertion import SnapshotAssertion
+
+
+po_files = """
+tests/cli/data/test_pocount_po_csv/one.po
+tests/cli/data/test_pocount_po_file/one.po
+tests/cli/data/test_pocount_po_fuzzy/one.po
+""".split()
```
nschonni, line 12: Shouldn't this be declared as an array, instead of a set of strings that is split into an array?

nschonni:
I'm guessing moving these tests to the existing file https://github.com/translate/translate/blob/61132d19424d97acb630b22fb52d8687c716ff35/translate/tools/test_pocount.py would get rid of some of the errors in the CI. The other error is for overriting the built-in `opts` I think

last-partizan:
If i understand correctly, `test_pocount.po` is for unit-tests, my snapshot tests is functional tests targeted to replace `tests/cli/`.

Would be ok to put my those snapshot tests into `tests/cli/test_pocount.py`?

And probably one other `test_help.py`, to contain all `--help` checks (they're completely identical for all apps, so it we wouldn't need to repeat it for each.

nijel:
Looks nice.

I'm uncertain if it's still reasonable to split the tests once they both use pytest. So I'd place all `pocount` tests to `translate/tools/test_pocount.py`.

There is really no separation of unit and functional tests for the rest of the code. Convertor tests (`translate/convert/test_*.py`) typically also include both unit and functional tests.

last-partizan:
Btw, should i remove old cli tests for pocount and junitmsgfmt?

last-partizan:
This turned out trickier then i expected.

When adding "--help" snapshots for all tools, i can see that output is different for different python versions.

Should we remove them? I think that's safe option - testing ArgumentParser should be left to it's own testsuite.

nijel:
How does it differ between Python versions? Is optparse output different or something we can influence?

last-partizan:
For example, python 3.7 vs 3.10 - different names for "optional arguments". Between other versions indents or text aligning may be different.

```
>       assert stdout == snapshot
E       AssertionError: assert [+ received] == [- snapshot]
E           '''
E           ...
E
E         - optional arguments:
E         + options:
E             -h, --help            show this help message and exit
E            ...
E           '''
```

nijel:
Hmm, that really doesn't look suitable for snapshot testing.

last-partizan:
I was thinking how it was working before, because i removed "snapshot" for old-style bash tests.

But turns out it was not checking snapshots for `--help` tests, it was only checking that it "has_stdout".

```
python $PYTHON_ARGS $(which pocount) -h
start_checks
has_stdout
end_checks
```

If you think this type of test may be useful, i can add somethink like `stdout != ""`. But i don't think it makes much  sense.

nschonni:
Maybe https://pypi.org/project/pytest-snapshot/ would work with adding a python version to the file path `snapshot.assert_match(foo('function input'), 'help_output_37.txt')`

last-partizan:
@nschonni it could work even with pytest-syrupy, but we still need to generate those snapshots. And i don't expect every developer would have all python versions installed.

And leaving snapshot generation to autocommit bot would defeat the purpose of having them at all (becouse the goal is to have only expected changes, and not update it automatically).

last-partizan:
One possibility that i can imagine - is restricting those `--help` snapshots to latest version of python.

It would definitely work, but we will require for devs to use latest version of python. That's easier that all of them, but still a limiting factor.

I suggest to not use snapshots for `--help` at all. There is no way we can break argparser default functionality, so we should not test it.

We should be testing only our code - and we're already doing it.

nijel:
Potentially, we could break optparse to not emit help, so testing that help works and contains some commands makes sense (I believe we currently have such tests). Testing full output is not really desired here as it a moving target.

last-partizan:
Added check that stdout from `commandname --help` contains at leas command name and "--help".

nijel:

PR #4882, opened 2023-04-12
Refactor pocount

last-partizan:
This splits code into separated stats collection and rendering parse, with each renderer in it's own class.

It should not change any of the outputs, but let's not merge this until #4881 is merged, to be completely safe.

### DIFF
--- a/translate/tools/pocount.py
+++ b/translate/tools/pocount.py
@@ -1,3 +1,4 @@
+# pyright: strictDictionaryInference=true
 #
 # Copyright 2003-2009 Zuza Software Foundation
 #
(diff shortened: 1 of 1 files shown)


last-partizan:
This is ready for review now.

Rebased new changes, snapshots of output from the old code and from the new code are identical.

last-partizan:
Next step is to replace hand-written csv with `csv` module, but that will include some changes to the output. So it's better to leave that to another PR.

last-partizan:
@nijel @nschonni Please take a look here.

nijel:

PR #4887, opened 2023-04-19
Use black profile for isort

nijel:
This triggers slight changes in the formatting, but makes the code more consistent with other black using projects.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,10 +1,9 @@
+test_patterns = ["**/test_*.py"]
 version = 1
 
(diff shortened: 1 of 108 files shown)


nschonni:
Might want to check out https://beta.ruff.rs/docs/. I supports isort and a few other tools that have been rewritten in rust

nijel:
I've already switched to it for Weblate, where we were pushed to change by the slowness of flake8 on our code base and plugin compatibility issues. None of that happens here, so it's not a high priority for me…

PR #4891, opened 2023-04-21
aresource: fix single whitespace

### LINKED ISSUE #4884: Single whitespace is not preserved while enclosed in double quotes on Android
According to the [Android documentation](https://developer.android.com/guide/topics/resources/string-resource#escaping_quotes):

> By default Android will collapse sequences of whitespace characters into a single space. You can avoid this by enclosing the relevant part of your string in double quotes. In this case all whitespace characters (including new lines) will get preserved within the quoted region. Double quotes will allow you to use regular single unescaped quotes as well.

Which allows for the following use cases:
image
are displayed as
image
(note: I don't know why the first escaped tab is bigger and the second is almost non-existant, tabs on textviews are weird, but at least it can be seen that they are different to the unescaped one, which is converted to a space)

As the screenshots show, inside a double quote part, all whitespace are preserved. When using weblate (which apparently uses this) all previous examples are converted to a single space.

I suspect the issue is on this file: https://github.com/translate/translate/blob/fa62e23bfee56f011af34052fb17fecfc1df2be0/translate/storage/aresource.py#L118-L128

which sould be enclosed in a `if not active_quote or c is EOF:`
or more specifically, this whole else case can be removed and included in the previous one.

https://github.com/translate/translate/blob/fa62e23bfee56f011af34052fb17fecfc1df2be0/translate/storage/aresource.py#L100-L117

replacing it with `elif space_count >= 1:`. Note that this other part manages multiple whitespaces, and do keep linebreaks in that case.

Note: I can provide a PR with the mentioned fix if required.

TrianguloY:
### Description
PR to fix single whitespace not preserved (#4884)

Includes three new tests to check single whitespace in no-quote, single-quote and full-quote.

### Explanation
A single space on android is not treated differently than multispace, the logic to replace/collapse multiple strings should be the same, as opposed to the source code comment.

### Extra modification
About the modified test (test_parse_quoted_newlines)

The input string on android:
image
when displayed on a textview:
image
and fetched
image
returns
image

Which makes sense, because the '\n' are always converted to linebreaks, and the actual linebreaks are only converter while inside a quoted part, which we are.

Fixes #4884

### DIFF
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -99,8 +99,8 @@ class AndroidResourceUnit(base.TranslationUnit):
                 # Handle whitespace collapsing
                 if c is not EOF and c in WHITESPACE:
                     space_count += 1
-                elif space_count > 1:
-                    # Remove duplicate whitespace; Pay attention: We
+                elif space_count >= 1:
+                    # Remove sequential whitespace; Pay attention: We
                     # don't do this if we are currently inside a quote,
                     # except for one special case: If we have unbalanced
                     # quotes, e.g. we reach eof while a quote is still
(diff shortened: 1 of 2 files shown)


nijel:

Apparently, I've added the newline handling intentionally this way in https://github.com/translate/translate/pull/4080. Unfortunately, I didn't reference reasoning there, so I was most likely wrong here.

TrianguloY:
You're welcomed! And thank you for the merge.

Resource parsing on Android is weird, I noticed this bug because I like to use quotes on my apps for multiline strings, and Weblate wasn't showing them (and I was 100% sure it should). A review of what could be the problem pointed me here, and you know the rest.
Now to wait until Weblate updates this dependency so the bug is finally gone!

Btw: while testing this change I noticed that unterminated quotes do keep whitespace, as opposed to that file comment: 
```xml
<string>"a
</string>```
This string is parsed as `a\n`, note that the line break is not converted to a space, even if the quotes weren't closed

Unterminated quoted strings are probably undefined behaviour, maybe they changed it on an older android version, I don't know. I'll try to investigate a bit to see if I can find documentation about it, if so I'll maybe make another PR for that fix.

PR #4910, opened 2023-05-12
refactor(pocount): Use csv module for writing csv

last-partizan:
This refactors csv rendering from hand-written style to using csv module.

Output is changed slightly, but this is expected.

### DIFF
--- a/translate/tools/pocount.py
+++ b/translate/tools/pocount.py
@@ -268,10 +268,7 @@ class CsvRenderer(Renderer):
     def entry(self, title, stats):
         data = stats.copy()
         data.update(title=title)
-        row = {
-            v: data[k]
-            for k, v in self._fields.items()
-        }
+        row = {v: data[k] for k, v in self._fields.items()}
         self._writer.writerow(row)
 
     def footer(self):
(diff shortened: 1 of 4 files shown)


translate/tools/pocount.py
```diff
@@ -240,35 +242,38 @@ def footer(self):
...
+            "fuzzysourcewords": "Fuzzy Source Words",
+            "untranslated": "Untranslated Messages",
+            "untranslatedsourcewords": "Untranslated Source Words",
+            "total": "Total Message",
+            "totalsourcewords": "Total Source Words",
+            "review": "Review Messages",
+            "reviewsourcewords": "Review Source Words",
+        }
+        self._file = io.StringIO()
+        self._writer = csv.DictWriter(self._file, self._fields.values())
```
nijel, line 263: Wouldn't be better to use `sys.stdout` instead of buffering content in `StringIO`?

last-partizan: Nice idea, updated code.

nijel:

last-partizan:
As the next step i would like to rewrite old shell-based tests for pocount to pytest/snapshot-based tests, but this would require dropping python 3.7.

Do you plan to drop it anytime soon?

nijel:
I don't mind dropping support for Python 3.7 (Weblate will need Python 3.8 from the next release), but please create a separate pull request for that so that other maintainers can express as well.

PR #4916, opened 2023-05-16
feat: Drop python 3.7

last-partizan:
Python 3.7 security support ends in 1 month and 1 week, and snapshot testing (syrupy) requires at least 3.8 to work, let's drop it from CI, and then start using snapshot testing instead of the old shell-based tests.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -2,4 +2,4 @@
 pytest==7.3.1
 pytest-cov==4.0.0
 pytest-xdist==3.3.0
-syrupy==4.0.2; python_version >= "3.8"
+syrupy==4.0.2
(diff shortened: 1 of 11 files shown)


nijel (approved):

nijel:
Can you please also adjust pyupgrade configuration in `.pre-commit-config.yaml`?

requirements/test.txt
```diff
@@ -1,5 +1,6 @@
 -r optional.txt
 pytest==7.3.1
 pytest-cov==4.1.0
+pytest-xdist==3.3.0
```
nijel, line 4:
```suggestion
```

nijel:
Merged, as I didn't see any objections to this. Thanks.

PR #4925, opened 2023-05-27
refactor(tests): Refactor pocount tests. Use snapshots instead of bash.

last-partizan:
Now we can replace old bash-based tests with snapshot-based tests.


translate/tools/test_pocount.py
```diff
@@ -169,14 +179,61 @@ def test_totalsourcewords(self):
...
+)
+def test_error_cases(opts, expected):
+    # We're using special case for this, to produce correct output.
+    # Also, using partial matching instead of snapshots, becouse mac-os argparse
+    # output is slightly different.
+    result = subprocess.run(
+        [sys.executable, pocount.__file__, *opts],
+        capture_output=True,
+        text=True,
+    )
```
nijel, line 237: The code executed this way will not be accounted in coverage. The existing functional tests do that via the `PYTHON_ARGS` environment passed to the python interpreter.

nijel: Ah nice. I didn't look at the coverage report, only at the code. I didn't consider the option that pytest might be handling that automatically.

nijel:

PR #4948, opened 2023-06-23
added function for getting context-groups by attribute name and value

MartinTopfstedtSinc:
Hi,

I added a function to collect all context-group tags with a specific attribute name and value.
This will be helpful to retrieve the source file and location context tags from a translation unit.

Related weblate issue : https://github.com/WeblateOrg/weblate/issues/7767.


nijel:
Thanks for the patch. The standard way to expose locations would be via `getlocations`, but this is already used for something else in the `xliffunit`. On the other side, this does not seem to be used anywhere in the translate-toolkit, and there are no tests for the behavior. 

In the end, I think that using the standard `getlocations` API to get locations is a better approach than requiring consumers for this specific format to use something else. But I'd like to hear opinions from other maintainers as well.

My proposal would be to have `getcontextgroupsbyattribute` as the low-level interface and then build `getlocations` on that (using code you've drafted in https://github.com/WeblateOrg/weblate/issues/7767#issuecomment-1604202776).

MartinTopfstedtSinc:
Hi,

i can change this, but wouldn't that be a breaking change? I don't know if someone is using this function as implemented.

nijel:
It is a potentially breaking change, but I haven't found any consumers of this API. I'd go for the change as this makes the API consistent across storages.

MartinTopfstedtSinc:
Ok, waht about `addlocation`?

MartinTopfstedtSinc:
Ok, i added `addlocation `as well, wasn't that complicated.

nijel:
Looks good, the test failure should be addressed by https://github.com/translate/translate/pull/4965 (apparently I didn't notice this usage).

nijel:
Can you please either run pre-commit locally, or enable edits of this PR so that pre-commit.ci can do this for you?

nijel:

PR #4963, opened 2023-07-11
Fluent: add some API for use in weblate

henry-torproject:
Follows from https://github.com/translate/translate/pull/4822

We do some general fixing up of the fluent code, then we add two methods to the FluentUnit that can be used in weblate:

- get_syntax_error: To extract a syntax error in the unit's source without needing to serialize the FluentFile.
- get_parts: To extract all the individual fluent-relevant parts of the unit's source, including its attributes and its variants. This will help weblate split a single source with multiple attributes and [select expressions](https://projectfluent.org/fluent/guide/selectors.html) into simple individual strings, which can then be checked.

### DIFF
--- a/translate/storage/fluent.py
+++ b/translate/storage/fluent.py
@@ -22,8 +22,11 @@ It is a monolingual base class derived format with :class:`FluentFile`
 and :class:`FluentUnit` providing file and unit level access.
 """
 
+from __future__ import annotations
+
 import re
 import textwrap
+from typing import BinaryIO
 
 from fluent.syntax import FluentSerializer, ast, parse, serialize, visitor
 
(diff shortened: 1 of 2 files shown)


henry-torproject:
This is used in https://github.com/WeblateOrg/weblate/pull/9540

In principle, the `get_parts` method and related classes could have been in weblate, but I kept them in translate toolkit for the following reasons:

1. The methods some code with the `FluentUnit` class.
2. It means that weblate doesn't have to work with `python-fluent` directly, but instead has only the relevant features of its structure exposed.

henry-torproject:
FYI: I'm going to be away next week and the week after (17th - 30th July), so I won't able able to respond to questions in that period until I'm back.

nijel:

Issue #3956, opened 2020-03-25
Markdown as a native source format


eighthave:
Markdown is a widely popular markup format for documentation and websites.  If translation tools can directly support it, this opens up a lot of possibilties for a really smooth user experience for both translators and developers/webmasters.  Some platforms like [Transifex](https://docs.transifex.com/formats/markdown) have naive support for Markdown where it breaks down by paragraph more or less.  This is a very limited approach.

For f-droid.org, we have been using [po4a](https://github.com/mquinson/po4a) which is a perl program that parses Markdown into _gettext_ PO files.  That has proven that parsing Markdown for translation can work.  But its approach is limited because it is basically a collection of regexps, and ignores large parts of the Markdown spec.  Here are some specific examples of things that are handled:
* https://github.com/mquinson/po4a/issues/195
* https://github.com/mquinson/po4a/issues/194

You can see that in action here: https://hosted.weblate.org/projects/f-droid/website-docs/

@uniqx and I have discussed some ideas with @nijel about how to handle Markdown well in the translation workflow.  It seems clear that it will be worthwhile to use a complete Markdown parser so that the translation workflow can usefully break the documents into strings to be translated, and back.  Then this will also enable syntax checkers directly in the user interface, and things like checking whether the links are working in the translation.

It seems to me the ideal would be having _docs/MyDoc.md_ be the source file, and the _docs/<LANG>/MyDoc.md_ be the translation file.  For projects that want to translate the document title, then the filename could be a key, e.g. _docs/my-doc.md_ and the title would be in the document:

_docs/my-docs.md_
```markdown

# My Document

```
_docs/de/my-docs.md_
```markdown

# Mein Dokument

```

There is also a widely used extension to Markdown known as [YAML Front Matter](https://jekyllrb.com/docs/front-matter/).  This should also be directly supported, and looks easy to do.  I've outlined how that can be supported here https://github.com/mquinson/po4a/issues/196

nijel:
There is existing [text storage](https://github.com/translate/translate/blob/master/translate/storage/txt.py), but it's probably better to start from scratch for Markdown as it is quite simple.

eighthave:
Seems to me that it will pay off in the long run to have a proper
parsing setup that fully understands Markdown.  That will handle the
vast amount of edge cases that will likely happen here, when you
consider things like inline HTML, ## vs --------- titles, tables, and more.

Seems to me that full parsing will allow for some of the nice gettext
[...]

PR #4966, opened 2023-07-13
Support for Markdown (#3956)

anderskaplan:
Added support for Markdown/CommonMark as a storage format and matching converter scripts `md2po` and `po2md`, similar to other formats. Please see the documentation for the intended use. It's very similar to the HTML format.

This adds an optional dependency in the form of the mistletoe parser.

Note: the converter scripts are by no means pretty. They were adapted from the html converters with minimal changes. To clean up the converters is a large-ish effort and is probably best left to a separate PR.

Note: The generated translation units don't have any line number information. This is because mistletoe does not have that feature yet (PR pending).


anderskaplan:
I'm going to add some basic unit tests for the converters, but don't have the time for it now.

nijel:
Thanks, I will review this once the tests are ready. As for converters cleanup, this is described in https://github.com/translate/translate/issues/3573.

docs/commands/md2po.rst
```diff
@@ -0,0 +1,130 @@
+
+.. _md2po:
+.. _po2md:
+
+md2po, po2md
+****************
```
nijel, line 6:
```suggestion
md2po, po2md
************
```

translate/storage/markdown.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+class markdownfile(base.TranslationStore):
+    UnitClass = mdunit
+
+    def __init__(self, inputfile=None, callback=None, max_line_length=None):
+        base.TranslationStore.__init__(self)
+        self.filename = getattr(inputfile, "name", None)
+        if callback is None:
+            self.callback = self._simple_callback
```
nijel, line 74: This code path should be covered by tests if it is needed.

translate/storage/markdown.py
```diff
@@ -0,0 +1,462 @@
...
+        self, token: span_token.SpanToken, target: str
+    ) -> Iterable[Fragment]:
+        if self.bypass:
+            yield from super().render_link_or_image(token, target)
+            return
+
+        yield from self.embed_span(Fragment("["), token.children, Fragment("]"))
+
+        if token.dest_type in ["uri", "angle_uri"]:
+            # "[" description "](" dest_part [" " title] ")"
```
nijel, line 179: The comments in this method are not really explaining. Either remove them or clarify what they describe.

translate/storage/markdown.py
```diff
@@ -0,0 +1,462 @@
...
+        content, into separate lists. Placeholders marked as important are kept
+        with the main content.
+        """
+        fragments = list(fragments)
+
+        l = 0
+        while l < len(fragments):
+            if getattr(fragments[l], "placeholder_content", None):
+                if getattr(fragments[l], "important", False):
+                    break
```
nijel, line 409: Can we have tests for this, please?

translate/storage/markdown.py
```diff
@@ -0,0 +1,462 @@
...
+    Fragment,
+    LinkReferenceDefinition,
+    LinkReferenceDefinitionBlock,
+    MarkdownRenderer,
+)
+
+from translate.storage import base
+
+
+class mdunit(base.TranslationUnit):
```
nijel, line 53: Please consider using consistent naming with markdownfile. Using MarkdownUnit / MarkdownFile might be even better. Unfortunately, current storage classes are inconsistent, but I think it is better to stick with PEP-8 for new code.

```suggestion
class MarkdownUnit(base.TranslationUnit):
```

translate/storage/markdown.py
```diff
@@ -0,0 +1,462 @@
...
+        self.locations = []
+
+    def addlocation(self, location):
+        self.locations.append(location)
+
+    def getlocations(self):
+        return self.locations
+
+
+class markdownfile(base.TranslationStore):
```
nijel, line 67:
```suggestion
class MarkdownFile(base.TranslationStore):
```

translate/storage/markdown.py
anderskaplan: Making the callback a mandatory parameter. The optional fallback is a remnant from the HTMLFile and I don't see much value in it.

translate/storage/markdown.py
anderskaplan: This function is tested by the test cases `test_remove_placeholders_from_both_ends_of_translation_units`, `test_paragraph_with_only_whitespace_and_placeholders`, `test_plain_image`, and `test_plain_link`.

Would you still like some more tests targeting this particular function directly?

translate/storage/markdown.py
nijel: Yes, that's enough.

translate/storage/markdown.py
nijel: The break statement is never reached while running tests. So it seems that important placeholders are not tested.

anderskaplan: Right. This is because the only placeholders considered "important" at the moment are links, and they are always on the trailing side. I'd rather not break the leading/trailing symmetry, though, so I'll add a unit test for this.

anderskaplan:
There: the converter unit tests have been added and the review comments addressed.

nijel:

Issue #4980, opened 2023-07-30
Support for Android resource converter


jordimas:
Hello. 
I'm currently using https://github.com/miracle2k/android2po to convert from Android resources (bilingual) to PO files. However, the project is no longer maintained. 

My understanding is that aresource.py provides the storage support for implementing a Android Resources to PO but this converter is not currently implemented. Is this correct?

nijel:
Yes, implementing android2po on top of current storages should be pretty easy.

jordimas:
Thanks, I will give a try. Is android2po something that you will be willing to consider as contribution to the project? 
(I do not know if you still accept new conversors)

nijel:
Yes, contributions are welcome.

nijel:
Fixed by https://github.com/translate/translate/pull/4983

PR #4983, opened 2023-08-01
Android2PO conversor - Converts from Android resources to PO file

jordimas:
See #4980

### DIFF
--- a/translate/convert/android2po.py
+++ b/translate/convert/android2po.py
@@ -35,9 +35,9 @@ class android2po:
     def convert_store(self, input_store, duplicatestyle="msgctxt"):
         output_store = po.pofile()
         output_header = output_store.header()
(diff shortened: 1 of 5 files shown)


translate/convert/android2po.py
```diff
@@ -0,0 +1,150 @@
...
+    pot=False,
+    duplicatestyle="msgctxt",
+):
+    """Reads in *input_file* using aresource, converts using :class:`android2po`,
+    writes to *output_file*.
+    """
+
+    output_store = _convertandroid(input_file, template_file, pot, duplicatestyle)
+    output_store.serialize(output_file)
+    return 1
```
nijel, line 132: Can you please extend the tests to cover this function as well?

jordimas: Sure. I did commit the additional UT.

jordimas:
Thanks for your help integrating this.

PR #4987, opened 2023-08-04
fix(ts2po): do not crash and skip unit if the context is empty

pereorga:
`ts2po` crashes if the first context name is empty. For example, in these files:

- https://raw.githubusercontent.com/MythTV/mythtv/master/mythtv/i18n/mythfrontend_ca.ts
- https://raw.githubusercontent.com/MythTV/mythtv/master/mythtv/i18n/mythfrontend_es.ts

```
> ts2po mythfrontend_ca.ts
processing 1 files...
ts2po.py: WARNING: Error processing: input mythfrontend_ca.ts, output None, template None: cannot access local variable 'messagenum' where it is not associated with a value
[###########################################] 100%
```

The PR skips the unit if the context is empty, rather than crash. Another approach could be setting a default context and initializing `messagenum = 0`, but I understand the context should not be empty in normal circumstances (as it is automatically generated)

### DIFF
--- a/translate/convert/test_ts2po.py
+++ b/translate/convert/test_ts2po.py
@@ -142,6 +142,28 @@ new line</translation>
             == "Appears in the Help menu (on Windows and Linux) or the app menu (on macOS)."
         )
 
+    def test_emptycontext(self):
+        """test conversion with a leading empty context"""
+        tssource = """<!DOCTYPE TS><TS>
+<context>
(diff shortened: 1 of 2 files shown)


nijel:
Can you please add a test case for this?

pereorga:
@nijel done

nijel:

PR #4988, opened 2023-08-05
getlocations should return a list instead of string

jordimas:
Hello. A bit of background on why I'm proposing this change.

In commit 9061cac1db35a98f2196c14546a456b98ce0b15f, the method _xliffunit.getlocations_ changed its return type from a Python list to a string.

Consider that xliffunit._getlocations_ inherits from class lisa.LISAunit which inherits from class base.TranslationUnit in which _getlocations_ is defined as:

```
    def getlocations():
        ...
        return []
```

3 reasons why I'm proposing this change: consistency with base class, consistency with the rest of inhered classes which use a list and for consistency in how we expose the API.

### DIFF
--- a/translate/storage/test_xliff.py
+++ b/translate/storage/test_xliff.py
@@ -748,7 +748,7 @@ class TestXLIFFfile(test_base.TestTranslationStore):
         xlifffile = xliff.xlifffile.parsestring(xlfsource)
         xlifffile.units[0].addlocation("some-directory/on-some-test/test.file2:333")
         locations = xlifffile.units[0].getlocations()
-        locations = ", ".join(x for x in locations if x is not None)
+        locations = ", ".join(locations)
         assert (
             locations
             == "some-directory/on-some-test/test.file:222, some-directory/on-some-test/test.file2:333"
--- a/translate/storage/test_xliff.py
+++ b/translate/storage/test_xliff.py
@@ -721,6 +721,7 @@ class TestXLIFFfile(test_base.TestTranslationStore):
 """
         xlifffile = xliff.xlifffile.parsestring(xlfsource)
         locations = xlifffile.units[0].getlocations()
+        locations = ", ".join(x for x in locations if x is not None)
 
         assert locations == "some-directory/on-some-test/test.file:222"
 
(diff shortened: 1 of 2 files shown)


translate/storage/xliff.py
```diff
@@ -526,7 +526,7 @@ def getlocations(self):
                 if sourceFile is not None and lineNumber is not None:
                     locations.append(sourceFile[1] + ":" + lineNumber[1])
 
-        return ", ".join(x for x in locations if x is not None)
+        return locations
```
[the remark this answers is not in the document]
jordimas: Hello. 
locations is set to [] at the beginning of the function, then it will return an empty array if no locations are found in the for loop.

Can you elaborate a bit more please on the _None_ scenario? Thanks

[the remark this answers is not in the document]
nijel: I mean that `None` should not be returned as a location. You add filtering of this in the test, but I think it should be here instead:

```suggestion
        return [x for x in locations if x is not None]
```

Of if there is no `None` in the list, the filtering should be removed from the testcase.

translate/storage/test_xliff.py
```diff
@@ -746,9 +747,8 @@ def test_addlocation():
 """
         xlifffile = xliff.xlifffile.parsestring(xlfsource)
         xlifffile.units[0].addlocation("some-directory/on-some-test/test.file2:333")
-
         locations = xlifffile.units[0].getlocations()
-
+        locations = ", ".join(x for x in locations if x is not None)
```
nijel, line 751: Remove `None` filtering here:

```suggestion
        locations = ", ".join(locations)
```

translate/storage/test_xliff.py
```diff
@@ -721,6 +721,7 @@ def test_getlocations():
 """
         xlifffile = xliff.xlifffile.parsestring(xlfsource)
         locations = xlifffile.units[0].getlocations()
+        locations = ", ".join(x for x in locations if x is not None)
```
nijel, line 724:
```suggestion
        locations = ", ".join(locations)
```

translate/storage/xliff.py
```diff
@@ -526,7 +526,7 @@ def getlocations(self):
                 if sourceFile is not None and lineNumber is not None:
                     locations.append(sourceFile[1] + ":" + lineNumber[1])
 
-        return ", ".join(x for x in locations if x is not None)
+        return [x for x in locations if x is not None]
```
nijel, line 529: I don't see where `None` would appear here. Seems you've just blindly copied this workaround from Weblate (where it is also no longer needed):

```suggestion
        return locations
```

jordimas:
Thanks for the suggestions. I do not see any pending actions for me here. Let me know otherwise. Thanks

nijel:

Issue #4992, opened 2023-08-15
prop2po: personality given in command line is ignored for .strings files


jordimas:
input file: https://github.com/vector-im/element-ios/blob/develop/Riot/Assets/ca.lproj/Localizable.strings
template file: https://github.com/vector-im/element-ios/blob/develop/Riot/Assets/en.lproj/Localizable.strings

When executing:

`prop2po  -i ca.lproj/Localizable.strings -t en.lproj/Localizable.strings -o ca.po `

it gives the following error:

```
root: WARNING: trying to parse ./element-ios/Riot/Assets/ca.lproj/Localizable.strings with encoding: utf-16 but detected encoding is utf-8 (confidence: 1.0)
root: WARNING: trying to parse ./element-ios/Riot/Assets/en.lproj/Localizable.strings with encoding: utf-16 but detected encoding is utf-8 (confidence: 1.0)
prop2po: WARNING: Error processing: input ./element-ios/Riot/Assets/ca.lproj/Localizable.strings, output ca.po, template ./element-ios/Riot/Assets/en.lproj/Localizable.strings: Cannot detect encoding for ./element-ios/Riot/Assets/en.lproj/Localizable.strings.
```

and produces an output of 0 bytes.

It's an encoding problem since the file is in utf-8. Then I do:

`prop2po  --personality strings-utf8  -i ca.lproj/Localizable.strings -t en.lproj/Localizable.strings -o ca.po `

and I get the same error. And the utf8 encoding associated to  _--personality strings-utf8_ is not used.

The problem is in this line of code:
https://github.com/translate/translate/blob/master/translate/convert/prop2po.py#L417

Basically enforces always _strings_ personally ignoring what has been passed (https://github.com/translate/translate/blob/master/translate/convert/prop2po.py#L407) coming from the command line.

As result, all files ending in _strings_ are always processed by _string_ personality ignoring the command line.

jordimas:
This seems to indicate that is personality is only for property files:
https://github.com/translate/translate/blob/master/translate/convert/prop2po.py#L488

And this is the way that the hardcode the personality for strings:
https://github.com/translate/translate/blob/master/translate/convert/prop2po.py#L417

If your remove this, then the default personality became java which makes sense for properties files but not for strings files. 

In my view the ideal will be to:
* Be able to express java as default for property files and then strings for strings files
* Allow also to modify the personality for strings files

jordimas:
Not worth fixing it due to complexity.

PR #4993, opened 2023-08-15
Fix bug which prevented to pass personality in strings type files

jordimas:
This a proposed fix for https://github.com/translate/translate/issues/4992 

If you remove the code changed, the unit test fails since it does not behave as expected.

### DIFF
--- a/translate/convert/prop2po.py
+++ b/translate/convert/prop2po.py
@@ -414,7 +414,7 @@ def convertstrings(
         inputfile,
         outputfile,
         templatefile,
-        personality,
+        personality=personality,
         pot=pot,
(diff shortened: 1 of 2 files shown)


translate/convert/prop2po.py
```diff
@@ -414,7 +414,7 @@ def convertstrings(
         inputfile,
         outputfile,
         templatefile,
-        personality="strings",
+        personality,
```
nijel, line 417:
```suggestion
        personality=personality,
```

jordimas:
This PR introduces a regression. I need to work more on this and I will re-open it again.

PR #4996, opened 2023-08-17
Convert block-level language statements in rc files

J5lx:
Currently, the rc file converters convert top-level language statements like this:

```
LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US

STRINGTABLE
BEGIN
    // ...
END
```

but not block-level ones like this:

```
STRINGTABLE
LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
BEGIN
    // ...
END
```

This PR adds basic support for block-level language statements so that they are converted in the same way as top-level ones are. There’s still no support for more complex scenarios though, e.g. multiple different languages in the same file.

This is my first time working with pyparsing. If there’s a better way to handle this, please let me know.

### DIFF
--- a/translate/convert/po2rc.py
+++ b/translate/convert/po2rc.py
@@ -64,32 +64,37 @@ class rerc:
             comment = comment[:-1]
         yield comment
 
+    def convert_caption(self, s, loc, toks, name=None):
+        yield "CAPTION "
+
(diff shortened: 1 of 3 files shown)


nijel:

PR #5011, opened 2023-08-31
commands can have arg --input several times

### LINKED ISSUE #4979: support several input parameters for pofilter (and other commands)
I want to run pofilter only for some po files and not all in the directory structure (like a/xx.po d/xx.po). But only one `--input` parameter is considered. (I can give many `--exclude` parameters which are all considered, but this gets too complicated with many files.)
So I suggest to support several input parameters for pofilter (and other commands).

#### How

To change this, `inputoption` must have `action="append"` in 
https://github.com/translate/translate/blob/65b70939937dcb3d10eb9791d61b884ad8aad8c8/translate/misc/optrecurse.py#L319
If it is not supposed to be for all commands, it must be changed in another place. (For pofilter only, adapt it in `translate/filter/pofilter.py` by adding method `setformats()` calling the super method and then adapting the option.)

#### More info

The code supports array for input:
https://github.com/translate/translate/blob/65b70939937dcb3d10eb9791d61b884ad8aad8c8/translate/misc/optrecurse.py#L435

I monkey-patched `pofilter` and it runs as expected.
- the input files or directories are used
- to create the output directories, the common ancestor directory of the input is used as base

Other commands could profit as well from this idea. (All those accepting a directory as input.) I did not check all commands, but converters (checked `po2tiki` and `po2xliff`) do, `pocompile` and `poclean` do, ...
I do not get if `pomerge` does, I do not understand the message.

I am happy to send a patch, but I would like to know for where because I do not have experience with most of the commands.

SimonHeimberg:
fixes #4979

This is helpful to run pofilter or convert for several directories or several files (and there are not the only ones in the directory).

### DIFF
--- a/translate/misc/optrecurse.py
+++ b/translate/misc/optrecurse.py
@@ -328,6 +328,7 @@ class RecursiveOptionParser(optparse.OptionParser):
             "-i",
             "--input",
             dest="input",
+            action="append",
             default=None,
             metavar="INPUT",
             help=f"read from INPUT in {inputformathelp}",


nijel:
Have you tested that all commands using this will work correctly with multiple inputs?

SimonHeimberg:
> Have you tested that all commands using this will work correctly with multiple inputs?

I have tested several commands, but not each one.

nijel:

PR #5021, opened 2023-09-07
Add GitHub sponsors configuration

nijel:
Nearly all maintenance on translate-toolkit is currently done by me, so I think it is reasonable to add sponsorship links to Weblate.


claudep (approved):
I don't know if my opinion really counts, as I'm not active since a long time, but I'm fine with this proposal.

PR #5171, opened 2024-01-22
Change resx XMLIdent to respect Visual Studio formatting

michalciolek:
Visual Studio uses 4 spaces instead of 2.


nijel:
The tests need to be adjusted as well.

PR #5179, opened 2024-02-02
Add support for i18next-v4 json files with different selectors via flat key handling

mkarol-neurosys:
Due to internal use of i18next-v4 translation files with nonstandard key separators, I've prepared the patch that allows for handling such use cases as flat json files, while preserving handling of plurals. This PR will be followed by  similar PRs to Weblate.

I've added changes to I18Next-V4 tests in order to reduce number of copy-pasted tests.

Have a great day!
Michał Karol

### DIFF
--- a/tests/translate/storage/test_jsonl10n.py
+++ b/tests/translate/storage/test_jsonl10n.py
@@ -59,6 +59,46 @@ JSON_I18NEXT_V4_FIXED = """{
 }
 """
 
+JSON_FLAT_I18NEXT_V4 = """{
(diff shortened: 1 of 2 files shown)


nijel:
Can you please describe what is your use case? The only difference between these two should be when adding new units.

nijel:
This should work fine with existing storage - it will parse and update such files just fine as they keys are preserved in the round trip. The only code path is different is when adding units manually using a string key.

nijel:

PR #5185, opened 2024-02-08
pypo: avoid adding extra msgctxt if there are 3 or more same msgids

### LINKED ISSUE #4394: json2po lists the the first context-ID multiple times for a string appearing multiple times in a JSON file
I have a JSON file which conatisn the same phrase in different places. But instead of listing all the context-IDs for all occurrences json2po leads to the same ID (the first one) being listed several times. Here is an example:

```
#: .app.Active
msgctxt ".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
".app.Active"
msgid "Active"
msgstr ""
```

gemmaro:
Hello,

This make json2po avoids multiple msgctxt in the resulting PO file.
The cause was in the duplicate removal phase; the location of the first occurrence, among other two or more PO entries with the same keys, was added after the third time and later.

Close #4394.
This also adds a minimal example case as follows.
If we have the JSON source:

```json
{
    "a": {
        "x": "X"
    },
    "b": {
        "x": "X"
    },
    "c": {
        "x": "X"
    }
}
```

... then json2po produced PO content whose `.a.x`'s msgctxt was `".a.x\n.a.x"` (with `pypo.py`, not specifying `USECPO=2` or `USECPO=1`), not `".a.x"`:

```po
#: .a.x
msgctxt ".a.x"
".a.x"
msgid "X"
msgstr ""

#: .b.x
msgctxt ".b.x"
msgid "X"
msgstr ""

#: .c.x
msgctxt ".c.x"
msgid "X"
msgstr ""
```

Thank you,
gemmaro.

### DIFF
--- a/translate/storage/pypo.py
+++ b/translate/storage/pypo.py
@@ -1,6 +1,7 @@
 #
 # Copyright 2002-2009 Zuza Software Foundation
 # Copyright 2013 F Wolff
+# Copyright 2024 gemmaro <gemmaro.dev@gmail.com>
 #
 # This file is part of the Translate Toolkit.
 #
@@ -987,8 +988,7 @@ class pofile(pocommon.pofile):
                         uniqueunits.append(thepo)
                 elif duplicatestyle == "msgctxt":
                     origpo = id_dict[id]
-                    if origpo not in markedpos and id:
-                        # if it doesn't have an id, we already added msgctxt
+                    if origpo not in markedpos and not origpo.msgctxt:
                         origpo.msgctxt.append(
                             '"%s"' % escapeforpo(" ".join(origpo.getlocations()))
                         )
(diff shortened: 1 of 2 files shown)


translate/storage/pypo.py
```diff
@@ -989,9 +990,10 @@ def addcomment(thepo):
                     origpo = id_dict[id]
                     if origpo not in markedpos and id:
                         # if it doesn't have an id, we already added msgctxt
```
nijel, line 992: The code was apparently written with the assumption that `id` will be blank/None if the `msgctxt` was already handled. If this is not the case, the `id` should be removed from the condition and the comment removed as well. Maybe you can also merge the `origpo.msgctxt` condition here?

gemmaro: I think you are right. I've applied your suggestion.

nijel:

PR #5203, opened 2024-03-07
quote: handle all control chars equally

### LINKED ISSUE #5199: Weird handling of C1 control characters in .properties files
When encoding, C1 control characters (U+0000 - U+001F) are not `\` escaped except for `\f`, `\n`, `\r`, and `\t`.

When decoding, escaped C1 control characters (except as above) are not decoded, but left as literal `\u` strings.

This leads to the following unexpected behaviour:

```python
from translate.misc import quote

quote.javapropertiesencode(quote.propertiesdecode('\u0001')) # → '\x01' (expected '\\u0001')

quote.javapropertiesencode(quote.propertiesdecode('\\u01')) # → '\\\\u0001' (expected '\\u0001')
```

The same behaviour also happens with `quote.java_utf8_properties_encode()`.

nijel:
This makes the behavior consistent for all control characters, not just to these few defined.

Fixes #5199

I'd really like to get this reviewed because I could not find reasoning for the previous behavior and I might have missed something.

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated control character escaping logic for improved efficiency and readability.
- **Tests**
	- Modified an assertion to check for inequality in control escape tests.
	- Added new tests for control character encoding and decoding.

### DIFF
--- a/tests/translate/misc/test_quote.py
+++ b/tests/translate/misc/test_quote.py
@@ -138,7 +138,7 @@ class TestEncoding:
         prefix, suffix = "bling", "blang"
         for control in ("\u0005", "\u0006", "\u0007", "\u0011"):
             string = prefix + control + suffix
-            assert quote.escapecontrols(string) == string
+            assert quote.escapecontrols(string) != string
 
     @staticmethod
     def test_propertiesdecode():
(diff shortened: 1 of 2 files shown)


translate/misc/quote.py
```diff
@@ -431,18 +431,45 @@ def mozillaescapemarginspaces(source: str) -> str:
 controlchars = {
     # the reverse of the above...
     "\\": "\\\\",
-    "\f": "\\f",
+    "\x00": "\\x00",
```
eemeli, line 434: These should really be four-digit `\u` escape codes, as `\x` is not one of the Java [escape sequences](https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-EscapeSequence). Same change applies to all of the `\x` codes below.

```suggestion
    "\x00": "\\u0000",
```

translate/misc/quote.py
```diff
@@ -431,18 +431,45 @@ def mozillaescapemarginspaces(source: str) -> str:
 controlchars = {
     # the reverse of the above...
```
eemeli, line 432: This is a bit misleading now.
```suggestion
```

tests/translate/misc/test_quote.py
```diff
@@ -149,6 +149,11 @@ def test_propertiesdecode():
         assert quote.propertiesdecode("abc\\") == "abc\\"
         assert quote.propertiesdecode("abc\\") == "abc\\"
 
+    @staticmethod
+    def test_controlchars_round_trip():
```
eemeli, line 153: It would be good to also check that the plain encode & decode operations work as expected, i.e. that:
```python
encode('\x01') == '\\u0001'
decode('\x01') == '\x01'
decode('\\u0001') == '\x01'
```

eemeli (changes requested):

translate/misc/quote.py
nijel: True, I've updated the patch.

eemeli (approved):

PR #5216, opened 2024-03-21
Fixes problem with removing children while iterating

### LINKED ISSUE #5215: Removing items during iteration can result in bugs
Removing the children of a node on [this line](https://github.com/translate/translate/blob/d7922d50628dfc2964e9ae69437bf0c95851e2ce/translate/misc/xml_helpers.py#L300) can result in missing some of the children.
A way that can fix this might be
```python
while len(node) > 0:
   node.pop()
```

AryazE:
Fixed #5215


nijel:

PR #5269, opened 2024-06-19
Fix added spaces in XWikiPageProperties format XML output

surli:
* Ensure to clean up node after performing operation on the XML
  * Fix existing tests

See: https://github.com/WeblateOrg/weblate/issues/11896


nijel:

PR #5324, opened 2024-08-23
Silently ignore the root element in lisa.py

### LINKED ISSUE #5323: FutureWarning: This search incorrectly ignores the root element, and will be fixed in a future version.  If you rely on the current behaviour, change it to './/body'
When converting using po2tmx, the following is raised:

```
/usr/lib/python3.12/site-packages/translate/storage/lisa.py:312: FutureWarning: This search incorrectly ignores the root element, and will be fixed in a future version.  If you rely on the current behaviour, change it to './/body'
  self.body = self.document.find("//%s" % self.namespaced(self.bodyNode))

```

This warning comes from https://github.com/lxml/lxml/commit/3a8caecc0f94d8e00ab116a4b32e57675140eb4d as of lxml 5.1.1

Context of the code executed:

https://github.com/translate/translate/blob/47c3819aef6998bde5a1e8c2d142eadad79e3967/translate/storage/lisa.py#L306-L312

Steps to reproduce:
- Have lxml 5.1.1 or newer installed
- Run `po2tmx -l your_language_code -i some_input.po -o some_output.tmx`

rffontenelle:
Fixes #5323

### DIFF
--- a/translate/storage/lisa.py
+++ b/translate/storage/lisa.py
@@ -309,7 +309,7 @@ class LISAfile(base.TranslationStore):
         again.
         """
         self.namespace = self.document.getroot().nsmap.get(None, None)
-        self.body = self.document.find(f"//{self.namespaced(self.bodyNode)}")
+        self.body = self.document.find(f".//{self.namespaced(self.bodyNode)}")
 
     def addsourceunit(self, source):
         """Adds and returns a new unit with the given string as first entry."""


nijel:
This silents the `FutureWarning`, but changes semantics as well, what I wanted to avoid. But as the tests pass, it should be safe to merge…

PR #5331, opened 2024-08-30
Update optional.txt

alvinlazz:
Reformated text for readable


nijel (commented):
Thanks for your effort, but this introduces breaking change and I don't think it's more readable.

Also what would be a more useful thing is migrating all this to `pyproject.toml` instead of this.

PR #5338, opened 2024-09-06
docs: modernize documentation build

### LINKED ISSUE #5336: docs.translatehouse.org <pre> tags are illegible
Sections attempting to display code, input, output use `<pre>` elements.  These are given a background color on line 734 of bootstrap.css:
`background-color: #f5f5f5;`
which overlays the parent element's background color from its `class="highlight"`; pygments.css line 7:
` background: #111111;`
So when the content inside the `<pre>` element sets its own text colors, predominantly white (`#ffffff`), which of course is too similar to `#f5f5f5` to be visible.

Here's a screenshot of one from [this page](https://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/txt2po.html):
translatehouse-invisible-pre-text

This affects all but the last `<pre>` element on this page.  In that case, the content does not specify any text color, so it inherits the color specified by bootstrap.css which overrides the color set by class `highlight`.

So, for most of these, adding a `.highlight > pre` selector might help, but that would break the last case, making black on black.

There are other issues with this website, which makes me wonder whether it's being actively maintained:
1. The "Site" and "Page" buttons in the top menu do nothing.
2. The footer has a copyright notice from last year (not that I care; it's just an indication of neglect).
3. The footer gives a link to a broken site: Created using [Sphinx](http://sphinx.pocoo.org/) 8.0.2.  (It gives a 400 HTTP error.)

nijel:
- switch to Furo theme instead of using own one (fixes #5336)
- add copy button to code samples
- add opengraph previews to all pages
- update for upcoming readthedocs.org changes (see https://about.readthedocs.com/blog/2024/07/addons-by-default/)
- dropped all submodule logic as the only submodule is now gone
- avoid redirects in intersphinx caused by trailing slash
- use separate requirements for docs
- add libgettext to doc build for cpo docs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,3 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
-- package-ecosystem: gitsubmodule
-  directory: /
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,8 +17,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-      with:
-        submodules: recursive
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
--- a/.github/workflows/setup.yml
+++ b/.github/workflows/setup.yml
@@ -19,8 +19,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-      with:
-        submodules: recursive
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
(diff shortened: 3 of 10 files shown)


nijel:
Built documentation is available at https://translate--5338.org.readthedocs.build/projects/translate-toolkit/en/5338/, looking for feedback!

phlax (approved):
lgtm, thanks @nijel

rffontenelle:
Update copyright year as well?

nijel:
IMHO it is better to remove it, did that just now.

rffontenelle (approved):

julen:
Looks good! Thanks for taking care of that.

PR #5346, opened 2024-09-19
Add translateable marking to units

harriebird:
## Proposed changes

This adds a `marktranslateable` function to the `TranslationUnit` and `AndroidResourceUnit` to enable the translateable marking to units.

### DIFF
--- a/translate/storage/aresource.py
+++ b/translate/storage/aresource.py
@@ -178,8 +178,8 @@ class AndroidResourceUnit(base.TranslationUnit):
     def istranslatable(self):
         return bool(self.getid()) and self.xmlelement.get("translatable") != "false"
 
-    def marktranslateable(self, value):
-        return self.xmlelement.set("translatable", value)
+    def marktranslatable(self, value: bool) -> None:
+        return self.xmlelement.set("translatable", "true" if value else "false")
 
     def isblank(self):
         return not bool(self.getid())
--- a/translate/storage/base.py
+++ b/translate/storage/base.py
@@ -408,7 +408,7 @@ class TranslationUnit:
         """
         return bool(self.source)
 
-    def marktranslateable(self, value):
+    def marktranslatable(self, value: bool) -> None:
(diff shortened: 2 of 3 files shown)


nijel:
Can you please add tests to verify that the round-trip of the flag workds?

harriebird:
Added a named `test_translatable_marking`.

translate/storage/base.py
```diff
@@ -408,6 +408,9 @@ def istranslatable(self):
         """
         return bool(self.source)
 
+    def marktranslateable(self, value):
```
nijel, line 411: As the getter is called `istranslatble`, this should be named consistently. Also I'd prefer it to accept boolean in the API as the string values are Android-specific:

```suggestion
    def marktranslatable(self, value: bool) -> None:
```

translate/storage/aresource.py
```diff
@@ -178,6 +178,9 @@ def __init__(self, source, empty=False, xmlelement=None, **kwargs):
     def istranslatable(self):
         return bool(self.getid()) and self.xmlelement.get("translatable") != "false"
 
+    def marktranslateable(self, value):
+        return self.xmlelement.set("translatable", value)
```
nijel, line 182: As the getter is called `istranslatble`, this should be named consistently. Also I'd prefer it to accept boolean in the API as the string values are Android-specific:


```suggestion
    def marktranslatable(self, value: bool) -> None:
        return self.xmlelement.set("translatable", "true" if value else "false")
```

translate/storage/base.py
harriebird: Updated the code to this change suggestion.

translate/storage/aresource.py
harriebird: Updated the code based on your suggestion. I agree on the point of making `marktranslatable` function to accept boolean value. This will make its implementation to other format easier.

nijel:

Issue #5348, opened 2024-09-22
PHP: keys/names are not getting escaped


johnstef99:
In a php translation file it's possible to have a key that contains a quote `'` for example:
```php
<?php

return [
  'john\'s ball' => 'john\'s ball',
];
```

when converted to po and back to php the above will turn to a broken php file
```php
<?php

return [
  'john's ball' => 'john\'s ball',
];
```

in [/translate/storage/php.py:282](https://github.com/translate/translate/blob/41bf063b281fe356a34bb9dca68424f517b0bb93/translate/storage/php.py#L282) the function getoutput should escape the name too.

Trying to add something like this
```python
 q1 = name.find(self.escape_type)
 q2 = name.rfind(self.escape_type)
 if q1 != -1 and q2 != -1:
     name = self.escape_type + phpencode(name[q1+1:q2], self.escape_type) + self.escape_type
```
works fine for me but I'am sure this is not the best way to do that so I will not open a PR.

Could we get this fixed?

johnstef99:
I think [this issue](https://github.com/WeblateOrg/weblate/issues/5048) is related

PR #5350, opened 2024-09-23
PHP: escape quotes in keys

johnstef99:
Fixes: #5348

### DIFF
--- a/tests/translate/storage/test_php.py
+++ b/tests/translate/storage/test_php.py
@@ -318,26 +318,20 @@ $foo = 'bar';
 
     def test_parsing_arrays_keys_with_quotes(self):
         """Ensure that our identifiers get escaped. Bug #5348."""
-        phpsource = """return [
-         'item\\'s 1' => 'value1',
-         "item's 2" => 'value2',
-         'item "3"' => 'value3',
-         "item \\"4\\"" => 'value4',
-      ];"""
+        phpsource = r"""<?php
+return [
+    'item\'s 1' => 'value1',
+    'item "2"' => 'value2',
+];
+"""
         phpfile = self.phpparse(phpsource)
-        assert len(phpfile.units) == 4
+        assert len(phpfile.units) == 2
         phpunit = phpfile.units[0]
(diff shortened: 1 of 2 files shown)


translate/storage/php.py
```diff
@@ -285,6 +285,8 @@ def getoutput(self, indent="", name=None):
             fmt = "{1}{2}{1},\n"
         elif "->" in self.name:
             fmt = "{0} => {1}{2}{1},\n"
+            if name is not None:
+                name = name[0] + phpencode(name[1:-1]) + name[-1]
```
nijel, line 289: It should include escape type:

```suggestion
                name = name[0] + phpencode(name[1:-1], name[0]) + name[-1]
```

Can you please also add a test case for double-quoted string?

tests/translate/storage/test_php.py
```diff
@@ -316,6 +316,29 @@ def test_parsing_arrays_keys_with_spaces(self):
...
+        assert phpunit.source == "value1"
+        phpunit = phpfile.units[1]
+        assert phpunit.name == "return[]->'item's 2'"
+        assert phpunit.source == "value2"
+        phpunit = phpfile.units[2]
+        assert phpunit.name == "return[]->'item \"3\"'"
+        assert phpunit.source == "value3"
+        phpunit = phpfile.units[3]
+        assert phpunit.name == "return[]->'item \"4\"'"
+        assert phpunit.source == "value4"
```
nijel, line 340: Reading through this again, it does test only parsing, not serialization, something like this will be needed for that:

```suggestion
        assert phpunit.source == "value4"
        
        assert bytes(phpfile).decode() == phpsource
```

johnstef99:
@nijel correct me if I am wrong but I think when serializing back to php file the keys in a associated array can only have single quotes

nijel:
The `phpunit.escape_type` should define that, and it should be populated while parsing. But it IMHO only applies for values, not for keys (and it doesn't make sense to apply it to both as they can differ). So you're probably right on this.

nijel:

Issue #5355, opened 2024-09-26
PHP: keys with numbers as values get duplicated 


johnstef99:
Example php file:
```php
<?php
return [
  'some_enum' => [
    1 => 'One',
    2 => 'Two',
  ],
];
```
Gets turned into:
```php
<?php
return [
  'some_enum' => [
    11 => 'One',
    22 => 'Two',
  ],
];
```

This bug was cause by this PR #5350

PR #5356, opened 2024-09-26
PHP: fix duplicate numbers as keys in arrays

johnstef99:
Fixes: #5355


translate/storage/php.py
```diff
@@ -285,8 +285,10 @@ def getoutput(self, indent="", name=None):
             fmt = "{1}{2}{1},\n"
         elif "->" in self.name:
             fmt = "{0} => {1}{2}{1},\n"
-            if name is not None:
-                name = name[0] + phpencode(name[1:-1], name[0]) + name[-1]
+            if name is not None and name[0] == name[-1] == self.escape_type:
```
nijel, line 288: This will break for strings having `self.escape_type == '"'` because the name has always single quote now. I'm not sure it's a good idea to use escape_type as that is the escape type detected for the value, not for the key (I've already mentioned this in https://github.com/translate/translate/pull/5350).

johnstef99: so should i change it to name[0] then?

tests/translate/storage/test_php.py
```diff
@@ -335,6 +335,29 @@ def test_parsing_arrays_keys_with_quotes(self):
 
         assert bytes(phpfile).decode() == phpsource
 
+    def test_parsing_arrays_keys_with_number_as_value(self):
+        """Ensure that our identifiers can have numbers as value. Bug #5355."""
+        phpsource = r"""<?php
+return [
+    1 => 'value1',
+    2 => 'value2',
+    '3' => 'value3',
```
nijel, line 344: Let's include some additional quotes to verify escaping actually works:

```suggestion
    '3\'' => 'value3',
    '4"' => "value4",
    '5\'' => "value4",
```

translate/storage/php.py
```diff
@@ -285,8 +285,12 @@ def getoutput(self, indent="", name=None):
             fmt = "{1}{2}{1},\n"
         elif "->" in self.name:
             fmt = "{0} => {1}{2}{1},\n"
-            if name is not None:
-                name = name[0] + phpencode(name[1:-1], name[0]) + name[-1]
+            if (
+                name is not None
+                and name[0] == name[-1]
+                and (name[0] == "'" or name[0] == '"')
```
nijel, line 291:
```suggestion
                and name[0] in {"'", '"'}
```

PR #5423, opened 2024-11-28
Allow po2rc output encoding to be specified

jon-turney:
Allow po2rc output encoding to be specified with the '--charset-output' option, defaulting to automatically choosing cp1252 if possible, otherwise utf-16-le, as before.

### DIFF
--- a/translate/convert/po2rc.py
+++ b/translate/convert/po2rc.py
@@ -355,6 +355,7 @@ def convertrc(
     lang=None,
     sublang=None,
     outputthreshold=None,
+    output_charset=None,
 ):
     inputstore = po.pofile(inputfile)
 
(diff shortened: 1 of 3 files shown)


jon-turney:
Not generating output files with a definite encoding seems less than ideal, and I want to specify UTF8.

If I recall correctly, I've implemented this with a consistent option name to other tools which have this functionality, but since I wrote it a while ago, so maybe I forget to check that.

nijel:
Maybe utf-8 would be a more reasonable default encoding for RC files these days...

jon-turney:
> Maybe utf-8 would be a more reasonable default encoding for RC files these days...

Quite possibly.  I don't use MSVS tools, only gcc, so I don't know what's appropriate there.

Plus someone would probably complain about changing the default :)

translate/convert/po2rc.py
```diff
@@ -367,11 +368,14 @@ def convertrc(
...
-        outputfile.write(codecs.BOM_UTF16_LE)
-        outputfile.write(outputrclines.encode("utf-16-le"))
+    if output_charset == "auto":
+        try:
+            outputfile.write(outputrclines.encode("cp1252"))
+        except UnicodeEncodeError:
+            outputfile.write(codecs.BOM_UTF16_LE)
+            outputfile.write(outputrclines.encode("utf-16-le"))
+    else:
+        outputfile.write(outputrclines.encode(output_charset))
```
nijel, line 378: Can you please add test so make sure this code path is covered as well?

jon-turney:
> consistent option name to other tools

Now I look at it again, I don't see this option for other tools, so I guess I just made it up.

> Can you please add test so make sure this code path is covered as well?

I've made a minimal test, but this is slightly complicated by TestConvertCommand.run_command() not anticipating that option names can contain dashes.

This should probably also check that the output actually has the required encoding.

nijel:

PR #5458, opened 2025-01-06
Use flat structure for Webextension JSON files

kewisch:
I noticed on Weblate that WebExtension JSON keys are pulled apart. I don't believe this is supported in Chrome or Firefox. This PR uses flat unit ids instead.


nijel:
Can you please describe when this happens in Weblate? It is supposed to keep the existing key structure, and changes made in this pull request should only affect newly added strings. So if you see Weblate breaking keys for existing strings, please describe when that happens in https://github.com/WeblateOrg/weblate/issues/13037 (the format is different, but the issue is the same).

PR #5465, opened 2025-01-13
feat(android): include trailing newline in generated XML

### LINKED ISSUE #5464: Handling of final newlines
### Describe the issue

Follow-up of https://github.com/WeblateOrg/weblate/discussions/13452

Sum up: Weblate shouldn't remove final newlines on text-based translation files:
- It's the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) (this includes the last line of the file)
- From a Git point of view, "if the last line in a file doesn't end with a newline then addition of next line affects two lines instead of one. This also pollutes diff on multiple files, so reader may wonder what has changed in a line whereas no significant change has occurred".
- Everytime an user edits the translation file (for example to remove obsolete translations) on its IDE/text editor, at least on Unix systems, the IDE/text editor will automatically (by default) adds a newline at the end of the file. However, after pushing the changes, Weblate will immediately commit a change (or multiple, if you don't squash) just to remove the final newline. This leads to unnecessary commit pollution and lint inconsistency of files across the project.

### Steps to reproduce the behavior

1. Remove an obsolete translation in Android Studio (on Linux)
2. Weblate immediately commits to remove the newline added by Android Studio

### Expected behavior

There are multiple ways to have a more reasonable behavior:
1. Always leave file endings as they are. If you receive a file with a newline at the end, leave it like that. Same if you receive it without. This is probably the most logical choice, but maybe not the easiest to implement technically. This can however lead to inconsistencies with one language file with a newline and another without, but that's the role of a lint system to report them if the maintainer cares about it.
2. Add a Weblate preference: there are already many, so probably not the best. IMO it should enable final newlines by default if you go that way.
3. If you don't want to add a preference on Weblate side, you could at least honor the `.editorconfig` file and its `insert_final_newline` preference. However, I don't think many projects use it and this could be a rather hidden way to fix the issue.
4. Just force everyone to use final newlines. After all, there are good reasons to have them, but none to not have them.

nijel:
Fixes #5464

I'm not really convinced this is a good idea, as this changes output for everybody, so I'd truly like to get some feedback on this change.

@NLLAPPS, @razor313, @johan12345, @TrianguloY, @eighthave, @JustSwiftIt, @nikclayton I'm mentioning people who recently reported issues on Android in this repository or Weblate and might provide valuable feedback here.

### DIFF
--- a/tests/translate/storage/test_aresource.py
+++ b/tests/translate/storage/test_aresource.py
@@ -586,7 +586,8 @@ class TestAndroidResourceFile(test_monolingual.TestMonolingualStore):
         content = b"""<?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="test">Test</string>
-</resources>"""
+</resources>
+"""
         store = self.StoreClass()
         store.parse(content)
         assert bytes(store) == content
@@ -605,7 +606,8 @@ class TestAndroidResourceFile(test_monolingual.TestMonolingualStore):
 <resources>
     <string name="test">Test</string>
     <string name="other">Test</string>
-</resources>"""
+</resources>
+"""
         store = self.StoreClass()
         store.parse(content)
         otherstore = self.StoreClass()
@@ -621,7 +623,8 @@ class TestAndroidResourceFile(test_monolingual.TestMonolingualStore):
 <resources>
     <string name="app_name">&appName;</string>
     <string name="app_core">&appName; Core</string>
-</resources>"""
+</resources>
+"""
         store = self.StoreClass()
         store.parse(content)
         assert store.units[0].source == "&appName;"
@@ -638,7 +641,8 @@ class TestAndroidResourceFile(test_monolingual.TestMonolingualStore):
 <resources>
     <string name="app_name">&appName;</string>
     <string name="app_core">&appName; Core</string>
-</resources>"""
+</resources>
+"""
         store = self.StoreClass()
         store.parse(content.encode())
         store.units[0].target = "&appName;"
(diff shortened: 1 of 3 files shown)


nikclayton:
Assuming this performs as advertised, I'm in favour. The PRs from Weblate with stripped newlines always make me do a slight double-take when I'm looking over them, as I think "Wait, why is the final line of the file being changed?". Then I remember that Weblate removes the existing newlines.

Handily I have an open PR that demonstrates this -- https://github.com/pachli/pachli-android/pull/1192/files#diff-5571d7ae177e0281e478a190bcdce6d8f78db1cc09b012d0c54990d9a36bd932, The perfectly correct `</resources>\n` is being replaced by the slightly incorrect `</resources>`.

TrianguloY:
Personally I don't mind if this change is merged or not, although I'm more in favor of doing so.

On the repository where I use weblate I rarely touch the translation files directly, so I rarely encounter this change. I do remember this in the past though, but as I manually review and merge them when I need, there is no problem (also I think I have xml formatting turned off due to other issues with android studio).

But as I said, I will welcome this change. Having a line break at the end is the default in basically all IDEs (plus a good practice even if you don't).

eighthave:
This behavior needs to match Android Studio's behavior since the vast majority of Android apps are developed using Android Studio.  Last I checked, Android Studio, stripped new lines from XML.  That was years ago, they do sometimes make arbitrary changes.  If this behavior does not match Android Studio, then most Android devs will get regular merge conflicts with Weblate.

eighthave:
The best would be if Weblate does not change the end of the file, and just keeps the ending of the file as is.

Issue #5536, opened 2025-03-11
Bug with Ruby Plurals.


Richar-p:
Hello.

I discover some bugs with RubyYAML parsing, i created some tests to reproduce them :

```python
    def test_bug_ruby_remove_zero_few_and_mix_others(self):
        data = """en:
  string:
  message:
    one: value 1
    other: value 2
    zero: value 3
    few: value 4
"""
        store = self.StoreClass()
        store.parse(data)
        assert bytes(store).decode() == data
```

With this test, we see that "few" and "zero" keys are deleted after parsing.
Also, the values are mixed up, the key "one" have the value "value 3" and the key "other" have the "value 1" 

```text
(Pdb) print(bytes(store).decode())
en:
  string:
  message:
    one: value 3
    other: value 1
```

Also another point, in ruby, replace empty string to '' lock totally the possibility of fallback to another language (https://github.com/translate/translate/commit/2c21e1f46c0f1d0250787aa271d331c55ef6af2b) sorry i didn't think to it before.

I will try my best to make a PR on it, but python is not my specialty lmao

Have a nice day

nijel:
The available plurals depend on a language, so you should set the language on the storage to support more plurals than one/other.

Richar-p:
Okay, i didn't understood at first time that is a filter on it by langages, mybad.
But it keep having a problem with values which go to the wrong key, no ?

Richar-p:
https://github.com/translate/translate/pull/5537 MR created

nijel:
Plurals should be parsed into a `multistring` class in defined order. The same order is used while serializing.

nijel:
https://github.com/translate/translate/pull/5573 should be a proper fix not breaking existing API promises.

Richar-p:
Ok, i understood the issue, thanks for fixing it.
The problem is the PR i created was here also to fix another problem :

![Image](https://github.com/user-attachments/assets/be786d78-9ed9-4488-82a3-69583ed70163)

Should i create another ticket for it ?

nijel:
Empty string would not be parsed at all:

```
en:
  test:
```

Results in none units and would be serialized without this string:

```
en:
```

That should end up in using fallback in Ruby.

This behavior is intentional because with lack of value, the expected structure is not known, it can be a string:

```
en:
  test: Value
```

But it can also be a nested key:

```
en:
  test: 
    success: Value
    failure: Other
```

Richar-p:
[...]

PR #5537, opened 2025-03-12
fix(RubyYAML) : Plural issues

Richar-p:
Hope this is good, i'm really beginner with Python, GPT help me a lot

Goals of this edit : 

- Empty string should continue to be empty, not begin : ''

In ruby on rails, there is a big difference between empty and null strings.

For example, my en.yml look like this : 

```yaml
en:
  test: ''
```

I will have this result :
```ruby
irb(main):001> I18n.t('test')
=> ""
```

But with an empty string like this : 

```yaml
en:
  test:
```

I will have this result (and a fallback into default langage in production) :
```ruby
irb(main):001> I18n.t('test')
=> "Translation missing: en.test"
``` 

- Locales values was mixed up

As i mentionned on the based issue, this sample test: 

```python
    def test_bug_ruby_remove_zero_few_and_mix_others(self):
        data = """en:
  string:
  message:
    one: value 1
    other: value 2
    zero: value 3
    few: value 4
"""
        store = self.StoreClass()
        store.parse(data)
        assert bytes(store).decode() == """en:
  string:
  message:
    one: value 1
    other: value 2
"""
```

Was not working because values was mixed up. the key "one" took the value 3 and the key "other" took the value 1.
Not rly sure of why this happened exactly, maybe because of the multistring system ?

Issue https://github.com/translate/translate/issues/5536

### DIFF
--- a/tests/translate/storage/test_yaml.py
+++ b/tests/translate/storage/test_yaml.py
@@ -1,6 +1,7 @@
 import pytest
 import ruamel.yaml
 
+from translate.misc.multistring import multistring
 from translate.storage import base, yaml
 
 from . import test_monolingual
@@ -541,6 +542,9 @@ class TestRubyYAMLResourceStore(test_monolingual.TestMonolingualStore):
         store = self.StoreClass()
         store.parse(data)
         assert len(store.units) == 1
+        assert store.units[0].target == multistring(
+            ["There is one message", "There are %{count} messages"]
+        )
         assert bytes(store) == data.encode("ascii")
 
     def test_empty(self):
@@ -620,6 +624,7 @@ class TestRubyYAMLResourceStore(test_monolingual.TestMonolingualStore):
 """
         store = self.StoreClass()
         store.parse(data)
+        assert store.units[0].target == multistring(["", ""])
         assert len(store.units) == 1
         assert (
             bytes(store).decode()
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
       with:
         python-version: '3.13'
-    - uses: astral-sh/setup-uv@0c5e2b8115b80b4c7c5ddf6ffdd634974642d182 # v5.4.1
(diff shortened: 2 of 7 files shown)


tests/translate/storage/test_yaml.py
```diff
@@ -540,7 +540,7 @@ def test_ruby_plural(self):
 """
         store = self.StoreClass()
         store.parse(data)
-        assert len(store.units) == 1
+        assert len(store.units) == 2
```
nijel, line 543: This seems wrong, it should be a single unit with plurals.

Richar-p: Yes, fixed, i understood the issue.

```
$ pytest tests/translate/storage/test_yaml.py                    
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.12.5, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/peter-richard/RubymineProjects/translate-toolkit
configfile: pytest.ini
plugins: xdist-3.6.1, cov-6.0.0, syrupy-4.8.1
collected 76 items                                                                                                                                                                                            

tests/translate/storage/test_yaml.py ............................................................................                                                                                       [100%]

============================================================================================= 76 passed in 0.36s ==============================================================================================
(3.12.5) 
```

spuyet:
@nijel would you mind having another look to this PR please ?

spuyet:
up @nijel :)

winnieoursbrun:
Would be useful for myself !

translate/storage/yaml.py
```diff
@@ -218,17 +220,56 @@ def get_root_node(self):
...
+        if isinstance(data, dict):
+            plural_keys = [k for k in data if k in cldr_plural_categories]
+
+            # ✅ If all plural categories are present but some are empty, we still need the unit
+            if plural_keys and len(plural_keys) == len(data):
+                if any(v is not None for v in data.values()):
+                    # At least one non-empty value, so create a unit
+                    yield (
+                        prev,
+                        {k: data[k] for k in cldr_plural_categories if k in data},
```
nijel, line 233: This makes parsing return `dict` instead of `multistring`, that will break any Python API user because nobody expects `dict` here. This part was not covered by the tests, I've just added that in https://github.com/translate/translate/pull/5572.

PR #5561, opened 2025-04-03
po2rc: Correctly escape double-quotes embedded in a string in a RC file

jon-turney:
See https://learn.microsoft.com/en-us/windows/win32/menurc/stringtable-resource

> To embed quotes in the string, use the following sequence: "".

Because we may have some already translated strings which contain a literal "" (i.e. \"\" in a .po file) to work-around po2rc not doing this correctly so far, for backwards compatibility look around the " and only escape it if it isn't already part of a "" escape.

Future work: This doesn't do anything for untranslated strings. That's hard because those are passed through verbatim, including their existing formatting and linebreaks.

### DIFF
--- a/translate/convert/po2rc.py
+++ b/translate/convert/po2rc.py
@@ -23,6 +23,7 @@ for examples and usage instructions.
 """
 
 import codecs
+import re
 from collections.abc import Iterable
 
 from translate.convert import convert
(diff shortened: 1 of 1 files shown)


nijel:
Looking at this again, the quotes are already escaped when `inputdict` is built. If you encounter any specific issue with the behavior, please describe it. https://github.com/translate/translate/pull/5643 adds tests to prove that the current solution works as expected.

PR #5574, opened 2025-04-23
feat(android): more robust quotes handling

### LINKED ISSUE #5543: Double quote mismatch when there is XML markup inside them (Android)
I was asked to file an issue here from: https://github.com/WeblateOrg/weblate/issues/14214

This string has double quote at the end but not at the start: https://hosted.weblate.org/translate/microg/play-services-nearby-core-strings/en/?q=exposure_confirm_start_summary
While the original string has both: https://github.com/microg/GmsCore/blob/3d54307cc02aed50c0f546962437c8f49606b1f7/play-services-nearby/core/src/main/res/values/strings.xml#L54

Specifications: https://developer.android.com/guide/topics/resources/string-resource?hl=en#FormattingAndStyling

nijel:
- handle quotes across XML elements
- gracefully handle unpaired quotes

Fixes #5543

### DIFF
--- a/tests/translate/storage/test_aresource.py
+++ b/tests/translate/storage/test_aresource.py
@@ -1,6 +1,5 @@
 from copy import copy, deepcopy
 
-import pytest
 from lxml import etree
 
 from translate.misc.multistring import multistring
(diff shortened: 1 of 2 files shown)


lucasmz-dev:
is this already live in Weblate?

nijel:
Yes, it was published in the Docker containers so any up to date instal has it.

PR #5585, opened 2025-05-05
Fix: Polish are missing a plural form

Richar-p:
Hello,

There is currently a problem with the polish plural, which are missing a form : "other" as we can see on the CLDR 44 (which is the version used according to the comment)

https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html
image

Edit: 
After double checking, i think it is mostly ruby frameworks which force the usage of this key. But it cause problem for the weblate usage.. :/
image

### DIFF
--- a/translate/lang/data.py
+++ b/translate/lang/data.py
@@ -689,7 +689,7 @@ plural_tags = {
     "per": ["one", "other"],
     "pes": ["one", "other"],
     "pk": ["one", "other"],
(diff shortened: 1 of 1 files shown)


nijel:
Unfortunately, this is a more complex topic and such a change won't make it work correctly, see https://github.com/WeblateOrg/weblate/issues/7520#issuecomment-1116032189.

PR #5601, opened 2025-06-04
Parse "administrativeStatus" and "Translation needed" metadata in TBX files

amCap1712:
Related to https://github.com/WeblateOrg/weblate/issues/12795

### DIFF
--- a/translate/storage/tbx.py
+++ b/translate/storage/tbx.py
@@ -84,9 +84,14 @@ class tbxunit(lisa.LISAunit):
             note.set("from", origin)
 
     def _getnotenodes(self, origin=None):
+        """Get all nodes matching ``origin`` in the XML document."""
         return self.xmlelement.iterdescendants(self._get_origin_element(origin))
 
     def _getnodetext(self, node):
+        """
+        Get the plaintext content of the given node considering the xml namespace
+        and space configuration.
+        """
         return lisa.getText(node, getXMLspace(self.xmlelement, self._default_xml_space))
 
     def _is_administrative_status_term_node(self, node) -> bool:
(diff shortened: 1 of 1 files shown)


translate/storage/tbx.py
```diff
@@ -113,6 +139,21 @@ def _getnotelist(self, origin=None):
...
+        return super().istranslatable()
+
+    def isobsolete(self) -> bool:
+        for note in self._getnotenodes(origin="pos"):
+            if self._is_administrative_status_term_node(note) and self._getnodetext(
+                note
+            ).strip().lower() in {"forbidden", "obsolete"}:
+                return True
+
+        return super().isobsolete()
```
nijel, line 172: I'm not sure if this is a good mapping. `isobsolete` units are currently not imported into Weblate at all because these contain strings that were translated in the past and are no longer used. They are stored in the translation files so that they can be reused in case the string re-appears (kind of translation memory). See also https://github.com/WeblateOrg/weblate/issues/1993.

amCap1712: Looking at https://www.tbxinfo.net/ns/dct/basic which seems to be sort of official documentation for TBX, there seems to be `deprecatedTerm-admn-sts   (deprecated)` status value that would map to the obsolete concept here, should I update it to handle deprecated here and handle weblate's concept of `obsolete` in weblate itself? 

The helper `_is_adminstrative_status_node` will simplify the handling for it and also ensure it is excluded from notes so let's keep it here anyway.

nijel: Yes, that sounds as a good approach.

nijel:

PR #5604, opened 2025-06-08
update

ajh1206:
Add exception handling in convert.py

Check if file path exists


nijel:
Both added functions are unused. Thus, this has no effect on the functionality.

PR #5613, opened 2025-06-13
Toki Pona language settings

jnpoJuwan:
Toki Pona (ISO 639-3: `tok`) is a constructed language and small world language.

The settings chosen to ignore the simplecaps and startcaps tests as the language does not use capitalisation for common words, only for proper names, which by its grammar can never show up at the start of the sentence without a common word before it.


nijel:

PR #5648, opened 2025-07-20
`md2po`: Add line numbers, remove block token names, add recursive options.

### LINKED ISSUE #5174: Locations for Markdown
Since https://github.com/translate/translate/pull/5173 mistetoe exposes line numbers, so this should be used as location in md2po.

@anderskaplan Maybe you have prepared for this already?

kattni:
This PR includes the following changes to `md2po`:
- Removes block token names in favor of line numbers for locations.
- Updates `unit.addlocation()` to f-string, and to accommodate line number change.
- Enables `--allowrecursiveoutput` and `--allowrecursiveinput` to allow for providing directories of files for conversion purposes. I may be missing something obvious here to make this work without these options, but after a significant amount of digging into the code and trying various things. I'm relatively sure this is necessary to make this work.

I'm currently figuring out whether switching our project documentation from Sphinx to MkDocs is viable, and this tool is a critical piece to ensure I can do translations via Weblate. These changes will make that process more straightforward, and allow for easier conversion of larger numbers of existing documentation files.

Fixes #5174

### DIFF
--- a/translate/convert/md2po.py
+++ b/translate/convert/md2po.py
@@ -40,6 +40,7 @@ class MD2POOptionParser(convert.ConvertOptionParser):
         super().__init__(formats, usetemplates=False, usepots=True, description=__doc__)
         self.add_duplicates_option()
         self.add_multifile_option()
+        self.add_options(["--allowrecursiveoutput", "--allowrecursiveinput"])
 
     def _extract_translation_units(
         self,
--- a/translate/storage/markdown.py
+++ b/translate/storage/markdown.py
@@ -116,7 +116,8 @@ class MarkdownFile(base.TranslationStore):
 
         # emit a translation unit. The PO store takes care of the escaping.
         unit = self.addsourceunit(text)
-        unit.addlocation("{}+{}".format(self.filename or "", ".".join(path)))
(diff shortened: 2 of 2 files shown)


kattni:
Ah right, tests and docs. I'll look into this soon.

kattni:
Apologies for the delay between submission and getting CI passing; I was running a Python conference last weekend.

I updated the primary comment to reflect the current changes.

Please let me know if this works for you. I read the linked issue, and my interpretation was that the block token names in the locations should be replaced with line numbers. If that is not the case, let me know how you would prefer this be implemented. Thanks!

nijel:
I think this approach makes sense. There is really no consistency in this so far.

For example, HTML also includes the tags path:

https://github.com/translate/translate/blob/c687bad35f8660c20ea2d054bffb226b5a6f4533/translate/storage/html.py#L190C17-L195

kattni:
Please let me know if there's anything else needed to make this merge-ready. The one thing I can think of is documentation - are the block token names documented anywhere, that should be updated to line numbers? If so, I'd appreciate being pointed to the doc that needs to be updated. Thanks!

nijel:

PR #5676, opened 2025-08-30
Fix #5655

### LINKED ISSUE #5655: Text within <button></button> is not translated
**Steps to reproduce:**
1. Go to https://www.golem.de
2. Translate the page (seems to be reproducible with any target language).
3. See the "Zustimmen und weiter" button.

**Expected results:**
The text in the "Zustimmen und weiter" button is translated.

**Actual results:**
The text in the "Zustimmen und weiter" button is NOT translated.

Additional comments:
Is that site-specific?.. The mentioned element is a regular <button>, not sure what else might have caused the issue. However, if the button text is never translated, that's severe.

Image

johnzhou721:
Add button to translated elements.

### DIFF
--- a/translate/storage/html.py
+++ b/translate/storage/html.py
@@ -53,6 +53,7 @@ class htmlfile(html.parser.HTMLParser, base.TranslationStore):
         "article",
         "aside",
         "blockquote",
+        "button",
         "caption",
         "dd",
         "dt",


nijel:
Can you please add a test case to make sure this works as expected?

johnzhou721:
Not having time to work on this anymore.  Closing.  Sorry!

PR #5684, opened 2025-09-04
Fixes `--timestamp` check failing when `--multifile=onefile` is enabled

kattni:
It seems the way that `onefile` works is to generate a series of files, the paths for which are concatenated onto the provided `output` file path and name. As those `fulloutputpath`s don't exist by the time `_output_is_newer` is run, the `.exists` checks in `_output_is_newer` fail and return `False`. Therefore, regardless of a content update, the POT/PO files are regenerated with a new `POT-creation-date` in the header. 

The added check sets the output path used by `_output_is_newer` to the provided output file if `onefile` is enabled, and leaves it as `fulloutputpath` otherwise. 

I tested it with `md2po` which is the tool I am using that brought the issue to light.

I don't see relevant tests in `test_convert.py`, so I didn't add anything there. I am uncertain whether I need to update the tests for every tool that allows `--timestamp` or not. I am creating this PR with the knowledge that I may still need to deal with test updates. If the situation is not clear to me, I will let you know.

### DIFF
--- a/translate/convert/convert.py
+++ b/translate/convert/convert.py
@@ -229,7 +229,12 @@ class ConvertOptionParser(optrecurse.RecursiveOptionParser):
     def processfile(
         self, fileprocessor, options, fullinputpath, fulloutputpath, fulltemplatepath
     ):
-        if options.timestamp and _output_is_newer(fullinputpath, fulloutputpath):
+        if options.multifilestyle == "onefile":
+            timecheck_output_path = options.output
+        else:
+            timecheck_output_path = fulloutputpath
+
+        if options.timestamp and _output_is_newer(fullinputpath, timecheck_output_path):
             return False
 
         return super().processfile(
--- a/translate/convert/convert.py
+++ b/translate/convert/convert.py
@@ -229,6 +229,11 @@ class ConvertOptionParser(optrecurse.RecursiveOptionParser):
     def processfile(
         self, fileprocessor, options, fullinputpath, fulloutputpath, fulltemplatepath
     ):
+        # The `fulloutputpath`s parsed when `onefile` is enabled don't exist, which causes
+        # the `_output_is_newer` check to return False. This means that even if there is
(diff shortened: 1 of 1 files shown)


kattni:
While both work with `md2po`, the current status fails in a couple of ways I'm not sure are acceptable changes to make. The previous commit https://github.com/translate/translate/pull/5684/commits/e2ba59ba393f4a3389b005ac0dd5e1e6b928f127 fails all the tests in a pretty serious way that tells me what I did originally isn't going to work either. 

It's possible neither solution is acceptable. I'm going to wait on a response before continuing to try to figure out what to do here. 

The issue obviously exists, and it's affecting our setup. So, if my solutions aren't the right thing, I would appreciate guidance or let me know if you would prefer to resolve it yourself. Thank you.

nijel:
The problem is that `options.multifilestyle` fails with `AttributeError` for most of the converters—the option is not defined there.

translate/convert/convert.py
```diff
@@ -229,7 +229,20 @@ def run(self, argv=None):
...
+        # no content update, the POT/PO files are regenerated. When `onefile` is enabled,
+        # This check sets the output  path used by `_output_is_newer` to the output file
+        # provided to the command, which will exist. It remains `fulloutputpath` otherwise.
+        try:
+            if options.multifilestyle == "onefile":
+                timecheck_output_path = options.output
+            else:
+                timecheck_output_path = fulloutputpath
+        except AttributeError:
+            timecheck_output_path = fulloutputpath
```
nijel: It's unclear from the code where you expect the attribute to be missing. I think this would make it easier to read:

```suggestion
        if getattr(options, "multifilestyle", None) == "onefile":
            timecheck_output_path = options.output
        else:
            timecheck_output_path = fulloutputpath
```

kattni: That is indeed easier to read. Thank you for the suggestion!

nijel:

PR #5756, opened 2025-10-17
feat(tmx): Add support for context in TMX

gersona:
The suggestion here is to use the `<prop type="x-content">` tag for context
```xml
<prop type="x-context">Context</prop>
```

the other alternative (not implemented here) is to use the `<note>` tag with a keyword like "Context: "
```xml
<note>Context: Software UI</note>
```


gersona:
Is there a reason for this to remain a `@staticmethod` ? It creates minor inconsistencies with descendant classes
https://github.com/translate/translate/blob/859d615ee6a519b622baea9a1e0e6a6b4b148403/translate/storage/base.py#L329-L332

nijel:
I don't think so.

docs/releases/3.16.4.rst
```diff
@@ -0,0 +1,28 @@
...
+-----
+
+
+
+Contributors
+============
+
+This release was made possible by the following people:
+
+Michal Čihař
```
nijel: The release notes are created from Git history here, see https://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/releasing.html#create-release-notes

docs/releases/index.rst
```diff
@@ -17,6 +17,7 @@ Final releases
 .. toctree::
    :maxdepth: 1
 
+   3.16.4 <3.16.4>
```
nijel:
```suggestion
```

nijel:

docs/releases/3.16.4.rst
gersona: My bad, I didn't correctly check the documentation.

PR #5867, opened 2025-11-25
html2po: add data-translate-context for custom msgctxt

Frenzie:
There's sometimes a lack of both stability and clarity in automatically generated `msgctxt` based on DOM tree + line number for duplicate entries. The solution is simple: allow for setting a `data-translate-context` attribute manually.

### DIFF
--- a/translate/storage/html.py
+++ b/translate/storage/html.py
@@ -44,6 +44,7 @@ class htmlunit(base.TranslationUnit):
         self.locations.append(location)
 
     def getlocations(self):
+        """Get the list of locations for this unit."""
         return self.locations
 
     def getcontext(self):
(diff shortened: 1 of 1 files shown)


Frenzie:
I'll ignore DeepSource now. Its demand for more documentation is nonsensical.

nijel:

PS: DeepSource is not really something we enforce, it sometimes gives useful hints, sometimes the warnings are wrong. Maybe it's time to reconsider its usage...

Frenzie:
No problem. :-)

The principle sounds mostly reasonable: add docstrings to functions you add, which is what I thought it wanted. But after doing so it seems to be upset about the ratio of added docstring lines to other new lines, which I don't see how that could ever make sense.

nijel:
Ideally all existing code would be documented and we could then enforce such checks in ruff, but we're nowhere close to that...

PR #5931, opened 2025-12-18
add filename in log

Jibec:
when working Fedora Linux community stats, only for Fedora 43, there is 227 830 po files on which I use calstats
Having the filename saves me a bit of time to find issues ;)


nijel:

PR #6014, opened 2026-01-13
fix: update storevalues method to handle metadata conditionally

Korpyc:
Currently, empty metadata dictionaries are serialized as "@key": {} in ARB JSON output, adding visual noise without value.
This change modifies ARBJsonUnit.storevalues() to conditionally store metadata only when self.metadata is non-empty (has at least one key), eliminating empty metadata entries from the generated files.


nijel:
Thanks! Can you please also add a test case to catch such behavior?

nijel:

PR #6089, opened 2026-02-14
fix(prop2po): preserve template values as msgid in .strings merge mode

pereorga:
Commit ae1b90b97 added bilingual behavior in `convertunit` for single-file mode but didn't account for the merge (template) path, breaking `mergestore`. When using `prop2po -t template.strings input.strings`, both msgid and msgstr now end up as the key name, losing all translated values.

The fix makes the bilingual behavior conditional: it only applies in single-file mode (`convertstore`), not in merge mode (`mergestore`), preserving the standard monolingual conversion where the template value becomes the msgid. Also, this adds regression tests for merge mode with both strings (UTF-16) and strings-utf8 personalities.

(Also: removing an outdated comment in a passing test).


nijel:

PR #6348, opened 2026-06-02
feat(markdown): add value-only YAML front matter translation mode

kvdb:
https://github.com/translate/translate/pull/6282 implemented initial MDX support. It doesn't quite work for me without this PR yet.

Add an opt-in frontmatter_translate_values flag to MarkdownFile. When enabled, YAML front matter is parsed with ruamel.yaml (round-trip mode) and one translation unit is emitted per scalar string value, using the key path as the unit location/docpath (e.g. frontmatter.metaTitle, frontmatter.authors[0]). On serialization the keys, structure, quoting style and comments are preserved and only values are replaced by their translations. Malformed YAML falls back to the existing opaque-blob behavior.

The flag defaults to False, so the existing extract_frontmatter blob behavior is unchanged. ruamel.yaml is imported lazily so the markdown extra keeps working without it when the flag is off. The flag is also threaded through MDXFile.

First cut: block scalars, block-sequence indentation, and non-string scalars are handled per the documented TODOs.

Assisted-by: AI assistant

### DIFF
--- a/translate/storage/markdown.py
+++ b/translate/storage/markdown.py
@@ -100,7 +100,7 @@ class MarkdownFile(base.TranslationStore[MarkdownUnit]):
           for translation. If False, it is preserved as-is.
         :param frontmatter_translate_values: if True, front matter is parsed as
           YAML and one translation unit is emitted per scalar string value, using
-          the key path as the unit location/context (e.g. ``frontmatter.metaTitle``).
+          the key path as the unit location and docpath (e.g. ``frontmatter.metaTitle``).
           On serialization the YAML structure, keys, quoting style and comments are
           preserved and only the values are replaced by their translations. If
           False (default), ``extract_frontmatter`` governs front matter handling and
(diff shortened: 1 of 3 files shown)


translate/storage/markdown.py
```diff
@@ -160,6 +182,159 @@ def parse(self, data) -> None:
...
+            and the comment / fold positions are carried over to the re-wrapped
+            node. See ``_translate_frontmatter_child``.
+          * Only scalar *string* values are translated. Numbers, booleans, dates
+            and ``null`` are left untouched on purpose.
+          * Nested maps and lists are walked recursively; list items use a
+            ``[i]`` index segment in the location/docpath.
+          * If the body fails to parse as YAML, fall back to the opaque-blob
+            behavior so a malformed file never raises here.
+        """
+        from ruamel.yaml import YAML, YAMLError  # noqa: PLC0415
```
nijel, line 228: I don't see a strong reason for importing this here, it should be at the top level.

nijel (commented):
Thanks for implementing this, I've made one minor suggestion and triggered a copilot review, which might bring more feedback.

translate/storage/markdown.py
kvdb: added a small code remark about it, let me know if you still want it different

nijel:
@codex review this pull request.

translate/storage/markdown.py
```diff
@@ -160,6 +182,178 @@ def parse(self, data) -> None:
...
+            )
+
+        self._walk_frontmatter(data, "frontmatter")
+
+        buffer = StringIO()
+        yaml.dump(data, buffer)
+        rendered_body = buffer.getvalue()
+        # ruamel terminates the dump with a trailing newline; the fence lines and
+        # any preserved trailing blank lines are added back explicitly.
+        rendered_body = rendered_body.rstrip("\n")
```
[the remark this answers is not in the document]
nijel: This looks like a valid concern. Once addressed, I think this is good to merge.

nijel:

PR #6397, opened 2026-06-23
Use ternary operator in po2txt.run()

Adichapati:
Simplify the template_file conditional assignment in `po2txt.run()` to a single-line ternary expression per ruff SIM108 guidance.

This is a small, safe code quality improvement that makes the converter more idiomatic without changing behavior.

## Files changed
- `translate/convert/po2txt.py`

## Verification
- `python3 -m py_compile translate/convert/po2txt.py` passes

### DIFF
--- a/translate/convert/po2txt.py
+++ b/translate/convert/po2txt.py
@@ -135,7 +135,9 @@ class po2txt:
         if not self.should_output_store:
             return False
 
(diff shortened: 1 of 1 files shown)


nijel:
Sorry, but this does not simplify the code, it makes it harder to read because of the lenghty ternary operator.

PR #6453, opened 2026-07-14
chore: align package license with GPL-3 code

nijel:
odf_shared.py includes GPL-3.0-or-later code copied from itools, so the distributed package cannot be offered under GPL-2.0. Select GPL-3.0-or-later in package and REUSE metadata and ship the matching COPYING text.


nijel:
This was uncovered by https://github.com/translate/translate/pull/6451

PR #6459, opened 2026-07-15
fix(po2xliff): propagate target language from PO header

### LINKED ISSUE #6391: po2xliff not writing the target language in xliff file tag of generated file yet it exists in po file
I used the `po2xliff` command to convert a bunch of po files to xliff.
The resultant xliff file has the header below
```xml
<?xml version="1.0" encoding="UTF-8"?>
<xliff xmlns="urn:oasis:names:tc:xliff:document:1.1" version="1.1">
  <file original="po/_docs.ar.po" source-language="en-US" datatype="plaintext">
    <body>
      <trans-unit xml:space="preserve" id="2" approved="no" restype="x-gettext-domain-header">
```
The target language is missing from the file section as it is optional. as per the information about the [file section](https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#file).

The original po file has the header below
```po
"Project-Id-Version: PACKAGE VERSION\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2026-06-09 11:45+0300\n"
"PO-Revision-Date: 2026-05-18 07:32+0000\n"
"Last-Translator: "
"Language-Team: Arabic \n"
"Language: ar\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=6; plural=n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 ? 4 : 5;\n"
"X-Generator: Weblate 2026.6.dev0\n"
```
The target language exists in the po file and can be accessed with the Language key.

I think it would make sense to have the xliff header also have the target language in the file tag if it exists in the po file. Something like
```xml
<?xml version="1.0" encoding="UTF-8"?>
<xliff xmlns="urn:oasis:names:tc:xliff:document:1.1" version="1.1">
  <file original="po/_docs.ar.po" source-language="en-US" target-language="ar" datatype="plaintext">
    <body>
      <trans-unit xml:space="preserve" id="2" approved="no" restype="x-gettext-domain-header">
```

 I looked up the target language in the xliff spec and it is listed as undefined as per the info at [target language](https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#target-language) .

Sanjays2402:
PO files can declare their target locale in the `Language` header, but `po2xliff` only populated XLIFF's `target-language` when callers passed it explicitly. This uses the PO header as the fallback while preserving explicit overrides, with a regression test covering both paths.

Fixes #6391

### DIFF
--- a/tests/translate/convert/test_po2xliff.py
+++ b/tests/translate/convert/test_po2xliff.py
@@ -384,6 +384,27 @@ msgstr "مرحبا"
         assert self.po2xliff(minipo).targetlanguage == "ar"
         assert self.po2xliff(minipo, targetlanguage="es").targetlanguage == "es"
 
+    def test_target_language_missing_from_po_header(self) -> None:
+        minipo = r"""msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+
+msgid "hello"
+msgstr "hello"
+"""
+        assert self.po2xliff(minipo).targetlanguage is None
+
+    def test_empty_target_language_in_po_header(self) -> None:
+        minipo = r"""msgid ""
+msgstr ""
+"Language: \n"
+"Content-Type: text/plain; charset=UTF-8\n"
+
+msgid "hello"
+msgstr "hello"
+"""
+        assert self.po2xliff(minipo).targetlanguage is None
+
     def test_variables(self) -> None:
         minipo = r'''msgid "%s%s%s%s has made %s his or her buddy%s%s"
 msgstr "%s%s%s%s het %s sy/haar vriend/vriendin gemaak%s%s"'''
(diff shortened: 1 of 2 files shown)


nijel:
Thanks for the pull request, it looks good, but it would be good to add tests that cover behavior with lack of Language header and when it is empty. I will also trigger Codex review now.

@codex review

Sanjays2402:
@nijel Added explicit coverage for both requested cases in `767ffa917`: a missing `Language` header and an empty `Language` header. Both verify that the generated XLIFF leaves `targetlanguage` unset.

Verification: **23 po2xliff tests passed**, and all prescribed `prek` hooks passed.

nijel:
@codex review

nijel:

PR #6480, opened 2026-07-25
fix(po2xliff): only split a location into linenumber when numeric

### LINKED ISSUE #5553: po2xliff does not convert location
> Seems like po2xliff did not transfer the DOM information stored in the po file into the `id` property:
> ```
> #: office:document-content[0]/office:body[0]/office:text[0]/table:table[0]/table:table-row[0]/table:table-cell[0]/text:p[0]/text:span[2]
> msgid "Culture 1"
> msgstr "Kultur 1"
> ``` 

 _Originally posted by @SantosSi in [#5516](https://github.com/translate/translate/issues/5516#issuecomment-2749314442)_

Sanjays2402:
Closes #5553

`po2xliff.contextlist()` split every location on the first `:`, so an ODF-derived location like `office:document-content[0]/office:body[0]/text:p[0]` became `sourcefile="office"` plus a `linenumber` holding the rest of the path, and the location was lost on round-trip. It now only splits when the trailing segment is numeric, and uses `rsplit` — matching `XliffUnit.addlocation`/`getlocations` in `translate/storage/xliff.py`.

New `test_location_without_linenumber` fails on master (`('sourcefile', 'office')` + bogus linenumber) and passes with the fix; existing `test_locationcomments` (`file.c:123 asdf.c`) is unchanged.

### DIFF
--- a/tests/translate/convert/test_po2xliff.py
+++ b/tests/translate/convert/test_po2xliff.py
@@ -157,6 +157,20 @@ msgstr "kunye"
             ("sourcefile", "asdf.c"),
         ]
 
+    def test_location_without_linenumber(self) -> None:
+        """Non-numeric locations must not be split into sourcefile/linenumber."""
+        location = "office:document-content[0]/office:body[0]/text:p[0]"
+        minipo = f"""#: {location}
+msgid "one"
+msgstr "kunye"
+"""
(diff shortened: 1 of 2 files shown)


nijel:
@codex review

nijel:

PR #6524, opened 2026-08-16
feat(toml): migrate from tomlkit to tomlrt

dimbleby:
Hi -

I have been working on [tomlrt](https://pypi.org/project/tomlrt/), which is a project occupying roughly the same space as tomlkit: but hopefully with a simpler API (and faster, and less buggy).

Mostly for my own interest, I experimented with migrating this project to tomlrt - but I think it turned out to be a good enough fit that it is worth offering you this pull request.  

The tomlrt comment API saves a lot of grubbing around in internal tomlkit types.  Per the commit message and new tests: we have fixed a few bugs along the way.

(re "Only comments directly above a key become notes...".  If the old behaviour of including earlier comments separated by blank lines from the value was intended: tomlrt can do that too).

Do feel free to close out if this is of no interest, you might reasonably think that you have tomlkit working well enough already and do not need to risk change.

---

tomlrt parses into plain dicts, lists and scalars, and exposes comments as mapping views, so the storage layer no longer works against CST wrapper types. Behaviour changes worth noting:

- Comments on keys inside a [section] now become notes. _get_key_comment walked tomlkit's CST behind a hasattr(table, "body") guard that was always False for a Table, so these were silently always dropped. Array elements and Go i18n section headers now get notes too.

- Only comments directly above a key become notes. Previously a banner separated by a blank line was swept into the following key's note.

- Go i18n untranslated plural forms serialize as empty strings. They were substituted with None, which TOML cannot represent, so serializing raised.

- Serializing uses tomlrt.dump, which keeps the file's line endings. A CRLF file missing a trailing newline is no longer given an LF one.

Reading and writing go through load/dump, which are always UTF-8 as the TOML spec requires, so the store's encoding setting is no longer used.

DictUnit gains dict_factory so a format can pick the mapping type built while serializing nested values; tomlrt renders a plain dict as an inline table, and TOML localization files want [section] blocks.

Round-trip tests only checked that some comments appeared somewhere in the output. They now assert the output exactly.

### DIFF
--- a/docs/formats/toml.rst
+++ b/docs/formats/toml.rst
@@ -86,7 +86,7 @@ Supported Features
 * **Comments**: Developer comments are extracted as translator notes
 * **String types**: Basic strings, literal strings (single quotes), and
   multiline strings (triple quotes)
-* **Formatting preservation**: tomlkit library preserves original formatting
+* **Formatting preservation**: tomlrt library preserves original formatting
   during round-trips
 * **Plurals** (Go i18n only): CLDR plural categories
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ subtitles = [
   "aeidon>=1.15,<2.1"
 ]
 toml = [
-  "tomlkit>=0.13.0,<0.16.0"
+  "tomlrt>=2.2.2,<3.0"
 ]
 yaml = [
   "ruamel.yaml>=0.19.1,<0.20.0"
--- a/tests/translate/storage/test_toml.py
+++ b/tests/translate/storage/test_toml.py
@@ -28,6 +28,19 @@ class TestTOMLResourceStore(test_monolingual.TestMonolingualStore):
         store.parse("")
         assert bytes(store) == b""
 
+    def test_crlf_line_endings(self) -> None:
+        """Test that CRLF line endings survive a roundtrip."""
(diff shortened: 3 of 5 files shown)


nijel:
Thanks for your pull request. Honestly, I'm not sure if this is a good idea. tomlrt is a new project with a single maintainer, while tomlkit is proven to be maintained for years. It definitely lacks some features, but I'm not sure if it is worth of the risk. We already depend on enough projects whose sole maintainer lost interest in maintaining them.

dimbleby:
I understand!

I've privately migrated lots of projects to tomlrt - as part of tomlrt's development, rather than intending to make changes to those projects.  And mostly I haven't bothered to show maintainers the results, tomlkit is good enough.

Here I judged that the API improvements and bugs fixed along the way were enough to make it worth asking you the question.  

For what it's worth, tomlkit is not overloaded with maintainers right now eg https://github.com/python-poetry/tomlkit/issues/574.  Indeed - as at today - I'd like your chances of getting a bug fix from tomlrt rather more than from tomlkit.  But who can say about tomorrow?  And your broader point that tomlkit is longstanding and well-known is certainly correct.

As I said: do feel free to close this out - it's very reasonable to consider that you have tomlkit working well enough already and there is no need to risk change.

